PR #82, opened 2018-08-23
注解支持设置origin

lostingz:
目前注解功能无法指定origin，只能设置limitApp为default 或者 other，导致部分限流功能无法使用（在使用注解情况下）

### Describe how you did it
注解添加origin，SentinelResourceAspect 中ContextUtil.enter方法判定annotation中origin是否不为空，
如果设置有值，设置origin的值，否则走默认origin
### Describe how to verify it
unit test

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
@@ -40,6 +40,13 @@ public @interface SentinelResource {
     String value();
 
     /**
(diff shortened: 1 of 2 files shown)


sczyh30:
Hi, thanks for your contribution. Adding origin to `@SentinelResource` can be confusing as `origin` does not belong to a resource. And it's weird to configure the `origin` literally in the annotation.

The `origin` should be set only in the **entrance** of the top-level resource (e.g. gateway or Dubbo service) using `ContextUtil.enter(entranceName, origin)`. Only the first `ContextUtil.enter` will take effect.

PR #83, opened 2018-08-23
add rule builder

kimmking:
### Describe what this PR does / why we need it

add FlowRuleBuilder and DegradeRuleBuilder for creating rules

### Describe how you did it

add FlowRuleBuilder and DegradeRuleBuilder for creating rules

### Describe how to verify it

add FlowRuleBuilder and DegradeRuleBuilder for creating rules

```
new FlowRuleBuilder()
       .addRule("app","key1",10,RuleConstant.FLOW_GRADE_QPS)
       .addRule("app","key2",20,RuleConstant.FLOW_GRADE_QPS)
       .addRule("app","key3",30,RuleConstant.FLOW_GRADE_QPS)
       .load();
```

PR #87, opened 2018-08-24
Polish Dubbo adapter

kimmking:
Polish Dubbo adapter

### DIFF
--- a/sentinel-adapter/sentinel-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/DubboAppContextFilter.java
+++ b/sentinel-adapter/sentinel-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/DubboAppContextFilter.java
@@ -36,7 +36,7 @@ public class DubboAppContextFilter implements Filter {
     public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
         String application = invoker.getUrl().getParameter(Constants.APPLICATION_KEY);
         if (application != null) {
-            RpcContext.getContext().setAttachment(DubboUtils.DUBBO_APPLICATION_KEY, application);
+            RpcContext.getContext().setAttachment(DubboUtils.SENTINEL_APPLICATION_KEY, application);
         }
         return invoker.invoke(invocation);
     }
--- a/sentinel-adapter/sentinel-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/DubboUtils.java
+++ b/sentinel-adapter/sentinel-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/DubboUtils.java
@@ -22,13 +22,13 @@ import com.alibaba.dubbo.rpc.Invocation;
  */
 public final class DubboUtils {
 
-    public static final String DUBBO_APPLICATION_KEY = "dubboApplication";
+    public static final String SENTINEL_APPLICATION_KEY = "sentinelApplication";
 
     public static String getApplication(Invocation invocation, String defaultValue) {
         if (invocation == null || invocation.getAttachments() == null) {
             throw new IllegalArgumentException("Bad invocation instance");
         }
-        return invocation.getAttachment(DUBBO_APPLICATION_KEY, defaultValue);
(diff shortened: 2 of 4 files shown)


sentinel-adapter/sentinel-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/DubboAppContextFilter.java
```diff
@@ -36,7 +36,7 @@
     public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
         String application = invoker.getUrl().getParameter(Constants.APPLICATION_KEY);
         if (application != null) {
-            RpcContext.getContext().setAttachment(DubboUtils.DUBBO_APPLICATION_KEY, application);
+            RpcContext.getContext().setAttachment(DubboUtils.SENTINEL_APPLICATION_KEY, application);
```
sczyh30: Using `SENTINEL_APPLICATION_KEY` may confuse users as the attachment value refers to application name in Dubbo, not Sentinel appName.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/DegradeRuleBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.alibaba.csp.sentinel.slots.block.degrade.DegradeRule;
+import com.alibaba.csp.sentinel.slots.block.degrade.DegradeRuleManager;
+import com.alibaba.csp.sentinel.slots.block.flow.FlowRule;
+import com.alibaba.csp.sentinel.slots.block.flow.FlowRuleManager;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class DegradeRuleBuilder {
```
sczyh30: Duplicate with #83?

sentinel-adapter/sentinel-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/DubboAppContextFilter.java
kimmking: the key inserted into dubbo context, so marked as sentinel key may be much correcter than dubbo key.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/DegradeRuleBuilder.java
kimmking: branch conflict

sentinel-adapter/sentinel-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/DubboAppContextFilter.java
kimmking: @sczyh30

sczyh30: @CarpenterLee What do you think of here?

CarpenterLee: I prefer @kimmking 's opinion. Since we use `DubboAppContextFilter` to modify the Dubbo's context, uses should know that the attachment belongs to Sentinel, instead of Dubbo.

sczyh30: That's OK, but we need to add a comment on this field to indicate that its value refers to application name configured in Dubbo, not Sentinel appName (though our document has emphasized that).

sczyh30:
Hi, due to the fact that changing the key may bring breaking change to current users who've using Sentinel Dubbo adapter in production, we cannot directly modify the key. We'll add javadoc to clarify this so that users will not be confused with the key. Thanks for contributing!

PR #113, opened 2018-08-30
FileRefreshableDataSource读取配置文件bug

pengminxuan:
在使用sentinel-demo-dynamic-file-rule项目进行测试时,发现FileRefreshableDataSource没有对文件路径进行解码.如果文件路径中含有空格会被编译为"%20",导致读取rule配置文件失败.

### Describe how you did it
FileRefreshableDataSource的构造函数在接收fileName时会对fileName进行UTF-8解码


sczyh30:
Hi, the cause is that `classLoader.getResource(xxx)`  returns encoded URL path, so we have to manually decode it. You can fix in demo like:

```java
String flowRulePath = URLDecoder.decode(classLoader.getResource("FlowRule.json").getFile(), "UTF-8");
```

We can't decode this in `FileRefreshableDataSource` directly because we assume the file path is normal. The filename may be actually like `a%20b.json`. Decoding this can lead to failure.

PR #116, opened 2018-09-02
add jdbc implement of DataSource

cdfive:
add JdbcDataSource,  for reading rules from database, like mysql.

NONE

### Describe how you did it
1.design four tables for store the app and rule data, and the ddl.sql is in the README.md;
2.create a JdbcDataSourceClass extends from AutoRefreshDataSource;
   use SpringTemplate execute sql to query data, which is an argument of constructor;
   override the readSource() method, use sql query rules data from database.

### Describe how to verify it
1.run unit test case.
2.a simple dubbo application demo which use it.
https://github.com/cdfive/sentinel-support
https://github.com/cdfive/sentinel-support-demo/tree/master/dubbo-demo/provider

### Special notes for reviews
NONE


sczyh30:
Hi, thanks for your contribution! We are not supposed to restrict the table structure, and it's difficult to implement a common, universal JDBC data source. So it's better to be an third-party module in individual repository instead of an official module of Sentinel.

You can create a new repository and put your code into here, then we'll create a *Sentinel Awesome` page to show nice extensions by third-party users, and you can reference your implementation here.

cdfive:
@sczyh30 Last night i thought out a new idea. Removing the Spring JdbcTemplte and the fixed SQL query,  use DataSource, sql, sqlParameter instead. The datasource is the place where user's database, the sql and sqlParameter are used for query effective sentinel rules from databse. By this，user can freely design tables for storeage, and the JdbcDataSource extension only dependency on sentinel-datasource-extension. The JdbcDataSource Extension do jdbc query and convert the query result to sentinel rule list for users. Do you think it works acceptable?

sczyh30:
Since the JDBC driver, SQL and SQL parameters are all passed by users, I think it's more appropriate to implement it directly by users. Users can choose to use Spring JDBC, MyBatis or directly JDBC API. Abstracting a common, universal solution for RDBMS may bring other restrictions and complexity (e.g. config, object mapper).

sczyh30:
Hi, our awesome repo is ready. You can refer to: https://github.com/alibaba/sentinel-awesome

sczyh30:
Closing this PR as it has been added to [Awesome Sentinel](https://github.com/alibaba/sentinel-awesome).

PR #146, opened 2018-09-12
Add support for asynchronous invocation entry

### LINKED ISSUE #31: [RFC] Add support for asynchronous invocation
## Motivation

The invocation chain (Context) of Sentinel does not work well with asynchronous invocation, leading to incorrect context and response.

## Target

Support asynchronous invocation.

## Solution

Pending

sczyh30:

Add support for asynchronous invocation entry so that users can use Sentinel in asynchronous logic.

Resolves #31 

#### Internal

- Add several `asyncEntry` method in `Sph` interface and `SphU` class.
- Add a new `AsyncEntry` class for asynchronous entry representation, which resembles the linked `CtEntry`. The difference:
    - When calling `SphU.asyncEntry`, the created async entry is set to current context. As soon as the entry passed the slot chain (or blocked), it will be cleared from current context intermediately so that it won't affect the proceeding entries.
    - When the created async entry successfully passed the slot chain in `CtSph#asyncEntryInternal`, the async context inside the async entry will be initialized. We can get the async context via `AsyncEntry#getAsyncContext` method and use with `ContextUtil#runOnContext` to do context switching in asynchronous callback.
- Add `runOnContext` and `replaceContext` method in `ContextUtil` for context switching.

#### For users

We can call `SphU.asyncEntry(xxx)` to entry for asynchronous invocation, then `exit` it in the callback (or other place, often in different thread). If we want nested entry in asynchronous entry, we have to wrap the logic within the async context via `ContextUtil.runOnContext(asyncContext, lambda)`.

Here is an example:

```java
public void testAsync() {
    try {
        final AsyncEntry entry = SphU.asyncEntry(resourceName);

        CompletableFuture.runAsync(() -> {
            ContextUtil.runOnContext(entry.getAsyncContext(), () -> {
                try {
                    TimeUnit.SECONDS.sleep(2);
                    
                    // Call another sync resource under async context.
                    doAnotherSync();
                } catch (InterruptedException e) {
                    // Ignore.
                } finally {
                    // User should exit the resource in the callback.
                    entry.exit();
                }
            });
        });
    } catch (BlockException ex) {
        // Resource blocked. 
        ex.printStackTrace();
    }
}
```

### Describe how to verify it

See test cases.


wavesZh:
as you said: 
> When calling SphU.asyncEntry, the created async entry is set to current context. As soon as the entry passed the slot chain (or blocked), it will be cleared from current context intermediately so that it won't affect the proceeding entries.

Other processing will get wrong entry for the context during `SphU.asyncEntry` changed curEntry, it's ok?

PR #194, opened 2018-10-23
Config whether pid is needed in the log file name

CarpenterLee:

Now sentinel log file name contains pid(process id) to distinguish different instances of same app in same machine, but in most of the case, only one instance will be deployed in the same machine, there is no need to add pid in file name in this case.

So we add a `LOG_NAME_USE_PID` System property to config whether pid is needed in the log file and metric file name.

By the way, we change the log file name and metric file name to a more readable one.

NONE.

### Describe how you did it

Add a `LOG_NAME_USE_PID` System property to config whether pid is needed in the log file and metric file name.

### Describe how to verify it

Run test cases.

### Special notes for reviews
NONE.


sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/LogBase.java
```diff
@@ -33,9 +36,20 @@
...
+    public static final String LOG_NAME_USE_PID = "csp.sentinel.log.use.pid";
+    private static boolean logNameUsePid = false;
+
     private static String logBaseDir;
 
     static {
+        try {
+            init();
+        } catch (Throwable t) {
+            t.printStackTrace();
```
sczyh30: Once an error is thrown (e.g. class initialization error due to no permission), the application might need to fail immediately, as it may cause weird behaviors (e.g. NPE when publishRecord).

PR #219, opened 2018-11-01
Modify log info about fetching metrics

### LINKED ISSUE #215: INFO  c.t.c.s.d.metric.MetricFetcher - handleBody line error: No metrics
## Issue Description

Type: *bug report*

Sentinel Dashboard improperly handling "no metrics" response when fetch application metrics.

1. "No metrics" response is a normal return result, can't be considered as an error, so should not print error message.
2. In the case where the response content is "No metrics", there is a performance loss due to a "java.lang.NumberFormatException: For input string: "No metrics"" exception.
3. The log content does not contain host and process information, so the problem can't be accurately located.
4. The log is printed when an exception occurs, so it should not be the info level.

Because of the above, I think should judge the response content first, special handling of the "no metric" case, such as printing an info log，modifying the content and level of the log in the `try-catch` code block. 

Start a client and a server dashboard.

### Tell us your environment
macOS Sierra Version 10.12.6

java version "1.8.0_144"
Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)

sentinel SDK version: 1.3.0-GA
sentinel dashboard version: master branch(commit: e322df2)

xcaspar:
See #215  for details.

Fixes #215 
### Describe how you did it
Add a "No metrics" content judgment, modify the log level and message in the try-catch block, and add a method in the `MachineInfo`  class to identify the application in the log message.

### Describe how to verify it
Run a client and dashboard and test for metrics and no metrics

### Special notes for reviews
The `MetricFetcher ` file is formatted, so some of attributes line numbers are changed.

### DIFF
--- a/sentinel-dashboard/src/main/java/com/taobao/csp/sentinel/dashboard/discovery/MachineInfo.java
+++ b/sentinel-dashboard/src/main/java/com/taobao/csp/sentinel/dashboard/discovery/MachineInfo.java
@@ -130,4 +130,13 @@ public class MachineInfo implements Comparable<MachineInfo> {
     public int hashCode() {
         return Objects.hash(app, ip, port);
     }
+
+    /**
+     * Information for log
+     *
+     * @return
+     */
+    public String toLogString() {
+        return app + "|" + ip + "|" + port + "|" + version;
+    }
 }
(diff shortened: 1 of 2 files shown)


sczyh30:
Thanks for contributing!

Issue #207, opened 2018-10-29
[Feature] Support priority of resource entries


jialianglinjl:
## Issue Description

Type: *feature request*

### Describe what feature you want

Support priority of resource entries.

- [x] Add priority flag support in `SphU` (`ProcessorSlot`) (#255) 
- [ ] Add priority flag support in `Context` (implicit for invocation chain)
- [x] Add implementation for prioritized flow control (cluster mode)
- [x] Add optimized implementation for prioritized resource entries (local mode)

tigerMoon:
hi. can you give more details about why we need priority flag? and the scenario to use it.

sczyh30:
> hi. can you give more details about why we need priority flag? and the scenario to use it.

When requests with two kinds of flag (`flag1` and `flag2`) are coming respectively and simultaneously, we want one kind of request (e.g. `flag1`) pass first, but the current quota (pass QPS) may be full. Thus some strategies are needed to guarantee the prioritized requests can finally pass rather than rejected.

Perhaps we need a more vivid representation to describe the feature :)

SugisakiKen:
I write a demo to verify this feature,  but maybe there is something wrong with it

````
/**
 * demo to simulate two operation : read/write , they share same resource
 * read operation use SphU.entryWithPriority() , write operation use SphU.entry()
 * create 25 write thread and 25 read thread,and flow limit count is 20
 * in order to create race condition , use CyclicBarrier to let all thread to start in relatively fair condition
 * when SphU.entryWithPriority() or SphU.entry() success,corresponding counter will add 1
 */
public class FlowQpsReadWritePriorityDemo {

    private static final String WRITE_KEY = "write";
    private static final String READ_KEY = "read";
    private static CyclicBarrier cyclicBarrier = new CyclicBarrier(50);
    private static List<String> list = Collections.synchronizedList(new ArrayList<>(30));
    private static AtomicInteger readCounter = new AtomicInteger(0);
    private static AtomicInteger writeCounter = new AtomicInteger(0);
    private static final int threadCountWrite = 25;
    private static final int threadCountRead = 25;
    private static final Object ENTRY_LOCK = new Object();
    private static ConcurrentHashMap<String, Object> lockMap = new ConcurrentHashMap<>();

    public static void main(String[] args) {
        initFlowQpsRule();
        simulateTraffic();
        rouse();
        System.out.println("===== begin to do flow control");
        System.out.println("only 10 requests per second can pass");
    }

[...]

PR #255, opened 2018-11-19
Support prioritized entry in ProcessorSlot interface

sczyh30:

Support prioritized entry in ProcessorSlot interface so that each slot can handle requests with priority.

Related to #207 

### Describe how you did it

- Refactor `ProcessorSlot` interface (add a parameter `prioritized` in entry-related methods)
- Add `entryWithPriority` methods in `SphU` (separate from `entry` method so that it won't be confused with previous `entry` with args)

### Describe how to verify it

This is only a change for interface. Run all test cases to ensure nothing broken.

### Special notes for reviews

NONE

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/SphU.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/SphU.java
@@ -242,4 +242,31 @@ public class SphU {
     public static AsyncEntry asyncEntry(String name, EntryType type, int count, Object... args) throws BlockException {
         return Env.sph.asyncEntry(name, type, count, args);
     }
+
+    /**
+     * Checking all {@link Rule}s related the resource. The entry may have priority.
+     *
+     * @param name        the unique name for the protected resource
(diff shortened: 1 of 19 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/SphU.java
```diff
@@ -242,4 +242,31 @@ public static AsyncEntry asyncEntry(String name, EntryType type) throws BlockExc
...
+
+    /**
+     * Checking all {@link Rule}s related the resource. The entry may have priority.
+     *
+     * @param name        the unique name for the protected resource
+     * @param prioritized whether the entry is prioritized
+     * @throws BlockException if the block criteria is met, eg. when any rule's threshold is exceeded.
+     * @since 1.4.0
+     */
+    public static Entry entryWithPriority(String name, boolean prioritized) throws BlockException {
```
CarpenterLee: The `prioritized` parameter can be removed, since the method name `entryWithPriority` already indicates the priority meaning.

PR #291, opened 2018-12-07
fix grpc-adapter bug,replace the Entry by AsyncEntry #289

### LINKED ISSUE #289: [Bug] grpc adapter-Tracer.trace() can't success
## Issue Description
Here is code url:
[code](https://github.com/alibaba/Sentinel/tree/master/sentinel-adapter/sentinel-grpc-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/grpc)
 the interceptor's(Server or Client) interceptCall method and the Listener method  may run on different Thread, So,Tracer.trace(throwable) may not success if current Thread don't  have the Context. I think here is should use `AsyncEntry` and call 

``` 
ContextUtil.runOnContext(finalEntry.getAsyncContext(), new Runnable() {
                                            @Override
                                            public void run() {
                                                Tracer.trace(new Exception());
                                            }
                                        });
```

CodingSinger:

fix the bug and perfect the grpc-adapter ,replace the Entry by AsyncEntry

Fixes #289


sczyh30:
Thanks for contributing!

PR #321, opened 2018-12-20
add some unit test for sentinel-transport-netty-http module

cdfive:

The `sentinel-transport-netty-http` module doesn't have test package,  add some unit test.

NONE

### Describe how you did it

Import `junit` and `mockito-core` dependency, and use Netty's `EmbeddedChannel` for test ChannelHandler.

### Describe how to verify it

run mvn test

### Special notes for reviews
one test case `testGetRuleCommandFlowSomeRules`, since I haven't found a good way yet，comment the `@Test` annotation and marked it with the `FIXME` comment, for improving it in the future.


sczyh30:
Great! Thanks for contributing!

PR #353, opened 2018-12-28
示例中 isEmpty 导致规则不能删除

foreveryang321:
Translated from Google:
In the example isEmpty causes the rule to not be deleted

NONE

Describe how you did it
Describe how to verify it
Special notes for reviews


sczyh30:
Thanks for contributing!

PR #409, opened 2019-01-12
 Add a CommandCenterProvider to resolve CommandCenter API and cache the instance

### LINKED ISSUE #345: CommandCenter providers stop method but we can not get instance and invoke it
## Issue Description

Spring Boot expose `/shutdown` Endpoint to shutdown Application. Sentinel-Starter includes `sentinel-transport-simple-http` module and this module using SPI to trigger `SimpleHttpCommandCenter`.

After invoke `/shutdown` endpoint, the process will not exit because the `ServerSocket` created in `SimpleHttpCommandCenter` is alive.

So I can invoke `CommandCenter#stop` to shutdown ThreadPool if I can get `CommandCenter` instance.

Type: feature request

The single `CommandCenter` in `CommandCenterInitFunc` maybe should be hold in some where.

So we can get it and invoke `CommandCenter#stop` at any time.

cdfive:

Cache the CommandCenter instance so that we can invoke any methods with it when needed.

Fixes #345 

### Describe how you did it

Add a Class `CommandCenterProvider` which hold a CommandCenter instance, resolve the instance 
by SPI and add a `getCommandCenter()` method to get the instance.

### Describe how to verify it

Run sentinel-dashboard and make a breakpoint to test if CommandCenter loaded as expected.

### Special notes for reviews

I found a useful method `SpiLoader.loadFirstInstance(CommandCenter.class);` to get the API，
which only get the first API instance, 
but previous `CommandCenterInitFunc` checked "Only single command center can be started",
I think it's okay not to check the unique, since load first API can worked and user can also
check in sentinel-record.log.

### DIFF
--- a/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/command/CommandCenterProvider.java
+++ b/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/command/CommandCenterProvider.java
@@ -1,0 +1,37 @@
+package com.alibaba.csp.sentinel.command;
+
+import com.alibaba.csp.sentinel.log.RecordLog;
+import com.alibaba.csp.sentinel.transport.CommandCenter;
+import com.alibaba.csp.sentinel.util.SpiLoader;
+
+/**
+ * Provider for a universal {@link CommandCenter} instance.
+ *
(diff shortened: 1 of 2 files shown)


sczyh30:
Thanks for contributing!

PR #457, opened 2019-01-25
fix error value type and rename variable

mjaow:

1. fix the  error type int to long
2.rename rt in avgRt to be total according to its actual meaning

NONE

### Describe how you did it

fix int to long 
rename variable to be more meaningful

### Describe how to verify it

checked


sczyh30:
Thanks for contributing!

PR #460, opened 2019-01-26
fix default argusedtokens

### LINKED ISSUE #459: doubt about default value of avgUsedTokens
## Issue Description

the default traffic limiter has code
```
    private int avgUsedTokens(Node node) {
        if (node == null) {
            return -1;
        }
        return grade == RuleConstant.FLOW_GRADE_THREAD ? node.curThreadNum() : (int)node.passQps();
    }

```
I'm not sure why return -1 as default value.Seems Zero is more meaningful

Type: *bug report* or *feature request*

*bug report*

it return -1 as default value

expected return 0 as default value

mjaow:

fix default value (-1 -> 0) of arg_used_tokens

Fixes #459 

### Describe how you did it

fix default value -1 to 0

### Describe how to verify it

CI

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/RuleConstant.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/RuleConstant.java
@@ -53,6 +53,8 @@ public final class RuleConstant {
 
     public static final int DEFAULT_SAMPLE_COUNT = 2;
     public static final int DEFAULT_WINDOW_INTERVAL_MS = 1000;
+    public static final int DEFAULT_AVG_USED_TOKENS = 0;
 
-    private RuleConstant() {}
+    private RuleConstant() {
+    }
 }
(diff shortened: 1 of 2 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/RuleConstant.java
```diff
@@ -53,6 +53,8 @@
 
     public static final int DEFAULT_SAMPLE_COUNT = 2;
     public static final int DEFAULT_WINDOW_INTERVAL_MS = 1000;
+    public static final int DEFAULT_AVG_USED_TOKENS = 0;
```
sczyh30: The `DEFAULT_AVG_USED_TOKENS` should not be put into `RuleConstant` as it is supposed to keep all constants for rules. You can put it inside the `DefaultController` directly and keep the original `RuleConstant`.

mjaow: reasonable

mjaow:
fix is done

sczyh30:
Thanks for contributing!

PR #461, opened 2019-01-26
fix ratelimit bugs

mjaow:

Fix bugs when count is zero.

when count is zero,the costTime value will be max of long.After some retries of canPass,the waitTime will overflow and permit any request (as it will return true)

### Describe how to verify it

Test case in RateLimiterControllerTest.testPaceController_zeroattack


mjaow:
ok,i'll keep the bugfix part and  delete the commit which fix concurrent issue and reference annother issue to discuss it

PR #465, opened 2019-01-27
HashMap init optimize when adding new ClusterNode to cluster node map

luoxn28:
1. Init a new hash map size using old hash map size, avoid new hash map rehash when putAll operation .

NONE

### Describe how you did it
use old hash map size init new hash map size.

### Describe how to verify it
checked

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/clusterbuilder/ClusterBuilderSlot.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/clusterbuilder/ClusterBuilderSlot.java
@@ -81,7 +81,7 @@ public class ClusterBuilderSlot extends AbstractLinkedProcessorSlot<DefaultNode>
                 if (clusterNode == null) {
                     // Create the cluster node.
                     clusterNode = Env.nodeBuilder.buildClusterNode();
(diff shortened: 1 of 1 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/clusterbuilder/ClusterBuilderSlot.java
```diff
@@ -81,7 +81,7 @@ public void entry(Context context, ResourceWrapper resourceWrapper, DefaultNode
                 if (clusterNode == null) {
                     // Create the cluster node.
                     clusterNode = Env.nodeBuilder.buildClusterNode();
-                    HashMap<ResourceWrapper, ClusterNode> newMap = new HashMap<ResourceWrapper, ClusterNode>(16);
+                    HashMap<ResourceWrapper, ClusterNode> newMap = new HashMap<ResourceWrapper, ClusterNode>(Math.max(clusterNodeMap.size(), 16));
```
satansk: First, we can use type inference instead of explicit type parameter here:

```Java
new HashMap<>(...);
```

Second, the initial size should be the max of `(int) clusterNodeMap.size() / 0.75 + 1` and 16.

sczyh30:
Hi, could you please sign the CLA here: https://cla-assistant.io/alibaba/Sentinel?pullRequest=465

sczyh30:
Thanks for contributing!

Issue #235, opened 2018-11-09
Upgrade the minimum supported JDK version to 1.7


sczyh30:
## Issue Description

Type: *feature request*

### Describe what feature you want

Currently Sentinel supports JDK 1.6 or above version. This gets in the way of leveraging some new features of JDK. And according to statistics, there are very few developers still using JDK 1.6 now.

So since next milestone version we are planning to upgrade the minimum supported JDK version to 1.7.

sczyh30:
Accomplished in #484 (https://github.com/alibaba/Sentinel/commit/b4422407ae4b9aabd5d78f5e410c1f500fa260d4)

PR #484, opened 2019-02-12
[Feature] Add CPU usage checking support in system protection rule

### LINKED ISSUE #450: Add CPU utilization in system rule
## Issue Description

Type:  *feature request*

Currently, system rule is based on CPU load, global average rt, global in inbound QPS and total threads,  but not concern about  CPU utilization, which is more sensitive than CPU load.

Add CPU utilization in system rule, to gain more sensitive protection.

CarpenterLee:

Sentinel System Guard takes average RT, QPS, thread count and system load into account, but does not consider CPU usage. Comparing to system load, CPU usage is more sensitive, which is useful when vast requests coming in sudden.

Resolves #450 

### Describe how you did it

Add CPU usage threshold into system rule, and when current CPU usage reaches the threshold, a BBR likely checking will be done.

### Describe how to verify it

Rule SystemGuardDemo.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemRuleManager.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemRuleManager.java
@@ -66,6 +66,10 @@ import com.alibaba.csp.sentinel.slots.block.BlockException;
 public class SystemRuleManager {
 
     private static volatile double highestSystemLoad = Double.MAX_VALUE;
+    /**
+     * cpu usage, between [0, 1]
+     */
+    private static volatile double highestCpuUsage = Double.MAX_VALUE;
     private static volatile double qps = Double.MAX_VALUE;
     private static volatile long maxRt = Long.MAX_VALUE;
     private static volatile long maxThread = Long.MAX_VALUE;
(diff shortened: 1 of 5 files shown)


sczyh30:
As the `OperatingSystemMXBean#getSystemCpuLoad` API is introduced since JDK 1.7, the minimum JDK version of sentinel-core needs to be upgraded to 1.7 too (related to #235). This will bring breaking change so we're targeting this feature to next big milestone version (1.5.0).

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemRuleManager.java
```diff
@@ -89,7 +94,7 @@
     static {
         checkSystemStatus.set(false);
         statusListener = new SystemStatusListener();
-        scheduler.scheduleAtFixedRate(statusListener, 5, 1, TimeUnit.SECONDS);
+        scheduler.scheduleAtFixedRate(statusListener, 5, 100, TimeUnit.MILLISECONDS);
```
sczyh30: 100ms seems too frequent and might affect the performance. Maybe we can adjust it to 500ms or 1s?

sczyh30:
Thanks!

PR #491, opened 2019-02-18
Add a ApiCommandHandler which fetch all registered CommandHandlers so that user can get all available CommandHandlers by request /api

all4you:

Add a ApiCommandHandler which fetch all registered CommandHandlers so that user can get all available CommandHandlers by request /api

No. Only add new features

### Describe how you did it

Create a new CommandHandler called ApiCommandHandler

### Describe how to verify it

1. Run Application in sentinel-demo-command-handler module
2. Request 127.0.0.1:8080/init to init the Sentinel and CommandCenter
3. Request 127.0.0.1:8719/api to get all the available CommandHandlers


sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/command/annotation/CommandMapping.java
```diff
@@ -30,4 +26,7 @@
 public @interface CommandMapping {
 
     String name();
+
+    String desc();
```
sczyh30: Maybe it's better to be optional (default empty)?

all4you: Define the desc to be not empty aims to force user create CommandHandler with a detailed description. If set default empty when user get the api list they don't know what the exactly usage of the api. So I would prefer it not be default empty.

sczyh30: Okay, it's reasonable.

all4you:
> Maybe it's better to add an example to show how to register a customized `CommandHandler` via SPI in `sentinel-demo-command-handler` as it's a demo module.

done!

sczyh30:

PR #538, opened 2019-03-04
Optimize circuit breaking state transformation using CAS in DegradeRule

ATAXGT:
I feel that the method of modifying variables by synchronization is relatively poor, so I changed to CAS

NONE

### Describe how you did it
Modify the type of the property to AtomicBoolean and use compareAndSet method to change it.

### Describe how to verify it
Use DegradeTest.java to verity it

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
@@ -18,6 +18,7 @@ package com.alibaba.csp.sentinel.slots.block.degrade;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 
 import com.alibaba.csp.sentinel.concurrent.NamedThreadFactory;
@@ -80,7 +81,7 @@ public class DegradeRule extends AbstractRule {
      */
     private int grade = RuleConstant.DEGRADE_GRADE_RT;
 
-    private volatile boolean cut = false;
+    private final AtomicBoolean cut = new AtomicBoolean(false);
 
     public int getGrade() {
         return grade;
@@ -93,8 +94,6 @@ public class DegradeRule extends AbstractRule {
 
     private AtomicLong passCount = new AtomicLong(0);
 
-    private final Object lock = new Object();
-
     public double getCount() {
         return count;
     }
@@ -104,12 +103,12 @@ public class DegradeRule extends AbstractRule {
         return this;
     }
 
-    public boolean isCut() {
-        return cut;
+    private boolean isCut() {
+        return cut.get();
     }
 
     private void setCut(boolean cut) {
-        this.cut = cut;
+        this.cut.set(cut);
     }
 
     public AtomicLong getPassCount() {
@@ -162,7 +161,7 @@ public class DegradeRule extends AbstractRule {
 
     @Override
     public boolean passCheck(Context context, DefaultNode node, int acquireCount, Object... args) {
-        if (cut) {
+        if (cut.get()) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
```diff
@@ -206,16 +205,14 @@ public boolean passCheck(Context context, DefaultNode node, int acquireCount, Ob
             }
         }
 
-        synchronized (lock) {
-            if (!cut) {
-                // Automatically degrade.
-                cut = true;
+        if (!cut.get()) {
+            if (cut.compareAndSet(false, true)) {
+                System.out.println("123123123123");
```
sczyh30: Please remove the unused code.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
```diff
@@ -206,16 +205,14 @@ public boolean passCheck(Context context, DefaultNode node, int acquireCount, Ob
             }
         }
 
-        synchronized (lock) {
-            if (!cut) {
-                // Automatically degrade.
-                cut = true;
+        if (!cut.get()) {
```
sczyh30: It's unnecessary to check here as the CAS will compare the old value.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
```diff
@@ -104,11 +103,11 @@ public DegradeRule setCount(double count) {
         return this;
     }
 
-    public boolean isCut() {
+    public AtomicBoolean getCut() {
```
sczyh30: Here it's better to be compatible with the original method. You can wrap AtomicBoolean with the original method.

```suggestion
    public boolean isCut() {
        return cut.get();
    }
```

Also the getter/setter method may not necessary to be `public`. For test we can just make it package-private.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
ATAXGT: OK,Sorry

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
ATAXGT: I think if this property is true, then there is no need to perform CAS operations because CAS also affects performance.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
```diff
@@ -80,7 +81,7 @@ public DegradeRule(String resourceName) {
      */
     private int grade = RuleConstant.DEGRADE_GRADE_RT;
 
-    private volatile boolean cut = false;
+    private volatile AtomicBoolean cut = new AtomicBoolean(false);
```
sczyh30: It's can be declared `final`.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
```diff
@@ -104,11 +103,11 @@ public DegradeRule setCount(double count) {
...
     }
 
-    public boolean isCut() {
-        return cut;
+    private boolean isCut() {
+        return cut.get();
     }
 
-    private void setCut(boolean cut) {
+    private void setCut(AtomicBoolean cut) {
```
sczyh30: Please also keep the original signature like:

```suggestion
    private void setCut(boolean b) {
        this.cut.set(b);
    }
```

sczyh30:
Thanks for contributing!

PR #543, opened 2019-03-05
Add exceptionsToTrace configuration support in @SentinelResource annotation

### LINKED ISSUE #540: Add configuration of exception classes in degrade rule
In some scenarios，this resource will be degraded when it throws a specific exception. 
So the configuration of exception class which should be traced (blocked), is needed.
For example:
  `
    private static final String RES_NAME = "RedisLockDegrade";

    @Resource
    private DsLock dsLock;

    @SentinelResource(value = RES_NAME, entryType = EntryType.IN, fallback = "fallback",
            exceptionClasses = {RedisLockException.class})
    public String execute(String arg) {
        return doBiz(arg);
    }

    /**
     * fallback
     * 降级函数，参数与原函数一致.
     */
    public String fallback(String arg) {
        dsLock.switchDbLock();
        try {
            System.out.println("Redis lock is degraded, use db lock instead");
            return doBiz(arg);
        } finally {
            dsLock.switchRedisLock();
        }
    }
`

beston123:
In some scenarios，this resource will be degraded when it throws a specific exception.
So the configuration of exception class that should be traced (blocked), is needed.

Fixes #540 

### Describe how you did it
Add the configuration of exception classes that would be traced.

### Describe how to verify it
None

### Special notes for reviews
None

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
@@ -15,13 +15,9 @@
  */
 package com.alibaba.csp.sentinel.annotation;
 
-import java.lang.annotation.ElementType;
-import java.lang.annotation.Inherited;
-import java.lang.annotation.Retention;
-import java.lang.annotation.RetentionPolicy;
-import java.lang.annotation.Target;
-
 import com.alibaba.csp.sentinel.EntryType;
+
+import java.lang.annotation.*;
 
 /**
  * The annotation indicates a definition of Sentinel resource.
@@ -63,4 +59,9 @@ public @interface SentinelResource {
      * @return name of the fallback function, empty by default
      */
     String fallback() default "";
+
+    /**
+     * @return the exception classes to trace, Throwable.class by default
+     */
+    Class<? extends Throwable>[] exceptionClasses() default {Throwable.class};
 }
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
@@ -63,5 +63,5 @@ public @interface SentinelResource {
     /**
      * @return the exception classes to trace, Throwable.class by default
      */
-    Class<? extends Throwable>[] exceptionClasses() default {Throwable.class};
+    Class<? extends Throwable>[] exceptionsToTrace() default {Throwable.class};
 }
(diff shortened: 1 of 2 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
```diff
@@ -63,4 +59,9 @@
      * @return name of the fallback function, empty by default
      */
     String fallback() default "";
+
+    /**
+     * @return the exception classes to trace, Throwable.class by default
+     */
+    Class<? extends Throwable>[] exceptionClasses() default {Throwable.class};
```
sczyh30: `exceptionClasses` might not be clear enough. Maybe it's better to be `exceptionsToTrace`?

beston123: Agree with your proposal.

sentinel-extension/sentinel-annotation-aspectj/src/main/java/com/alibaba/csp/sentinel/annotation/aspectj/SentinelResourceAspect.java
```diff
@@ -59,12 +59,33 @@ public Object invokeResourceWithSentinel(ProceedingJoinPoint pjp) throws Throwab
...
+     * @param ex
+     * @param exceptionsToTrace
+     * @return
+     */
+    private boolean isTrackedException(Throwable ex, Class<? extends Throwable>[] exceptionsToTrace) {
+        if (exceptionsToTrace == null) {
+            return false;
+        }
+        for (Class exceptionToTrace : exceptionsToTrace) {
+            if (ex.getClass().isAssignableFrom(exceptionToTrace)) {
```
sczyh30: It's incorrect. For each exception to trace, it should be:

```suggestion
            if (exceptionToTrace.isAssignableFrom(ex.getClass())) {
```

**Please be sure to add test cases and run the demo to verify it.**

sentinel-extension/sentinel-annotation-aspectj/src/main/java/com/alibaba/csp/sentinel/annotation/aspectj/SentinelResourceAspect.java
```diff
@@ -59,12 +59,33 @@ public Object invokeResourceWithSentinel(ProceedingJoinPoint pjp) throws Throwab
         } catch (BlockException ex) {
             return handleBlockException(pjp, annotation, ex);
         } catch (Throwable ex) {
-            Tracer.trace(ex);
+            if (isTrackedException(ex, annotation.exceptionsToTrace())) {
```
sczyh30: Maybe we can move the logic to `AbstractSentinelAspectSupport` class to support extension for sub-classes. E.g.:

```java
protected void traceException(Throwable ex, SentinelResource annotation)
```

sczyh30:
Thanks for contributing. I'll help to update the test cases and refactor code later. Please note that every new feature or enhancement **must** carry corresponding test cases.

sczyh30:
Improved in https://github.com/alibaba/Sentinel/commit/61c525076fb057c73a18f3b3ea00ef738fc9c839

PR #546, opened 2019-03-05
Solve the issue sleeping based tests are not stable

### LINKED ISSUE #541: LeapArrayTest.testGetValidHead fails randomly
## Issue Description

LeapArrayTest.testGetValidHead fails randomly

A case should get the same result always.

Try to run it repeatedly.

jasonjoo2010:

By mock the class `TimeUtil` we can precisely control the `sleep` delay.

Not only the tests become stable but they would take less time to finish.

Resolves #541

### DIFF
--- a/sentinel-core/pom.xml
+++ b/sentinel-core/pom.xml
@@ -39,6 +39,14 @@
             <groupId>org.hamcrest</groupId>
             <artifactId>java-hamcrest</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-module-junit4</artifactId>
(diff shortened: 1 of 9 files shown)


sentinel-core/pom.xml
```diff
@@ -39,6 +39,14 @@
             <groupId>org.hamcrest</groupId>
             <artifactId>java-hamcrest</artifactId>
         </dependency>
+        <dependency>
+        	<groupId>org.powermock</groupId>
+        	<artifactId>powermock-module-junit4</artifactId>
```
sczyh30: Maybe this should be `test` scope?

sczyh30:

PR #559, opened 2019-03-08
Dashboard: fix a request mapping issue of removing machines api

jasonjoo2010:

The api `{app}/machine/remove.json` is called through GET in old version but it changed to POST now and definition of the mapping was not changed together.


sczyh30:
Thanks

PR #574, opened 2019-03-14
Apply p3c pmd plugin/rules and fix/ignore all violations of priority 1

blindpirate:
### Describe what this PR does / why we need it

This PR applies alibaba p3c pmd plugin/rules and fixes all violations of priority 1.

### Does this pull request fix one issue?

Fixes #573 

### Describe how you did it

Just fix all violations.

### Describe how to verify it

 ```
mvn clean install -DskipTests -DminimumPriority=1
```

PR #582, opened 2019-03-16
Use RetryRule to mitigate test flakiness

blindpirate:

In my previous https://github.com/alibaba/Sentinel/pull/574 , I triggered travis build 3 times to get a successful result. This is too painful. This PR introduces `RetryRule` to retry tests in order to mitigate the test flakiness.

This is not a perfect solution (which should be rewriting the tests), but it's better than none.

A more modern alternative would be switching to JUnit 5 and using https://github.com/artsok/rerunner-jupiter .

NONE

### Describe how you did it

Create a `RetryRule`, which reruns failed test cases.

Tests for `RetryRule` are also added.

### Describe how to verify it

Run travis build many times. There should be no flakiness anymore.

### Special notes for reviews

Note: During retry, `Before` methods will be invoked, but the test class instance won't be reset.


sentinel-core/src/test/java/com/alibaba/csp/sentinel/slots/statistic/metric/OccupiableBucketLeapArrayTest.java
```diff
@@ -18,13 +20,16 @@
  * @author jialiang.linjl
  */
 public class OccupiableBucketLeapArrayTest {
```
blindpirate: I added `@Retry` to all test cases of this class since I observed flaky failures in travis build history. I assume this is a common issue of all test cases in this class.

blindpirate:
Hmmmm... seems not work well: https://travis-ci.org/alibaba/Sentinel/jobs/507205057

sczyh30:
That's cool for auto-retrying unstable tests. Thanks! We might need to rewrite some of these tests later to be stable and time-independent (though it's tricky).

Issue #580, opened 2019-03-16
Supplement and improvement for test cases


sczyh30:
## Issue Description

Type: *proposal (call for all)*

### Describe what happened (or what feature you want)

In order to guarantee the code quality and avoid potential bugs, we now call for all developers to improve tests of Sentinel. Any kinds of enhancement is welcomed, including (but not limited to):

- add more test cases (including unit test and integration test) 
- improve existing tests

We use [codecov](https://codecov.io/gh/alibaba/Sentinel) to measure the test coverage of the project. Everyone can refer to [here](https://codecov.io/gh/alibaba/Sentinel) to pick modules with low-coverage and add test cases for it.

Let's hack to avoid potential bugs and improve our code quality!

PR #598, opened 2019-03-20
Add test cases for different ProcessorSlot

cdfive:

Add some test cases for different ProcessorSlot, `NodeSelectorSlot`,`ClusterBuilderSlot` and so on.

Related to #580

### Describe how you did it

* Add `testFireEntry` method for each Slot to verify `fireEntry` method has been called in `entry` method,
and called only once.

* Add `testFireExit` method for each Slot to verify `fireExit` method has been called in `exit` method,
and called only once.

* Add some test cases to test the interactions in `entry` method, by mocking and verify the interaction
of `Node` and `Slot`.

### Describe how to verify it

Run the test cases.

### Special notes for reviews
The `ClusterNodeBuilderTest` is renamed to `ClusterBuilderSlotTest` for more standard test Class name.

The `DegradeSlot` and `SystemSlot` using static method to check the rules:
`DegradeRuleManager.checkDegrade(resourceWrapper, context, node, count);`
`SystemRuleManager.checkSystem(resourceWrapper);`,
while `FlowSlot` and `AuthoritySlot` call the internal method.

Maybe they can be unified in the furture.
I think using static method in `XxxRuleManager` is better, since it makes the slot class more testable, 
and we can write separate test cases for `XxxRuleManager`.
This is a suggestion,  they're not modified yet.


sczyh30:
Thanks for contributing.

> Maybe they can be unified in the future.
> I think using static method in XxxRuleManager is better, since it makes the slot class more testable,
and we can write separate test cases for XxxRuleManager.

IMHO, `XxxRuleManager` is used for managing rules (update / get rules), not for rule checking. In legacy design, the rule checking logic is located in `Rule#passCheck` method, which will be deprecated and removed later. Maybe it's better to move all `checkXxx` methods to each slot? Discussions are welcomed.

cdfive:
> XxxRuleManager is used for managing rules (update / get rules), not for rule checking
Maybe it's better to move all checkXxx methods to each slot.

Oh, I see. Previously I just found the implement way is different in four slots, and thought unified may be better.
 
I agree with you, the `XxxRuleManager` is used for managing rules (update / get rules)，as the class name indicates its responsibility.

How about move `checkXxx` method in `XxxRuleChecker` class, 
like `FlowRuleChecker` and `FlowRuleCheckerTest`.

In slot `entry` method, just call `XxxRuleChecker.checkXxx()` and `fireEntry`,  make the slot code simple and clear.
Then in `XxxRuleChecker.checkXxx()` method, call `XxxRuleManager` to get all rules and do checkRule logic for each rule, and we can add separate test cases for `XxxRuleChecker` and `XxxRuleManager`.

sczyh30:
Could you please resolve the conflicts? :)

Issue #618, opened 2019-03-26
[bug][transport-netty-http]setRules will not receive rules correctly when rules is longer then url's limitation


jasonjoo2010:
## Issue Description

When rules grow modification requests through transport module will fail.

### Describe what happened (or what feature you want)
HTTP 400 occurred

### Describe what you expected to happen
Rules can be saved successfully.

### How to reproduce it (as minimally and precisely as possible)
Increasing the rules to the number 18 or more may reproduce the issue.

sczyh30:
Resolved in #620 and #667 by @jasonjoo2010

PR #620, opened 2019-03-27
Support HTTP POST request for update operations in SentinelApiClient of Sentinel dashboard

jasonjoo2010:
`Feature`: Make transport-netty-http/transport-simple-http support posting (split to #620 )
`Restructure`: SentinelApiClient to support posting and make it simple
`Enhancement`: SentinelVersion parsing logic
`Bug fix`: comparation bug of SentinelVersion

Relevant issue #618

### Describe how you did it
Make `netty-http` support both `form-encoded` / `multipart` post requests.
Make `simple-http` support `form-encoded`. `Multipart` requests can also be support too if necessary.

`Form-encoded` would be the standard support for transport modules.

> These have been split to #620 

### Describe how to verify it
Submit big rules collection. For example more than 30 entries which will overflow the URI limitation in RFC.


sczyh30:
Good job. I'll review these days :)

sczyh30:
I've reviewed the code about `setRules` in `SentinelApiClient` and summarized the following compatibility table:

| | transport server < 1.6.0 | transport server >= 1.6.0 |
| :---: | :---- | :---- |
| Dashboard < 1.6.0 | GET | GET |
| Dashboard >= 1.6.0 | GET | POST |

So for `setRules` API, every version of transport server and dashboard is compatible. Is it correct? @jasonjoo2010

jasonjoo2010:
> I've reviewed the code about `setRules` in `SentinelApiClient` and summarized the following compatibility table:
> 
> transport server < 1.6.0	transport server >= 1.6.0
> Dashboard < 1.6.0	GET	GET
> Dashboard >= 1.6.0	GET	POST
> So for `setRules` API, every version of transport server and dashboard is compatible. Is it correct? Just to confirm it :) @jasonjoo2010


> Not only for `setRules` but for All api with large data submitted like `setRules`

Yes. 
Transport layer will support parameters in both `GET` and `POST` methods from 1.6.0. 
While dashboard will only enable POST feature when the machine's transport >= 1.6.0.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/client/SentinelApiClient.java
```diff
@@ -121,19 +136,139 @@ protected boolean isRedirectable(final String method) {
...
-        final HttpGet httpGet = new HttpGet(uri);
-        httpClient.execute(httpGet, new FutureCallback<HttpResponse>() {
+        StringBuilder urlBuilder = new StringBuilder();
+        urlBuilder.append("http://");
+        urlBuilder.append(ip).append(':').append(port).append('/').append(api);
+        if (params == null || params.size() == 0) {
+            // No parameter, using GET
+            return executeCommand(new HttpGet(urlBuilder.toString()));
+        }
+        boolean supportPost = StringUtil.isNotEmpty(app) && Optional.ofNullable(appManagement.getDetailApp(app))
```
sczyh30: Can we add a flag in the parameter of `executeCommand` indicating whether the API is POST-based? Using null `app` might be confusing.

sczyh30:
And could you please split your commit into two parts, one about changes for the transport module, another for the dashboard. Commits about dashboard should be prefixed with `dashboard: `.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/client/SentinelApiClient.java
jasonjoo2010: Append a new parameter to indicate posting intent

sczyh30:
Thanks for contributing!

Issue #634, opened 2019-03-29
FileWritableDataSource :file:/xx/xx.jar!/BOOT-INF/classes!/FlowRule.json (no such file or directory)


dqing0:
## Issue Description

FileWritableDataSource ：file:/xx/xx.jar!/BOOT-INF/classes!/FlowRule.json (没有那个文件或目录)
at java.io.FileOutputStream.open0(Native Method)
	at java.io.FileOutputStream.open(FileOutputStream.java:270)
	at java.io.FileOutputStream.<init>(FileOutputStream.java:213)
	at java.io.FileOutputStream.<init>(FileOutputStream.java:162)
	at com.alibaba.csp.sentinel.datasource.FileWritableDataSource.write(FileWritableDataSource.java:73)
	at com.alibaba.csp.sentinel.command.handler.ModifyRulesCommandHandler.writeToDataSource(ModifyRulesCommandHandler.java:108)
	at com.alibaba.csp.sentinel.command.handler.ModifyRulesCommandHandler.handle(ModifyRulesCommandHandler.java:68)
	at com.alibaba.csp.sentinel.transport.command.http.HttpEventTask.run(HttpEventTask.java:89)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)

### Describe what happened (or what feature you want)
i wanted to load local FlowRule.json,when i run application with jar,i found the error.
i think the **new File(filepath)** method  dont work,when the file in jar

sczyh30:
Hi, please check whether your file path is correct.

dqing0:
file path is correct, when i use intellij ,it work.But ,when i package, use 'java -jar xx.jar',it faild.  i think the reason is that  the method **new File(path) ** cannot read the file in jar

sczyh30:
Maybe you can try this: https://stackoverflow.com/questions/20389255/reading-a-resource-file-from-within-jar

fangjian0423:
@sczyh30 

`FileRefreshableDataSource` needs `java.io.File` as constructor parameter and we couldn't get file in jar.

Maybe we need `java.io.InputStream` as the constructor parameter of `FileRefreshableDataSource`.

[...]

PR #638, opened 2019-03-30
Fix read file in jar 20190330 dq

dqing0:

Fixes https://github.com/alibaba/Sentinel/issues/634

### Describe how you did it
1.参考FileRefreshableDataSource，FileWritableDataSource实现专门读取和写入jar文件的方法
2.通过jdk的JarFile,和jarEntry读取jar包中的文件
3.能够读取jar包中的文件
4.jar包中的文件修改可以更新到内存
5.在控制台更改规则后，能修改本地jar文件中的值

为什么没有通过添加构造器的方法去获取呢？
因为jar文件和普通的文件的读取和写有一些的区别，特别是写，还有classLoader.getResourceAsStream()方法只能获取inputstream对象，只能读取内容，但是它的更新时间获取不到，会导致修改配置文件不能更新规则

### Describe how to verify it
1.运行JarFileDataSourceDemo，JarFileDataSourceInit 中的main方法
2.读取文件内容，修改jar包中的配置文件会更新到内存，在控制台更改规则能更新到本地jar包中的配置文件，
这些功能已验证

### Special notes for reviews
1.写jar包中的某个文件时，其他不用变更的文件也要重写一遍，这点很坑，但jdk提供的JarFile好像只能
这样做，不知道有没有更好的方法取重写jar中的文件


sczyh30:
Hi, thanks for contributing. One more thing to discuss: is it really necessary to modify the config file inside the jar? I don't think it makes sense to modify files inside the jar because once the jar is upgraded, the files would be lost. Maybe writing to outside files (current implementation) is more appropriate.

What do you think of it? @CarpenterLee @fangjian0423

dqing0:
> files
now,i put the file outside jar.   this way just ，for those who want to configure in jar.  because springboot recommended use jar to run app

dqing0:
And ,According to the original method， if the jar is upgraded, the files also would be lost ,because 
the configuration is  in target classes

fangjian0423:
Sorry for the late reply.

I agree with @sczyh30,  outside file is better than file in jar.

Sorry for the first time I said adding a new constructor in `FileRefreshableDataSource` because check outside file modified is different in jar. Create new datasource is better.

sczyh30:
IMHO, it makes no sense for *dynamic file in jar* as it's not expected to be changed (users cannot modify the file in jar manually). But reading config from the file in jar is okay. So just a `FileInJarReadableDataSource` might be enough (don't need to refresh).

dqing0:
> IMHO, it makes no sense for _dynamic file in jar_ as it's not expected to be changed (users cannot modify the file in jar manually). But reading config from the file in jar is okay. So just a `FileInJarReadableDataSource` might be enough (don't need to refresh).

so,i change my pullrequest,only keep FileInJarReadableDataSource ?

sczyh30:
> > IMHO, it makes no sense for _dynamic file in jar_ as it's not expected to be changed (users cannot modify the file in jar manually). But reading config from the file in jar is okay. So just a `FileInJarReadableDataSource` might be enough (don't need to refresh).
> 
> so,i change my pullrequest,only keep FileInJarReadableDataSource ?

Feel free to do that :)
We can assume that the config file in jar won't be changed, so it don't need to be refreshable.

dqing0:
> We can assume that the config file in jar won't be changed, so it don't need to be refreshable.

ok,i change the pr

PR #645, opened 2019-04-02
fix CountDownLatch wait timeout in MetricFetcher when machine is dead

luoxn28:

in MetricFetche, if machine is dead, CountDownLatch will wait for timeout.

### Describe how you did it
if machine is dead, invoke latch.countDown()


sczyh30:
Thanks for contributing!

PR #649, opened 2019-04-03
Fix issue#612: too many connection of zookeeper datasource

### LINKED ISSUE #612: too many connection of zookeeper datasource
## Issue Description

Type: bug report

zookeeper server connections has increased five-fold 

4 rules one connection enough

1. 4 rules enabled
2. telnet zookeeper.host 2181
3. stat

zhaixiaoxiang:
 Fix "too many connection of zookeeper datasource"

Fix #612

### Describe how you did it
change zkClient to static in class ZookeeperDataSource

### Describe how to verify it
1. 5 rules enabled
2. echo cons | nc zookeeper.host 2181
3. check if there only exists one connection to zkServer


sczyh30:
@zhousiliang163 Would you like to take a review and give some suggestions?

sentinel-extension/sentinel-datasource-zookeeper/src/main/java/com/alibaba/csp/sentinel/datasource/zookeeper/ZookeeperDataSource.java
```diff
@@ -39,7 +39,7 @@
     private NodeCacheListener listener;
     private final String path;
 
-    private CuratorFramework zkClient = null;
+    private static CuratorFramework zkClient = null;
```
cdfive: Use upper case for static variable is better for readability.
This way is not thread-safe, use `volatile` and `synchronized` the same as singleton pattern may be better, 
but in most cases it's okay.

sczyh30:
> At present, a ZookeeperDataSource instance corresponds to a path. In design, as a caller, can we minimize the ZookeeperDataSource instance object? It is suggested that a new method can be added to register the monitored path, and multiple paths can share a ZookeeperDataSource instance object.

That's exactly what I was thinking of. Multiple paths of the same connection can be derived from the single data source.

cdfive:
@zhousiliang163 @sczyh30 
> multiple paths can share a ZookeeperDataSource instance object.

The paths are for rules of same type?
If same type, user can use json array and only have one connection.

There's a question that confuses me，if paths are for rules of different types, e.g. 
/Sentinel-Demo/SYSTEM-CODE-DEMO-FLOW
/Sentinel-Demo/SYSTEM-CODE-DEMO-DEGRADE
/Sentinel-Demo/SYSTEM-CODE-DEMO-SYSTEM
...

If add a new method to register the monitored path with a sharing ZookeeperDataSource instance object, but the parser and property field are in the super `AbstractDataSource` class, and one instance only have one parser and property.

sentinel-extension/sentinel-datasource-zookeeper/src/main/java/com/alibaba/csp/sentinel/datasource/zookeeper/ZookeeperDataSource.java
zhaixiaoxiang: This is not static final, so I think it should not use upper case.
I already use `synchronized` to ensure thread-safe, it's not necessary to use volatile

zhaixiaoxiang:
@sczyh30 @zhousiliang163 @cdfive 
I have tried to modify ZookeeperDataSource to support all path, but cause it implements `ReadableDataSource<S, T>`, this interface semantics is one input and one output, it's hard to modify ZookeeperDataSource directly. So I add a new class to wrap all ZookeeperDataSources

sentinel-demo/sentinel-demo-zookeeper-datasource/src/main/java/com/alibaba/csp/sentinel/demo/datasource/zookeeper/ZookeeperDataSourceDemo.java
```diff
@@ -1,57 +1,73 @@
...
-    private static void loadRules2() {
+    private static void loadRules2() throws Exception {
 
         final String remoteAddress = "127.0.0.1:2181";
         // 引入groupId和dataId的概念，是为了方便和Nacos进行切换
         final String groupId = "Sentinel-Demo";
-        final String flowDataId = "SYSTEM-CODE-DEMO-FLOW";
-        // final String degradeDataId = "SYSTEM-CODE-DEMO-DEGRADE";
-        // final String systemDataId = "SYSTEM-CODE-DEMO-SYSTEM";
+        final String flowDataIdFlow = "SYSTEM-CODE-DEMO-FLOW";
```
cdfive: The variable name `flowDataId` is better for code uniformity, I'm sorry for obsession...

sentinel-extension/sentinel-datasource-zookeeper/src/main/java/com/alibaba/csp/sentinel/datasource/zookeeper/ZookeeperAllDataSource.java
```diff
@@ -0,0 +1,77 @@
...
+public class ZookeeperAllDataSource {
+    private Map<Class, ZookeeperDataSource> clazzDataSourceMap = Maps.newConcurrentMap();
+    private String serverAddr;
+
+    public ZookeeperAllDataSource serverAddr(final String serverAddr) {
+        this.serverAddr = serverAddr;
+        return this;
+    }
+
+    public <W,T> ZookeeperAllDataSource register(String path, Class<W> clazz, Converter<String, T> converter) {
```
cdfive: Since `T` must be `List<W>`,
`public <T> ZookeeperAllDataSource register(String path, Class<T> clazz, Converter<String, List<T> converter)`
is better, to prevement user from using wrong class type.
e.g. 
```java
zookeeperAllDataSource = new ZookeeperAllDataSource()
.serverAddr(remoteAddress)
.register(path1, FlowRule.class, new Converter<String, List<ParamFlowRule>>() {
	@Override
	public List<ParamFlowRule> convert(String source) {
		return JSON.parseObject(source, new TypeReference<List<ParamFlowRule>>() {});
	}
}
```

cdfive:
@zhaixiaoxiang Add a new class is feasible.  

Maybe we can just add the `register` method in the `ZookeeperDataSource` class.
The same way as `ZookeeperAllDataSource`, firstly init the zkClient with remoteAddress, then register path with creating new `ZookeeperDataSource` Object.

A single class seems to be more simply and clear.

zhousiliang163:
Similar problems exist for other data sources, such as FileRefreshableDataSource having multiple ScheduledExecutorServices that generate multiple threads. 
maybe only one thread is enough.

So can we consider changing  ReadableDataSource ? 
For example, adding "a ReadableDataSource instance  can correspond to multiple Sentinel Property" implementations.

zhaixiaoxiang:
> @zhaixiaoxiang Add a new class `ZookeeperAllDataSource` is feasible.
> 
> Maybe we can just add the `register` method in the `ZookeeperDataSource` class.
> The same way as `ZookeeperAllDataSource` does, firstly init the zkClient with remoteAddress, then register path by creating new `ZookeeperDataSource` Object.
> 
> A single class keeps simple and clear.

Good suggestion, I have tried to do this.

zhaixiaoxiang:
> Similar problems exist for other data sources, such as FileRefreshableDataSource having multiple ScheduledExecutorServices that generate multiple threads.
> maybe only one thread is enough.
> 
> So can we consider changing ReadableDataSource ?
> For example, adding "a ReadableDataSource instance can correspond to multiple Sentinel Property" implementations.

In ZookeeperDataSource, I have resolved multiple pool problems. If this modification way are accepted, I'd resolve other same problems by the same way in this PR(or a new PR).

zhousiliang163:
> 
> 
> > Similar problems exist for other data sources, such as FileRefreshableDataSource having multiple ScheduledExecutorServices that generate multiple threads.
> > maybe only one thread is enough.
> > So can we consider changing ReadableDataSource ?
> > For example, adding "a ReadableDataSource instance can correspond to multiple Sentinel Property" implementations.
> 
> In ZookeeperDataSource, I have resolved multiple pool problems. If this modification way are accepted, I'd resolve other same problems by the same way in this PR(or a new PR).



> 
> 
> > Similar problems exist for other data sources, such as FileRefreshableDataSource having multiple ScheduledExecutorServices that generate multiple threads.
> > maybe only one thread is enough.
> > So can we consider changing ReadableDataSource ?
> > For example, adding "a ReadableDataSource instance can correspond to multiple Sentinel Property" implementations.
> 
> In ZookeeperDataSource, I have resolved multiple pool problems. If this modification way are accepted, I'd resolve other same problems by the same way in this PR(or a new PR).



Current modifications, the relevant RuleManager and ZookeeperDataSource are highly coupled, inconvenient to maintain, and preferably decoupled.

zhaixiaoxiang:
> > > Similar problems exist for other data sources, such as FileRefreshableDataSource having multiple ScheduledExecutorServices that generate multiple threads.
> > > maybe only one thread is enough.
> > > So can we consider changing ReadableDataSource ?
> > > For example, adding "a ReadableDataSource instance can correspond to multiple Sentinel Property" implementations.
> > 
> > 
> > In ZookeeperDataSource, I have resolved multiple pool problems. If this modification way are accepted, I'd resolve other same problems by the same way in this PR(or a new PR).
> 
> > > Similar problems exist for other data sources, such as FileRefreshableDataSource having multiple ScheduledExecutorServices that generate multiple threads.
> > > maybe only one thread is enough.
> > > So can we consider changing ReadableDataSource ?
> > > For example, adding "a ReadableDataSource instance can correspond to multiple Sentinel Property" implementations.
> > 
> > 
> > In ZookeeperDataSource, I have resolved multiple pool problems. If this modification way are accepted, I'd resolve other same problems by the same way in this PR(or a new PR).
> 
> Current modifications, the relevant RuleManager and ZookeeperDataSource are highly coupled, inconvenient to maintain, and preferably decoupled.

If decoupled, developers need to call all register2Property methods themselves?

zhousiliang163:
> 
> 
> > > > Similar problems exist for other data sources, such as FileRefreshableDataSource having multiple ScheduledExecutorServices that generate multiple threads.
> > > > maybe only one thread is enough.
> > > > So can we consider changing ReadableDataSource ?
> > > > For example, adding "a ReadableDataSource instance can correspond to multiple Sentinel Property" implementations.
> > > 
> > > 
> > > In ZookeeperDataSource, I have resolved multiple pool problems. If this modification way are accepted, I'd resolve other same problems by the same way in this PR(or a new PR).
> > 
> > 
> > > > Similar problems exist for other data sources, such as FileRefreshableDataSource having multiple ScheduledExecutorServices that generate multiple threads.
> > > > maybe only one thread is enough.
> > > > So can we consider changing ReadableDataSource ?
> > > > For example, adding "a ReadableDataSource instance can correspond to multiple Sentinel Property" implementations.
> > > 
> > > 
> > > In ZookeeperDataSource, I have resolved multiple pool problems. If this modification way are accepted, I'd resolve other same problems by the same way in this PR(or a new PR).
> > 
> > 
> > Current modifications, the relevant RuleManager and ZookeeperDataSource are highly coupled, inconvenient to maintain, and preferably decoupled.
> 
> If decoupled, developers need to call all register2Property methods themselves?



According to the single responsibility principle, it is recommended that ZookeeperDataSource not join register2Property


> 
> 
> > > > Similar problems exist for other data sources, such as FileRefreshableDataSource having multiple ScheduledExecutorServices that generate multiple threads.
> > > > maybe only one thread is enough.
> > > > So can we consider changing ReadableDataSource ?
> > > > For example, adding "a ReadableDataSource instance can correspond to multiple Sentinel Property" implementations.
> > > 
> > > 
> > > In ZookeeperDataSource, I have resolved multiple pool problems. If this modification way are accepted, I'd resolve other same problems by the same way in this PR(or a new PR).
> > 
> > 
> > > > Similar problems exist for other data sources, such as FileRefreshableDataSource having multiple ScheduledExecutorServices that generate multiple threads.
> > > > maybe only one thread is enough.
> > > > So can we consider changing ReadableDataSource ?
> > > > For example, adding "a ReadableDataSource instance can correspond to multiple Sentinel Property" implementations.
> > > 
> > > 
> > > In ZookeeperDataSource, I have resolved multiple pool problems. If this modification way are accepted, I'd resolve other same problems by the same way in this PR(or a new PR).
> > 
> > 
> > Current modifications, the relevant RuleManager and ZookeeperDataSource are highly coupled, inconvenient to maintain, and preferably decoupled.
> 
> If decoupled, developers need to call all register2Property methods themselves?

According to the single responsibility principle, it is recommended that ZookeeperDataSource not make register2Property calls.
If  want to integrate all register2Property calls, can we do this by building an integration tool class?

zhaixiaoxiang:
@zhousiliang163 
I  call `FlowRuleManager.register2Property(property);`  and all other `XxxRuleManager.register2Property(property);` directly.

This is actually highly coupled, for example, someone add a new YyyRule and YyyRuleManager some day and he forgets(or don't know) to add them to ZookeeperDataSourceBuilder.build(), there will be errors happened.

So if we don't call XxxRuleManager.register2Property directly, but use reflect, how do you think ?
image

image

zhousiliang163:
> 
> 
> @zhousiliang163
> I call `FlowRuleManager.register2Property(property);` and all other `XxxRuleManager.register2Property(property);` directly.
> 
> This is actually highly coupled, for example, someone add a new YyyRule and YyyRuleManager some day and he forgets(or don't know) to add them to ZookeeperDataSourceBuilder.build(), there will be errors happened.
> 
> So if we don't call XxxRuleManager.register2Property directly, but use reflect, how do you think ?





In this way, although the coupling degree is reduced, it is not completely decoupled. It is suggested that it can be extended from ReadableDataSource.as follows：
图片

zhaixiaoxiang:
@zhousiliang163 

I modify using the following structure (I will modify other datasource later if this modification is OK).

image

zhaixiaoxiang:
@zhousiliang163 Any idea about this PR?

sentinel-extension/sentinel-datasource-zookeeper/src/main/java/com/alibaba/csp/sentinel/datasource/zookeeper/ZookeeperDataSource.java
```diff
@@ -1,168 +1,275 @@
...
-                        RecordLog.info(String.format("[ZookeeperDataSource] New property value received for (%s, %s): %s",
-                                serverAddr, path, newValue));
-                        // Update the new value to the property.
-                        getProperty().updateValue(newValue);
-                    } catch (Exception ex) {
-                        RecordLog.warn("[ZookeeperDataSource] loadConfig exception", ex);
-                    }
-                }
-            };
+    private synchronized void initPoolAndClient(final String serverAddr, final List<AuthInfo> authInfos) {
```
zhousiliang163: Synchronization is not recommended on Methods.

PR #714, opened 2019-04-28
dashboard:add Zookeeper flow rule publisher sample

kexianjun:
add zookeeper flow rule publisher sample for dashboard


sczyh30:
Thanks for contributing!

PR #716, opened 2019-04-28
remove unnecessary init logic in the ZookeeperDataSource

kexianjun:
remove unnecessary init logic


sczyh30:
Hi, thanks for contributing. Could you please illustrate your changes?

kexianjun:
> Hi, thanks for contributing. Could you please illustrate your changes?

the main logic of loadInitialConfig method is the same as the NodeCacheListener, as following
```
this.listener = new NodeCacheListener() {
                @Override
                public void nodeChanged() {

                    try {
                        T newValue = loadConfig();
                        RecordLog.info(String.format("[ZookeeperDataSource] New property value received for (%s, %s): %s",
                                serverAddr, path, newValue));
                        // Update the new value to the property.
                        getProperty().updateValue(newValue);
                    } catch (Exception ex) {
                        RecordLog.warn("[ZookeeperDataSource] loadConfig exception", ex);
                    }
                }
            };
```
so when call loadInitialConfig in com.alibaba.csp.sentinel.datasource.zookeeper.ZookeeperDataSource#init method, it will make the properties update twice.Sometimes in my test,the flowRule or other config is null in the end.And when I remove loadInitialConfig method, it is always worked fine for me.

PR #723, opened 2019-04-30
use calculateTimeIdx to calculate idx

kexianjun:


sczyh30:
Thanks for contributing!

PR #730, opened 2019-05-05
[Feature]Add an extension to Sentinel internal statistics

CarpenterLee:
Sentient has various flow control strategies, all these strategies are based on Sentinel internal statistics. The real-time statistics are also useful for system monitoring. Unfortunately, Sentinel does not provide an extension to obtain these data directly.

This PR provides an extension to Sentinel internal statistics.

NONE
### Describe how you did it
Add `MetricExtension` interface and related SPI classes as an extension to Sentinel internal statistics. This extension provides callbacks when a request *passes* rule checking, *blocked* by flow control, *successfully* end or *exception* occurred. Accompanied by these events, *response time* and current *thread number* will be recorded too.

### Describe how to verify it
Run test cases.


sentinel-core/src/main/java/com/alibaba/csp/sentinel/metric/extension/MetricExtensionInit.java
```diff
@@ -0,0 +1,37 @@
...
+import com.alibaba.csp.sentinel.init.InitFunc;
+import com.alibaba.csp.sentinel.log.RecordLog;
+
+/**
+ * Get all {@link MetricExtension}s via SPI.
+ *
+ * @author Carpenter Lee
+ * @since 1.6.1
+ */
+public class MetricExtensionInit implements InitFunc {
```
sczyh30: It's better to be a `MetricExtensionProvider` based on `SpiLoader`.

Issue #736, opened 2019-05-07
Is it necessary to do strong check on fastjson?


jasonjoo2010:
## Issue Description
Rules' deserialization will be interrupted with old version of fastjson due to the bug on inherited bean classes.
Should we add some kinds of force check on it?

sczyh30:
Yes, it's really necessary as the deserialization problem of old fastjson versions often makes users confusing. Maybe we can add checking logic in relevant command handler and carry the error info in the command response, so that dashboard can get the error status. Any suggestions?

jasonjoo2010:
The issue doesn't only occurred when performing operations in `dashboard` but also loading rules from storages into application. I think putting it into sentinel initializing process is efficient because there're no such common suitable place in `sentinel-datasource-extension` etc. To prevent unused dependency we can use reflecting.

What do you think about?

And more i lock the minor version by tests `fastjson-1.2.12`

jasonjoo2010:
> > The issue doesn't only occur during operations in `dashboard` but also loading rules from storages into application. I think putting it into sentinel initializing process is efficient because there're no such common suitable place in `sentinel-datasource-extension` etc. To prevent unused dependency we can use reflecting.
> > What do you think about?
> > And more i lock the minor version by tests `fastjson-1.2.12`
> 
> That's okay, but just warn in the log might not be enough. We may also need to modify the rule command handlers in transport module to carry the error info in the command response.
> 
> PS: `fastjson` is introduced in transport common module, so maybe we can put the `InitFunc` to this module? (this cannot cover the users that don't use transport module, but i think it's okay, as the JSON library and related logic are not necessary in `sentinel-core`).

Indeed `sentinel-core` doesn't need fastjson library. The bug occurs when do deserialization as figure:

<img width="634" alt="Screen Shot 2019-05-09 at 09 22 17" src="https://user-images.githubusercontent.com/17573073/57421324-e611ab80-723d-11e9-83fd-a1857b63284e.png">

And as we know the effected functions are listed below:

<img width="289" alt="Screen Shot 2019-05-09 at 09 34 09" src="https://user-images.githubusercontent.com/17573073/57421350-004b8980-723e-11e9-84fd-ae61fe4c6264.png">

<img width="495" alt="Screen Shot 2019-05-09 at 09 34 15" src="https://user-images.githubusercontent.com/17573073/57421357-03467a00-723e-11e9-965c-13ef4b027488.png">

Which are pointed out in red.

[...]

PR #742, opened 2019-05-08
Stop initializing with old fastjson (earlier than 1.2.12)

jasonjoo2010:
Related to issue #736 
Many guys have been buried here and maybe it's a good idea to do it aggressively.

### Describe how you did it
Throw an exception when initializing.

### Describe how to verify it
Change your project's dependency

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/init/InitExecutor.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/init/InitExecutor.java
@@ -15,12 +15,14 @@
  */
 package com.alibaba.csp.sentinel.init;
 
+import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.ServiceLoader;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import com.alibaba.csp.sentinel.log.RecordLog;
+import com.alibaba.csp.sentinel.util.VersionUtil;
 
 /**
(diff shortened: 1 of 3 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/init/InitExecutor.java
```diff
@@ -31,6 +33,26 @@
 
     private static AtomicBoolean initialized = new AtomicBoolean(false);
 
+    /**
+     * Make sure the fastjson (if have one) no more less than 1.2.12
+     */
+    private static void checkFastjson() {
```
sczyh30: Maybe checking here is not appropriate? This can be implemented as a single `InitFunc`. And I think it should not block the initialization even if the fastjson version is low (the behavior might be weird for users). We may need other approaches to inform the users.

sczyh30:
Hi, any thoughts about this?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/init/InitExecutor.java
jasonjoo2010: Discussion is under #736 and what's your opinion for that? Let's synchronize it.

jasonjoo2010:
> Hi, any thoughts about this?

Don't to forget to take a look at #736 due to i quoted a wrong person in previous reply.

jasonjoo2010:
Close via #1377

PR #758, opened 2019-05-13
Improve the parameter flow module and refactor API gateway adapter common module

### LINKED ISSUE #756: Improvement for API gateway adapter common module
## Issue Description

Type: *feature request*

In current implementation, the `GatewayRuleManager` will convert `GatewayFlowRule` to `FlowRule` and `ParamFlowRule`, then load the rules with the respective rule manager. This will flush away the previous rules that users defined.

We might need to separate the converted gateway rule storage and checking from the original *flow* and *param* module.

Besides, since 1.6.0, the `ParamFlowRule` supports richer attributes than `FlowRule` (e.g. statistic interval per rule, burst count). Maybe we can convert the non-param gateway rules to "fake" parameter rules with the constant parameter passed, which can achieve the effect of normal flow control.

sczyh30:

Improve the parameter flow module and refactor the API gateway adapter common module. See #756 for details.

Resolves #756 

### Describe how you did it

- Refactor `FlowRuleChecker` to improve code reuse. Add `checkFlow` method to support applying rule retrieving function and then check.
-  Separate parameter metric storage from `ParamFlowSlot` and improve `ParamFlowRuleUtil`
  -  Add a `ParameterMetricStorage` specific for caching ParameterMetric (moved from ParamSlot)
  -  Add rule map building helper method in `ParamFlowRuleUtil` so that we can reuse it in other rule managers
- Refactor API gateway common module to separate converted rules from other rule managers
  - Separate converted parameter rules from ParamFlowManager. Now the converted rules will be kept in `GatewayRuleManager` directly.
  - Add a `GatewayFlowSlot` to do separate flow checking for generated rules.
  - Refactor rule converting mechanism: now non-param gateway rules (normal mode) will also be converted to parameter flow rules. The index will be the las position. In `GatewayParamParser` we put **a constant value** to the last position to achieve the effect of non-param flow control.

### Describe how to verify it

Run the test cases and demo.

### Special notes for reviews

Please take care of the *internal breaking changes*.

### DIFF
--- a/sentinel-adapter/sentinel-api-gateway-adapter-common/src/test/java/com/alibaba/csp/sentinel/adapter/gateway/common/param/GatewayParamParserTest.java
+++ b/sentinel-adapter/sentinel-api-gateway-adapter-common/src/test/java/com/alibaba/csp/sentinel/adapter/gateway/common/param/GatewayParamParserTest.java
@@ -70,10 +70,15 @@ public class GatewayParamParserTest {
             .setCount(5)
             .setIntervalSec(1)
         );
+        rules.add(new GatewayFlowRule(routeId1)
+            .setCount(10)
+            .setControlBehavior(2)
+            .setMaxQueueingTimeoutMs(1000)
+        );
         GatewayRuleManager.loadRules(rules);
 
         Object[] params = parser.parseParameterFor(routeId1, request, routeIdPredicate);
-        assertThat(params.length).isZero();
+        assertThat(params.length).isEqualTo(1);
     }
 
     @Test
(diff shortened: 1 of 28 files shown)


sentinel-adapter/sentinel-api-gateway-adapter-common/src/test/java/com/alibaba/csp/sentinel/adapter/gateway/common/param/GatewayParamParserTest.java
```diff
@@ -0,0 +1,205 @@
...
+        Set<GatewayFlowRule> rules = new HashSet<>();
+        String routeId1 = "my_test_route_A";
+        rules.add(new GatewayFlowRule(routeId1)
+            .setCount(5)
+            .setIntervalSec(1)
+        );
+        GatewayRuleManager.loadRules(rules);
+
+        Object[] params = parser.parseParameterFor(routeId1, request, routeIdPredicate);
+        assertThat(params.length).isZero();
```
CarpenterLee: This test case is for testing normal flow rules, and the `params.length` seems should be one, which is `SentinelGatewayConstants.GATEWAY_DEFAULT_PARAM`

sczyh30: Fixed, plz review.

Issue #673, opened 2019-04-16
Make the value of RT_MAX_EXCEED_N in DegradeRule configurable


cdfive:
## Issue Description

In the process of degrade,  when resource enters a quasi-degraded state, the max count (the value of `RT_MAX_EXCEED_N` in `DegradeRule` class) which statistics the next coming requests passed now is fixed value: **5**.

Users in community are wondering that why the value is fixed.

### Describe what happened (or what feature you want)

Users may expect that the `RT_MAX_EXCEED_N` can be configured in different scenes.

How about  adding a property like `csp.sentinel.degrade.rt.max.exceed.count`? 
and does it need to be modified dynamically?

### Anything else we need to know?

Related to #154

sczyh30:
I think this can be different per rules. We may need to add a new field for this, but this should be carefully designed, as this `RT_MAX_EXCEED_N` has multiple usages actually. In exception ratio mode, if total QPS < `RT_MAX_EXCEED_N`, even if the exception ratio exceeded the threshold, the request would pass.

sczyh30:
Resolved in #789. It's now configurable as two attributes:

- `rtSlowRequestAmount` for the minimum number of consecutive slow requests that can trigger RT circuit breaking.
- `minRequestAmount` for the minimum number of requests (in an active statistic time span) that can trigger circuit breaking.

PR #789, opened 2019-05-25
Fix  issue 673 Make the value of RT_MAX_EXCEED_N in DegradeRule configurable 

linlinisme:

 Make the value of RT_MAX_EXCEED_N in DegradeRule configurable 

fix https://github.com/alibaba/Sentinel/issues/673

### Describe how you did it

DegradeRule  add a new filed rtMaxExceedN to replace RT_MAX_EXCEED_N

### Describe how to verify it

Configure rtMaxExceedN to see if it takes effect.


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
```diff
@@ -55,7 +55,7 @@
  */
 public class DegradeRule extends AbstractRule {
 
-    private static final int RT_MAX_EXCEED_N = 5;
+    private int rtMaxExceedN = RuleConstant.DEGRADE_GRADE_RT_MAX_EXCEED_N;
```
sczyh30: The `RT_MAX_EXCEED_N` has two meanings actually:

- For RT mode, it indicates the minimum number of consecutive slow requests that can trigger RT circuit breaking.
- For exception ratio mode, it indicates the minimum number of requests (in an active statistic time span) that can trigger circuit breaking.

So maybe we need two attributes here? And the attribute name might need to be discussed.

linlinisme: hi, @sczyh30 
```
        DegradeRule rule = new DegradeRule();
        rule.setResource(KEY);
        // set limit exception ratio to 0.1
        rule.setCount(0.1);
        rule.setGrade(RuleConstant.DEGRADE_GRADE_EXCEPTION_RATIO);
        rule.setTimeWindow(10);
        rule.setRtMaxExceedN(20);
        rules.add(rule);
```
The downgrade rule can only be set to one mode at present. rt mode and exception ratio mode Only one can be selected. The RT_MAX_EXCEED_N attribute can represent different meanings in different modes. So I think that using only one attribute is ok. and you ?

sczyh30: They're indeed two different things, so only one attribute might be confusing. I think we can make it two attributes:

- `rtSlowRequestAmount` for minimum number of consecutive slow requests that can trigger RT circuit breaking.
- `minRequestAmount` for the minimum number of requests (in an active statistic time span) that can trigger circuit breaking.

What do you think of it? @CarpenterLee

sczyh30: Any progress?

linlinisme: two attribute may be better

sczyh30:
Could you please also update the test cases for these two new attributes?

linlinisme:
> Could you please also update the test cases for these two new attributes?

@sczyh30   Dear sczyh30.  I have updated test cases.  but i find another potential problem: may be the hashCode() method should be considered the problem of numerical out of bounds.

sczyh30:
> > Could you please also update the test cases for these two new attributes?
> 
> @sczyh30 Dear sczyh30. I have updated test cases. but I find another potential problem: maybe the `hashCode()` method should consider the problem of numerical out of bounds.

Maybe overflow isn't a problem for `hashCode()` as long as it could achieve good dispersion and uniqueness.

linlinisme:
> > > Could you please also update the test cases for these two new attributes?
> > 
> > 
> > @sczyh30 Dear sczyh30. I have updated test cases. but I find another potential problem: maybe the `hashCode()` method should consider the problem of numerical out of bounds.
> 
> Maybe overflow isn't a problem for `hashCode()` as long as it could achieve good dispersion and uniqueness.

ok, I see . Would you view what i had commited is ok or not?

sentinel-core/src/test/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeTest.java
```diff
@@ -83,17 +84,38 @@ public void testExceptionRatioModeDegrade() throws Throwable {
...
 
-        // Will fail.
+        // Will true. While totalQps < minRequestAmount
+        when(cn.totalQps()).thenReturn(8d);
+        assertTrue(rule.passCheck(context, node, 1));
+
+        // Will true. While totalQps > minRequestAmount and exceptionCount < minRequestAmount
+        when(cn.totalQps()).thenReturn(21d);
+        when(cn.exceptionQps()).thenReturn(19d);
+        assertTrue(rule.passCheck(context, node, 1));
```
sczyh30: There are some problems in the test case. Here `successQps` is not provided, so it's by default 0, which does not match the actual relation:

- in the same aligned statistic time window, success (aka. completed count) = exception count + non-exception count (realSuccess)

sentinel-demo/sentinel-demo-basic/src/main/java/com/alibaba/csp/sentinel/demo/degrade/ExceptionRatioDegradeDemo.java
```diff
@@ -113,8 +113,9 @@ private static void initDegradeRule() {
         rule.setResource(KEY);
         // set limit exception ratio to 0.1
         rule.setCount(0.1);
-        rule.setGrade(RuleConstant.DEGRADE_GRADE_EXCEPTION_RATIO);
+        rule.setGrade(RuleConstant.DEGRADE_GRADE_MIN_REQUEST_EXCEED_N);
```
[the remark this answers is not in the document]
linlinisme: > Typo?
Sorry, for my carelessness. It is typo.

sentinel-core/src/test/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeTest.java
linlinisme: > There are some problems in the test case. Here `successQps` is not provided, so it's by default 0, which does not match the actual relation:
> 
> * in the same aligned statistic time window, success (aka. completed count) = exception count + non-exception count (realSuccess)

Do you mean ?
```
            double realSuccess = success - exception;
            if (realSuccess <= 0 && exception < minRequestAmount) {
                return true;
            }
```
I am somewhat confused. My understanding is that success is counted when the request should respond successfully. The exception is counted when the program throws a unknown exception. Why do you need a realSuccess?  The number of success and excption should be independent of each other.

sczyh30: The name `successQps` might be confusing here. The `successQps` in Sentinel actually means **complete** (i.e. passed from Sentinel checking, and has finished invocation, no matter it succeeded or failed logically), so the *success count* actually includes the *exception count*. The real success count (what you've understood) should be `success - exception`.

linlinisme: understood, the latest code had commit.

sentinel-demo/sentinel-demo-basic/src/main/java/com/alibaba/csp/sentinel/demo/degrade/ExceptionRatioDegradeDemo.java
```diff
@@ -113,8 +113,9 @@ private static void initDegradeRule() {
         rule.setResource(KEY);
         // set limit exception ratio to 0.1
         rule.setCount(0.1);
-        rule.setGrade(RuleConstant.DEGRADE_GRADE_EXCEPTION_RATIO);
+        rule.setGrade(RuleConstant.DEGRADE_GRADE_EXCEPTION_COUNT);
```
sczyh30: Should be `DEGRADE_GRADE_EXCEPTION_RATIO`?

sczyh30:
Nice work, Thanks for contributing! 👍 
I'll refine the naming of some constants later.

PR #804, opened 2019-05-31
Fix issue62 : Support customized log directory and configuration properties directory 

### LINKED ISSUE #62: Support customized log directory and configuration properties directory
目前启动配置项如果走配置文件只能${home}/logs/csp/${project.name}.properties路径下，log文件输出也是在${home}/logs/csp/下面
1.看是否可以整合到具体项目配置文件中去
2.看是否支持自定义路径，配置文件和log文件路径等。
主要由于当前公司环境部署集中化部署，且多个环境在一个服务器。

linlinisme:

Resolves #62

### Describe how you did it

Split the configuration into two parts: the **log** and the **sentinel-core**. There are special tool classes to load the respective configurations. The tool class supports reading **specified files, default files, and virtual machine parameter configurations.**

the sentinel-core support configurations :
```
    public static final String APP_TYPE = "csp.sentinel.app.type";
    public static final String CHARSET = "csp.sentinel.charset";
    public static final String SINGLE_METRIC_FILE_SIZE = "csp.sentinel.metric.file.single.size";
    public static final String TOTAL_METRIC_FILE_COUNT = "csp.sentinel.metric.file.total.count";
    public static final String COLD_FACTOR = "csp.sentinel.flow.cold.factor";
    public static final String STATISTIC_MAX_RT = "csp.sentinel.statistic.max.rt";
```
Customize the loaded configuration file by configuring the `csp.sentinel.config.file` property.
for example `-Dcsp.sentinel.config.file=sentinel.properties`

the log:
```
    public static final String LOG_DIR = "csp.sentinel.log.dir";
    public static final String LOG_NAME_USE_PID = "csp.sentinel.log.use.pid";
    public static final String LOG_OUTPUT_TYPE = "csp.sentinel.log.output.type";
    public static final String LOG_CHARSET = "csp.sentinel.log.charset";
```
Customize the loaded configuration file by configuring the `csp.sentinel.config.file` property.
for example `-Dcsp.sentinel.config.file=sentinel.properties`

**Pay attention! the `log` and `sentinel-core `configuration share a same config file now**

Configure effective priority: **virtual machine parameters -> -Dcsp.sentinel.config.file specifies file configuration**

If `-Dcsp.sentinel.config.file` is not specified, the default value is `sentinel.properties`. That means you only need to add `sentinel.properties` to your project then you can load the custom configuration.

If you are the old version of the specified directory configuration: `${user.home}/logs/csp/appName.properties`
And you want to keep the existing configuration path, it is also ok. You don't need to configure `-Dcsp.sentinel.config.file` nor  need to add `sentinel.properties` to your project. Because it provides legacy compatibility.

### Describe how to verify it
Configure the attribute value and run the project to see if it works.
And you can run the  test cases in `SentinelConfigTest`、`LogBaseTest`、`ConfigUtilTest`


linlinisme:
All attribute configurations are supported, including log paths.
Support four configuration methods:
1. Traditional virtual machine parameters
2. Custom configuration file, virtual machine parameter specification: -Dcsp.sentinel.config.file = file
3. The configuration file in the project, the file name sentinel.properties. Remember to pack into the class path
4. If none of the above is configured, read the path specified in the old version csp.sentinel.log.dir

 The configuration takes effect as the same  above.

vitrun:
when will this be fixed ? we can not use it in production if custom configuration is unavailable.

sczyh30:
> when will this be fixed ? we can not use it in production if custom configuration is unavailable.

@vitrun I'll take a review these days. It's expected to be included in 1.7.0 (perhaps in the next few weeks). If you're using Spring Boot/Spring Cloud you could leverage Spring Cloud Alibaba to support configuring via `application.properties`: https://github.com/spring-cloud-incubator/spring-cloud-alibaba/wiki/Sentinel#more

vitrun:
> `application.properties`: https://github.com/spring-cloud-incubator/spring-cloud-alibaba/wiki/Sentinel#more

seems that `csp.sentinel.statistic.max.rt` is not present. For now, we set it when startup:
`
SentinelConfig.setConfig(SentinelConfig.STATISTIC_MAX_RT, "60000")
`

sczyh30:
> seems that `csp.sentinel.statistic.max.rt` is not present. For now, we set it when startup:
> `SentinelConfig.setConfig(SentinelConfig.STATISTIC_MAX_RT, "60000")`

You could open an issue in https://github.com/spring-cloud-incubator/spring-cloud-alibaba/issues

cc @fangjian0423 We can refer to https://github.com/alibaba/Sentinel/wiki/启动配置项

sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/ConfigLoadUtils.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.CopyOnWriteArraySet;
+
+/**
+ * @author lianglin
+ * @since 2019-05-31 15:19
```
sczyh30: `@since 1.7.0`?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/AppNameUtil.java
```diff
@@ -57,7 +55,9 @@ private AppNameUtil() {
 
     static {
         resolveAppName();
-        RecordLog.info("App name resolved: " + appName);
+        //here use RecordLog will lead to class-init circle-dependency problem
+        //RecordLog.info("App name resolved: " + appName);
+        System.out.println("App name resolved: " + appName);
```
sczyh30: Here we might need to print the `appName` to the log as it's significant information. That's why we had separated logging related properties from the `SentinelConfig`. Are there any solutions?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/ConfigLoadUtils.java
```diff
@@ -0,0 +1,151 @@
...
+        return properties;
+    }
+
+
+    private static void init() {
+
+        //Load from JVM -D user-specified file
+        String fileName = System.getProperty(SENTINEL_CONFIG_FILE);
+        if (StringUtil.isNotBlank(fileName)) {
+            if (fileName.startsWith("/")) {
```
sczyh30: What if the `fileName` is a relative path?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/ConfigLoadUtils.java
```diff
@@ -0,0 +1,151 @@
...
+                File file = new File(fileName);
+                if (file.exists()) {
+                    loadPropertiesFromFille(fileName);
+                }
+            }
+        }
+        //JVM parameter override file config
+        CopyOnWriteArraySet<Map.Entry<Object, Object>> entries = new CopyOnWriteArraySet<>(System.getProperties().entrySet());
+        for (Map.Entry entry : entries) {
+            properties.setProperty(entry.getKey().toString(), entry.getValue().toString());
```
sczyh30: We might also need to log here as before (if the value is replaced).

sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/ConfigLoadUtils.java
linlinisme: `                loadPropertiesFromResource(fileName);
`
deal with the relative path file

linlinisme: may be the method name make confused, i will rename later

sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/ConfigLoadUtils.java
linlinisme: Is better to add a separate log config file?  and the sentinel core config is another file

linlinisme: the config load order may be : log config -> sentinel-core config.

sczyh30: +1 for this

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/AppNameUtil.java
linlinisme: > Here we might need to print the `appName` to the log as it's significant information. That's why we had separated logging related properties from the `SentinelConfig`. Are there any solutions?

If the configuration is isolated, my idea is to add two new loader classes. `SentinelConfigLocator` is responsible for loading the Sentinel Core configuration, and `LogConfigLocator` is responsible for loading the Log configuration. `SentinelConfig` reads the configuration from `SentinelConfigLocator`, and `LogBase` reads the configuration from `LogConfigLocator`. This way they are completely independent, and the problem of class initialization interdependence is also solved. What needs to be considered now is whether the Log configuration and Sentinel Core are placed on the same configuration file? If so, `SentinelConfigLocator`, `LogConfigLocator` will read the same file. Is this ok?

linlinisme:
add  test cases

sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfigLocator.java
```diff
@@ -0,0 +1,96 @@
...
+ * @author lianglin
+ * @since 1.7.0
+ */
+public class SentinelConfigLocator {
+
+
+    private static final String DIR_NAME = "logs" + File.separator + "csp";
+    private static final String USER_HOME = "user.home";
+
+    private static String SENTINEL_CONFIG = "csp.sentinel.config.file";
```
sczyh30: Maybe this key can be public?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfigLocator.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * <p>
+ * class responsible for loading the Sentinel Core configuration
+ * </p>
+ *
+ * @author lianglin
+ * @since 1.7.0
+ */
+public class SentinelConfigLocator {
```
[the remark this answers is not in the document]
linlinisme: `Locator` it is the same meaning of `Loader`. Rename to `SentinelConfigLoader` it is also ok.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfigLocator.java
linlinisme: I make `SENTINEL_CONFIG` private modified , because unit  now it only be used in `SentinelConfigLocator.java`.  If it is accessed anywhere else in the future, it really needs to be public. Ok, I can change it into public and modify it with final

sczyh30: Each config key constant could be public.

linlinisme: emm, I had committed latest code. Could you help me view it?

linlinisme: File.separator rather than "/"

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
```diff
@@ -0,0 +1,119 @@
...
+            e.printStackTrace();
+        }
+        return properties;
+    }
+
+
+    private static Properties loadPropertiesFromRelativeFile(String fileName) {
+        List<URL> list = new ArrayList<URL>();
+        try {
+            Enumeration<URL> urls = getClassLoader().getResources(fileName);
```
sczyh30: I've tested the code in several scenarios. For directly run the program in IDE (i.e. `java xxx`), it could work well. However, when I tested this with a fat-jar, the relative path seemed not to work well.

For example, a jar is located in `temp` directory and the config file is located in `temp/config/my-config.properties`. In `temp` directory I execute `java -Dcsp.sentinel.config.file=config/my-config.properties -jar test.jar`, and Sentinel will not retrieve properties from the target file.

linlinisme: try` java -Dcsp.sentinel.config.file= my-config.properties -jar test.jar` ?

linlinisme:
feature support : **retrieve properties from relative file.** 
now, there is three config ways:

1. absolute path
`-Dcsp.sentinel.config.file=/user/application/config/sentinel.properties
`
2. class path
`-Dcsp.sentinel.config.file=classpath:sentinel.properties
`
3. relative path
`-Dcsp.sentinel.config.file=config/sentinel.properties
`
What need to be attention is the relative root path is the `user.dir`. For example if you start app in the `/user/application` dir and you config file is `/user/application/config/sentinel.properties`
the correct config way is `-Dcsp.sentinel.config.file=config/sentinel.properties
`
Dear @sczyh30  would you help me view the code?

linlinisme:
friendly ping :)

sczyh30:
> friendly ping :)

I'll review these days :)

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/api/ApiPathPredicateItem.java
```diff
@@ -15,9 +15,10 @@
  */
 package com.alibaba.csp.sentinel.adapter.gateway.common.api;
 
+import com.alibaba.csp.sentinel.adapter.gateway.common.SentinelGatewayConstants;
+
```
sczyh30: Could you please remove the unrelated changes in `sentinel-api-gateway-adapter-common` module? This PR will be squashed and it should only contain changes related to config :)

linlinisme: Thanks for reminding, this is a mistake, formatting the code when merging the master code. I have revert it.

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/api/ApiPathPredicateItem.java
```diff
@@ -15,10 +15,10 @@
  */
 package com.alibaba.csp.sentinel.adapter.gateway.common.api;
 
-import java.util.Objects;
-
 import com.alibaba.csp.sentinel.adapter.gateway.common.SentinelGatewayConstants;
 
+import java.util.Objects;
+
```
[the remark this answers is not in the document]
sczyh30: Still minor changes left... Could you please handle here?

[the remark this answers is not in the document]
linlinisme: Sorry I don't notice that, It is my  Intellij Idea Optimize Import Function change that. Some times this function may be it s annoying. The change I had reset.

sczyh30:
Excellent, thanks for contributing!

Just one more thing to consider: it would be convenient if we could also support setting appName (`project.name`) in the properties file. This might entail discarding the legacy logic (`~/logs/csp/${appName}.properties`) and should be discussed later.

PR #817, opened 2019-06-07
Fix unknown command issue in sentinel-transport-netty-http module

### LINKED ISSUE #810: cannot  access cluster flow rule information from http:ip:port/cluster/server/flowRules
<!-- Here is for bug reports and feature requests ONLY! 

HttpServerHandler.parseTarget这块代码可能是个bug, 由于只截取了第二个string，
导致访问 http://ip:port/cluster/server/info 提示“Unknown command "cluster/server/info"

因为 而HttpServer.handlerMap 里面存放CommandHandler 的key是  /cluster/server/info 而不是cluster

private String parseTarget(String uri) {
        if (StringUtil.isEmpty(uri)) {
            return "";
        }
        String[] arr = uri.split("/");
        if (arr.length < 2) {
            return "";
        }
        return arr[1];
    }

cdfive:

If a command whose mapping path and command name has more than one "/", 
the CommandHandler can't be found in `sentinel-transport-netty-http`,
such as:
`/cluster/server/flowRules`
`FetchClusterFlowRulesCommandHandler`
will case:
Unknown command "cluster"

Fixes #810

### Describe how you did it

Modify the `String parseTarget(String uri)` method in `HttpServerHandler` class.
Don't use `uri.split("/")` to parse, but use `indexOf("/")` instead, to remove the 
/ in uri, just like the `parseRequest` method of `HttpEventTask` class in `sentinel-transport-simple-http` module.

### Describe how to verify it

Add 2 test cases in `HttpServerHandlerTest` class: 
`testFetchClusterFlowRulesCommandEmptyRule`
`testFetchClusterFlowRulesCommandSomeFlowRules`
In order to test, `sentinel-cluster-server-default` dependency is added with `test` scope.

Run demo in `sentinel-demo-cluster-server-alone` module,
change the `sentinel-transport-simple-http` to `sentinel-transport-netty-http`,
start the `ClusterServerDemo` and visit http://localhost:8719/cluster/server/flowRules

### Special notes for reviews
The previous test case which marked FIXME in `HttpServerHandlerTest` class has been fixed, I found a way that we can call `embeddedChannel.readOutbound()` looply to get the whole result.

Some comments in test class have been improved, such as upper case the first letter.

In pom.xml of parent module, `sentinel-transport-netty-http` is added in `<dependencyManagement>` , so that we needn't add version number when use it.

### DIFF
--- a/sentinel-transport/sentinel-transport-netty-http/src/test/java/com/alibaba/csp/sentinel/transport/command/netty/HttpServerHandlerTest.java
+++ b/sentinel-transport/sentinel-transport-netty-http/src/test/java/com/alibaba/csp/sentinel/transport/command/netty/HttpServerHandlerTest.java
@@ -16,15 +16,14 @@
 package com.alibaba.csp.sentinel.transport.command.netty;
 
 import com.alibaba.csp.sentinel.Constants;
-import com.alibaba.csp.sentinel.cluster.flow.rule.ClusterFlowRuleManager;
 import com.alibaba.csp.sentinel.config.SentinelConfig;
 import com.alibaba.csp.sentinel.init.InitExecutor;
 import com.alibaba.csp.sentinel.slots.block.RuleConstant;
-import com.alibaba.csp.sentinel.slots.block.flow.ClusterFlowConfig;
 import com.alibaba.csp.sentinel.slots.block.flow.FlowRule;
 import com.alibaba.csp.sentinel.slots.block.flow.FlowRuleManager;
 import com.alibaba.csp.sentinel.transport.CommandCenter;
 import com.alibaba.csp.sentinel.transport.command.NettyHttpCommandCenter;
+import com.alibaba.csp.sentinel.transport.command.handler.MultipleSlashNameCommandTestHandler;
 import com.alibaba.fastjson.JSON;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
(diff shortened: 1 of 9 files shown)


sentinel-transport/sentinel-transport-netty-http/src/test/java/com/alibaba/csp/sentinel/transport/command/netty/HttpServerHandlerTest.java
```diff
@@ -152,60 +164,125 @@ public void testGetRuleCommandFlowSomeRules() {
...
+        String httpRequestStr = "GET /cluster/server/flowRules HTTP/1.1" + CRLF
+                + "Host: localhost:8719" + CRLF
+                + CRLF;
+        String expectedBody = "[]";
+
+        processSuccess(httpRequestStr, expectedBody);
+    }
+
+    @Test
+    public void testFetchClusterFlowRulesCommandSomeFlowRules() {
```
sczyh30: Introducing `sentinel-cluster-server-default` here causes the test case surrounded by the logic of loading cluster rules, which might be too complicated. How about we add a test handler (with multiple `/` in its name) and just test for it?

cdfive: Thanks, really a good idea!
The `sentinel-cluster-server-default` dependency has been removed.
A `MultipleSlashNameCommandTestHandler` class is added, with TestHandler suffix means that it's a test handler. Its command name has multiple /, the resources folder and META-INF/services are added in test folder, to load the Handler by SPI.
Now the test case method is `testMultipleSlashNameCommand`.

sczyh30:
Nice work, thanks!

Issue #688, opened 2019-04-19
Make maxParamByteSize configurable in ParamFlowRequestDataWriter of cluster client module


sczyh30:
## Issue Description

Type: *feature request*

### Describe what happened (or what feature you want)

We can make `maxParamByteSize` configurable in `ParamFlowRequestDataWriter` of cluster client module. In some circumstances the parameter might be very large, causing error (though large parameter is not recommended as it would affect performance).

Besides, there's also a bug that `maxParamByteSize` did not take effect. This should be fixed.

PR #823, opened 2019-06-11
fix issue688: Make maxParamByteSize configurable in ParamFlowRequestDataWriter of cluster client module

linlinisme:

fix: https://github.com/alibaba/Sentinel/issues/688

### Describe how you did it
1、maxParamByteSize take effect :  
 when  serialize parameters, use the return value of 
 calculateParamAmount()  as the numbers that  parameters  can be written.

2、make the maxParamByteSize configurable: 
 add property: **csp.sentinel.cluster.max.param.byte.size**   in SentinelConfig class
 in DefaultClusterClientInitFunc class,  when invoke initDefaultEntityWriters() method, read from Sentinel to get the property
  
### Describe how to verify it
run in cluster model，and configure the  **csp.sentinel.cluster.max.param.byte.size**.  Observe whether the configuration takes effect。and i add some log when the params size over head the configure size.


sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfig.java
```diff
@@ -52,12 +52,17 @@
     public static final String TOTAL_METRIC_FILE_COUNT = "csp.sentinel.metric.file.total.count";
     public static final String COLD_FACTOR = "csp.sentinel.flow.cold.factor";
     public static final String STATISTIC_MAX_RT = "csp.sentinel.statistic.max.rt";
+    /**
+     * cluster config
+     */
+    public static final String MAX_PARAM_BYTE_SIZE = "csp.sentinel.cluster.max.param.byte.size";
```
sczyh30: The `SentinelConfig` itself only reserves and initializes default constants for `sentinel-core`. For Sentinel cluster module we might need to add another config getter class (like [this in transport common module](https://github.com/alibaba/Sentinel/blob/master/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java), which delegates the getter methods to `SentinelConfig`).

linlinisme: ok, i try it.

linlinisme:
I add ClusterSentinelConfig class which dedicated to reading startup configurations of cluster client

sentinel-cluster/sentinel-cluster-client-default/src/main/java/com/alibaba/csp/sentinel/cluster/client/codec/data/ParamFlowRequestDataWriter.java
```diff
@@ -55,36 +54,37 @@ public void writeTo(ParamFlowRequestData entity, ByteBuf target) {
         target.writeInt(amount);
 
         // Serialize parameters with type flag.
-        for (Object param : entity.getParams()) {
-            encodeValue(param, target);
+        Object[] array = params.toArray();
+        for (int index = 0; index < amount; index++) {
```
sczyh30: There are some problems. Assume the parameter type list is `[double, float, SomePojo, int]`, the customized `SomePojo` object will be ignored so the calculated amount will be 3. Then if we use `[0..2]` as the index to encode values, the third `int` value would be lost. Maybe we need a `List<Object> resolveValidParams(Collection<Object> params)` method (could be refactored from `calculateParamTransportSize` method) , then we could get the amount and all values in order from the method.

sczyh30:
And could you please update the test cases in `ParamFlowRequestDataWriterTest`?

linlinisme:
emm, good suggestion.

linlinisme:
I lack some considerations....

linlinisme:
the latest code had commit

sczyh30:
Nice work, thanks for contributing!

PR #836, opened 2019-06-15
Add log to console support for common logs

### LINKED ISSUE #795: Is that possible to config log destination to stdout in container environment
## Issue Description
Type: feature request

In container environment, we desire output log to stdout, so that we can collection log with EFK, but I cannot find a config option

Output log to stdout

1. 
2. 
3. 

### Tell us your environment
Kubernetes

cdfive:

In container environment, user may desire output log to stdout so that the log can be collected by ELK/EFK.

This PR simply add a option(`csp.sentinel.log.type`) to support log to console for common log(`RecordLog`,`CommandCenterLog`) which use log handler mechanism.

By default,  log to file as previous.

To make it log to console(stdout), add the VM argument:
`-Dcsp.sentinel.log.type=console`

For other two kinds of log:
`metrics log` and `sentinel-block.log`, and abstract **logging SPI** interface we may need more discussions.

Fixes #795 

### Describe how you did it

Add `ConsoleHandler` class which extends from `java.util.logging.ConsoleHandler`,
and set the OutputStream to System.out

Modify `LogBase` class, add `csp.sentinel.log.type` for choose log to **file** or **console**.

### Describe how to verify it

Add `-Dcsp.sentinel.log.type=console`, then start sentinel-dashboard and view the log in console.

### Special notes for reviews

The **level** of log message is added in `CspFormatter` class, since I think it maybe useful for user who view the log message.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/ConsoleHandler.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/ConsoleHandler.java
@@ -43,15 +43,9 @@ class ConsoleHandler extends Handler {
      */
     private StreamHandler stderrHandler;
 
-    /**
-     * Current Handler according to the current publish log level.
-     */
-    private StreamHandler currentHandler;
-
     public ConsoleHandler() {
         this.stdoutHandler = new StreamHandler(System.out, new CspFormatter());
         this.stderrHandler = new StreamHandler(System.err, new CspFormatter());
-        this.currentHandler = this.stdoutHandler;
     }
 
     @Override
@@ -69,34 +63,23 @@ class ConsoleHandler extends Handler {
     @Override
     public void publish(LogRecord record) {
         if (record.getLevel().intValue() >= Level.WARNING.intValue()) {
-            currentHandler = stderrHandler;
+            stderrHandler.publish(record);
+            stderrHandler.flush();
         } else {
-            currentHandler = stdoutHandler;
+            stdoutHandler.publish(record);
+            stdoutHandler.flush();
         }
-
-        currentHandler.publish(record);
-        currentHandler.flush();
     }
 
     @Override
     public void flush() {
-        currentHandler.flush();
+        stdoutHandler.flush();
+        stderrHandler.flush();
     }
 
     @Override
     public void close() throws SecurityException {
-        currentHandler.close();
-    }
-
-    public StreamHandler getStdoutHandler() {
-        return stdoutHandler;
-    }
-
-    public StreamHandler getStderrHandler() {
-        return stderrHandler;
-    }
-
-    public StreamHandler getCurrentHandler() {
-        return currentHandler;
+        stdoutHandler.close();
+        stderrHandler.close();
     }
 }
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/ConsoleHandler.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/ConsoleHandler.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 1999-2018 Alibaba Group Holding Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.alibaba.csp.sentinel.log;
+
+/**
+ * This <tt>Handler</tt> publishes log records to <tt>System.out</tt>.
+ *
+ * Extends from {@link java.util.logging.ConsoleHandler}, and set the OutputStream to System.out.
+ *
+ * To use this handler, add the following VM argument:
+ * <pre>
+ * -Dcsp.sentinel.log.type=console
+ * </pre>
+ *
+ * @author cdfive
+ */
+class ConsoleHandler extends java.util.logging.ConsoleHandler {
+
+    public ConsoleHandler() {
+        super();
+        setOutputStream(System.out);
+    }
+}
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/ConsoleHandler.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/ConsoleHandler.java
@@ -15,12 +15,14 @@
  */
 package com.alibaba.csp.sentinel.log;
 
+import java.io.UnsupportedEncodingException;
 import java.util.logging.*;
 
 /**
- * This <tt>Handler</tt> publishes log records to <tt>System.out</tt>.
+ * This Handler publishes log records to console by using {@link java.util.logging.StreamHandler}.
  *
- * Extends from {@link java.util.logging.ConsoleHandler}, and set the OutputStream to System.out.
+ * Print log of WARNING level or above to System.err,
+ * and print log of INFO level or below to System.out.
  *
  * To use this handler, add the following VM argument:
  * <pre>
@@ -29,35 +31,72 @@ import java.util.logging.*;
  *
  * @author cdfive
  */
-class ConsoleHandler extends StreamHandler {
+class ConsoleHandler extends Handler {
+
+    /**
+     * A Handler which publishes log records to System.out.
+     */
+    private StreamHandler stdoutHandler;
+
+    /**
+     * A Handler which publishes log records to System.err.
+     */
+    private StreamHandler stderrHandler;
+
+    /**
+     * Current Handler according to the current publish log level.
+     */
+    private StreamHandler currentHandler;
 
     public ConsoleHandler() {
-        super();
-        setOutputStream(System.out);
+        this.stdoutHandler = new StreamHandler(System.out, new CspFormatter());
+        this.stderrHandler = new StreamHandler(System.err, new CspFormatter());
+        this.currentHandler = this.stdoutHandler;
     }
 
-    /**
-     * Publish a <tt>LogRecord</tt>.
-     * <p>
-     * The logging request was made initially to a <tt>Logger</tt> object,
-     * which initialized the <tt>LogRecord</tt> and forwarded it here.
-     * <p>
-     * @param  record  description of the log event. A null record is
-     *                 silently ignored and is not published
-     */
+    @Override
+    public synchronized void setFormatter(Formatter newFormatter) throws SecurityException {
+        this.stdoutHandler.setFormatter(newFormatter);
+        this.stderrHandler.setFormatter(newFormatter);
+    }
+
+    @Override
+    public synchronized void setEncoding(String encoding) throws SecurityException, UnsupportedEncodingException {
+        this.stdoutHandler.setEncoding(encoding);
+        this.stderrHandler.setEncoding(encoding);
+    }
+
     @Override
     public void publish(LogRecord record) {
-        super.publish(record);
-        flush();
+        if (record.getLevel().intValue() >= Level.WARNING.intValue()) {
+            currentHandler = stderrHandler;
+        } else {
+            currentHandler = stdoutHandler;
+        }
+
+        currentHandler.publish(record);
+        currentHandler.flush();
     }
 
-    /**
-     * Override <tt>StreamHandler.close</tt> to do a flush but not
-     * to close the output stream.  That is, we do <b>not</b>
-     * close <tt>System.err</tt>.
-     */
     @Override
-    public void close() {
-        flush();
+    public void flush() {
+        currentHandler.flush();
+    }
+
+    @Override
+    public void close() throws SecurityException {
+        currentHandler.close();
+    }
+
+    public StreamHandler getStdoutHandler() {
+        return stdoutHandler;
+    }
+
+    public StreamHandler getStderrHandler() {
+        return stderrHandler;
+    }
+
+    public StreamHandler getCurrentHandler() {
+        return currentHandler;
     }
 }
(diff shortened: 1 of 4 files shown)


jasonjoo2010:
And i noticed that the last commit with a commend `Fix System.err no effect` but i haven't seed any related change. Did you forget to include it?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/LogBase.java
```diff
@@ -55,27 +60,38 @@
...
-        String logDir = System.getProperty(LOG_DIR);
+        logType = System.getProperty(LOG_TYPE);
 
-        if (logDir == null || logDir.isEmpty()) {
-            logDir = System.getProperty(USER_HOME);
-            logDir = addSeparator(logDir) + DIR_NAME + File.separator;
+        // By default, log to file
+        if (StringUtil.isBlank(logType)) {
+            logType = LOG_TYPE_FILE;
+        } else if (LOG_TYPE_FILE.equals(logType) && LOG_TYPE_CONSOLE.equals(logType)) {
```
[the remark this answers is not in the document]
cdfive: Sorry, this judgement is mistake, and `equalsIgnoreCase` is better. thanks!

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/LogBase.java
```diff
@@ -87,6 +103,10 @@ public static boolean isLogNameUsePid() {
         return logNameUsePid;
     }
 
+    public static boolean isLogToConsole() {
+        return LOG_TYPE_CONSOLE.endsWith(logType);
```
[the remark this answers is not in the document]
cdfive: Should be equals, my typo. This method is my previous test, no other place used, I'll remove it.

cdfive:
> And i noticed that the last commit with a commend `Fix System.err no effect` but i haven't seed any related change. Did you forget to include it?

```java
class ConsoleHandler extends java.util.logging.ConsoleHandler {

    public ConsoleHandler() {
        super();
        setOutputStream(System.out);
    }
}
```
Previously, call `super()` then `setOutputStream(System.out);`
this has a problem, after that, I found call `System.err.println()` has no effect, maybe its due to the `super()` method close the writer which correspond to System.err.

This is test code:
```java
System.err.println(111);
Writer writer = new OutputStreamWriter(System.err);
writer.close();
System.err.println(111);
System.out.println(222);
```
The second 111 can't be print to console.

So I changed it extends from `java.util.logging.StreamHandler` and test it's okay.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/ConsoleHandler.java
```diff
@@ -29,35 +31,72 @@
...
-        super.publish(record);
-        flush();
+        if (record.getLevel().intValue() >= Level.WARNING.intValue()) {
+            currentHandler = stderrHandler;
+        } else {
+            currentHandler = stdoutHandler;
+        }
+
+        currentHandler.publish(record);
+        currentHandler.flush();
```
jasonjoo2010: And performance maybe down if we flush during every publishing.

sentinel-core/src/test/java/com/alibaba/csp/sentinel/log/ConsoleHandlerTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+        when(logRecord.getLevel()).thenReturn(Level.INFO);
+        when(logRecord.getMessage()).thenReturn("test info message");
+        consoleHandler.publish(logRecord);
+        assertEquals(consoleHandler.getStdoutHandler(), consoleHandler.getCurrentHandler());
+
+        when(logRecord.getLevel()).thenReturn(Level.WARNING);
+        when(logRecord.getMessage()).thenReturn("test warning message");
+        consoleHandler.publish(logRecord);
+        assertEquals(consoleHandler.getStderrHandler(), consoleHandler.getCurrentHandler());
```
jasonjoo2010: If you do unit testing you can simply mock the `System.out` & `System.err` for correct response.

And I think `LogRecord` has now need to be mocked because we can easily get an instance we need.

cdfive: > LogRecord has now need to be mocked because we can easily get an instance we need.
Good idea.

How to mock **System.out**? I found a way by using `ByteArrayOutputStream` like this:
```java
ByteArrayOutputStream baosOut = new ByteArrayOutputStream();
System.setOut(new PrintStream(baosOut));
```
And then verify `baosOut.toString()`, is it ok?

PS: Thanks for so detailed reviewing!

jasonjoo2010:
> ```java
> Writer writer = new OutputStreamWriter(System.err);
> writer.close();
> System.err.println(111);
> System.out.println(222);
> ```

Here for system sharing stream it's not polite closing it.
So from the implementation of ConsoleHandler we can see they only flush it when close()

cdfive:
@jasonjoo2010 
The `write.close()` are called in `flushAndClose` method, which is called by `setOutputStream` and `close` in `StreamHandler` class. Previously, 
```java
super();
setOutputStream(System.out);
```
`setOutputStream` is called twice, `setOutputStream(System.err)` are called in `super()`.

When publish(logRecord), it also call flush().
`java.util.logging.ConsoleHandler`:
```java
@Override
public void publish(LogRecord record) {
    super.publish(record);
    flush();
}
```

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/ConsoleHandler.java
```diff
@@ -29,35 +31,72 @@
...
-     * @param  record  description of the log event. A null record is
-     *                 silently ignored and is not published
+     * A Handler which publishes log records to System.err.
      */
+    private StreamHandler stderrHandler;
+
+    /**
+     * Current Handler according to the current publish log level.
+     */
+    private StreamHandler currentHandler;
```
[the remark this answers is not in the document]
cdfive: Yes, no need to currentHandler. It has been removed. When `close` is invoked, both two handler call `close`.

jasonjoo2010:
> @jasonjoo2010
> The `write.close()` are called in `flushAndClose` method, which is called by `setOutputStream` and `close` in `StreamHandler` class. Previously,
> 
> ```java
> super();
> setOutputStream(System.out);
> ```
> 
> `setOutputStream` is called twice, `setOutputStream(System.err)` are called in `super()`.
> 
> When publish(logRecord), it also call flush().
> `java.util.logging.ConsoleHandler`:
> 
> ```java
> @Override
> public void publish(LogRecord record) {
>     super.publish(record);
>     flush();
> }
> ```

Yes, indeed.
But it's for all kinds of streams and considers it generally but in our scenario we should close the System.out and System.err.

Certainly if you want you can get rid of the limitation of StreamHandler and use the System.out and System.err directly.

sentinel-core/src/test/java/com/alibaba/csp/sentinel/log/ConsoleHandlerTest.java
jasonjoo2010: Kinds of, for example:

```java
    @Test
    public void test() throws IOException {
        PrintStream stdout = mock(PrintStream.class);
        PrintStream stderr = mock(PrintStream.class);
        System.setOut(stdout);
        System.setErr(stderr);
        System.err.write("asdf\n".getBytes());
        System.out.write("asdf\n".getBytes());
        System.out.write("asdf\n".getBytes());
        verify(stdout, times(2)).write(any(byte[].class));
        verify(stderr).write(any(byte[].class));
    }
```

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/ConsoleHandler.java
cdfive: If not call `flush()` here, the console maynot get the log info immediately, as we call log.info()
to log oneline message, we expected to view them immediately.
Start dashboard can verify it.
Besides, the `java.util.logging.ConsoleHandler` class also call flush() when publish(),
so I think its okay.

sentinel-core/src/test/java/com/alibaba/csp/sentinel/log/ConsoleHandlerTest.java
cdfive: Thanks, I learned it. It can verify action.

cdfive: As encapsulated by `OutputStreamWriter`, to verify the byte[] write is not very convenient.
I tried with `ByteArrayOutputStream`, it works and no mock in this test case.Is it OK?

jasonjoo2010: > As encapsulated by `OutputStreamWriter`, to verify the byte[] write is not very convenient.
> I tried with `ByteArrayOutputStream`, it works and no mock in this test case.Is it OK?

Yes there is no need to verify the content i think we only try to verify the times invoked and it's enough.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/CspFormatter.java
```diff
@@ -37,6 +37,7 @@ public String format(LogRecord record) {
         final DateFormat df = dateFormatThreadLocal.get();
         StringBuilder builder = new StringBuilder(1000);
         builder.append(df.format(new Date(record.getMillis()))).append(" ");
+        builder.append(record.getLevel().getName()).append(" ");
```
jasonjoo2010: And sentinel-core's logging didn't have level included before? I haven't noticed that.

jasonjoo2010:
It seems i have no further question.

What's your opinion, @sczyh30 ?

sczyh30:
> It seems i have no further question.
> 
> What's your opinion, @sczyh30 ?

Sorry for the late reply. I'll take a review tomorrow.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/CspFormatter.java
sczyh30: Good, we'd ignored it before. And maybe we can improve the date format in `CspFormatter` to be `yyyy-MM-dd HH:mm:ss.SSS`. It might be helpful to have milliseconds recorded. @cdfive

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/LogBase.java
```diff
@@ -35,15 +36,19 @@
 
     public static final String LOG_CHARSET = "utf-8";
 
+    public static final String LOG_TYPE_FILE = "file";
+    public static final String LOG_TYPE_CONSOLE = "console";
+
     private static final String DIR_NAME = "logs" + File.separator + "csp";
     private static final String USER_HOME = "user.home";
 
+    public static final String LOG_TYPE = "csp.sentinel.log.type";
```
sczyh30: Maybe names like `csp.sentinel.record.log.output.type` or another brief one would be better? This could indicate this property is only for biz log (`RecordLog` and `CommandCenterLog`) and designed for the output type.

oneuone:
------------------ 原始邮件 ------------------
发件人: "Eric Zhao"<notifications@github.com>
发送时间: 2019年6月17日(星期一) 晚上9:03
收件人: "alibaba/Sentinel"<Sentinel@noreply.github.com>;
抄送: "Subscribed"<subscribed@noreply.github.com>;
主题: Re: [alibaba/Sentinel] Add log to console support for common logs(#836)



@sczyh30 commented on this pull request.
 
 
In sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/LogBase.java:
 >      private static final String DIR_NAME = "logs" + File.separator + "csp";      private static final String USER_HOME = "user.home";   +    public static final String LOG_TYPE = "csp.sentinel.log.type";  
Maybe names like csp.sentinel.record.log.output.type or another brief one would be better? This could indicate this property is only for biz log (RecordLog and CommandCenterLog) and designed for the output type.
 
&mdash;

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/LogBase.java
cdfive: Indicate the detail meaning is better, how about a little brief name `csp.sentinel.log.biz.output`, which first means log, then biz means(RecordLog and CommandCenterLog), then output means the output type, output to file or console.

In the future, we can support metric log config like `csp.sentinel.log.metric.output`.

As a choosing disorder, how about replace biz to system?
`csp.sentinel.log.system.output`
`csp.sentinel.log.biz.output`
`csp.sentinel.log.sys.output`
`csp.sentinel.log.system.output.type`
`csp.sentinel.log.biz.output.type`
`csp.sentinel.record.log.output.type`
`csp.sentinel.log.system.appender`

I prefer 1 or 2.
What's your opinion? @sczyh30 @jasonjoo2010

jasonjoo2010: Actually i prefer not to limit to record logging.

We can adjust a common interface in future for all sentinel modules i think. What's you plan @sczyh30 ?

jasonjoo2010: Actually i prefer not to limit to record logging.

We can adjust a common interface in future for all sentinel modules i think. What's you plan @sczyh30 ?

cdfive: > Actually i prefer not to limit to record logging.
> 
> We can adjust a common interface in future for all sentinel modules i think. What's you plan @sczyh30 ?

How about we separate this config parameter at present，since for metric log, as it may grows fast, it's Inappropriate output to console.

jasonjoo2010: > > Actually i prefer not to limit to record logging.
> > We can adjust a common interface in future for all sentinel modules i think. What's you plan @sczyh30 ?
> 
> How about we separate this config parameter at present，since for metric log, as it may grows fast, it's Inappropriate output to console.

I think `metric log` is a kind of structural storage but not a kind of logging. So just exclude it.

sczyh30: This could be a temporary solution. I think after the logging interface is ready, it would be better to choose different output types by introducing the specific SPI implementation module, rather than putting it in the command line args.

cdfive: Agree. This is to provide a easy temporary way for user who want to ouput biz log to console : )

sczyh30:
Thanks!

PR #840, opened 2019-06-17
Fix the README to the new metrics writing format

jasonjoo2010:
Someones have been asked why they got an addition column compared to what the README said.

So I update the README up to date.


sczyh30:
Thanks!

PR #850, opened 2019-06-20
Add batch request support for transport of cluster flow control

### LINKED ISSUE #849: Add batch request support for transport of cluster flow control
## Issue Description

Type: *feature request*

In the current implementation, if there are multiple cluster rules (=N) of a resource, when checking for cluster rules, the token client will send up to N requests to the token server respectively to acquire tokens. Thus the performance might be bad when there're many cluster rules (e.g. gateway rules for different types of HTTP request items) and the network is not well. We might need to add batch request support for `TokenService` and cluster transport implementations to handle this.

sczyh30:

Add batch request support for transport of cluster flow control to reduce network overhead in some circumstances.

Resolves #849

### Describe how you did it

- Add two batch APIs (`batchRequestToken` and `batchRequestParamToken`) in TokenService to support batch requests in cluster flow control:
- Change attachment type in `TokenResult` to `Map<String, Object>`
- Add batch token request and response entity in cluster common module
- Add codec, client and server implementation for batch cluster token request/response
  - Add batch token request/response decoder and writers respectively
  - Update `DefaultClusterTokenClient` to support batch requests
  - Add request processors for batch token request in token server module
  - Polish the `DefaultTokenService` in cluster server module
- Other code refinement

### Describe how to verify it

Run the test cases.

### Special notes for reviews

Please take care of the new marshalling and unmarshalling logic.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -142,7 +142,23 @@
             </dependency>
             <dependency>
                 <groupId>com.alibaba.csp</groupId>
+                <artifactId>sentinel-cluster-client-default</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.alibaba.csp</groupId>
+                <artifactId>sentinel-cluster-server-default</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.alibaba.csp</groupId>
                 <artifactId>sentinel-adapter</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.alibaba.csp</groupId>
+                <artifactId>sentinel-api-gateway-adapter-common</artifactId>
                 <version>${project.version}</version>
             </dependency>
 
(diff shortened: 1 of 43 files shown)


linlinisme:
the method `com.alibaba.csp.sentinel.cluster.TokenService#batchRequestParamToken 
`
`   TokenResult batchRequestParamToken(Set<Long> ruleIds, int acquireCount, Map<Integer, Object> paramMap);` 
why it is not 
`   TokenResult batchRequestParamToken(Set<Long> ruleIds, int acquireCount, Map<Integer, Collection<Object>> paramMap);`
The code looks more uniform

`com.alibaba.csp.sentinel.cluster.request.data.ParamFlowRequestData#params
`
if it should be ordered collection, why it still use Collection type？The caller may be does not  know that there is such a requirement

sczyh30:
> the method `com.alibaba.csp.sentinel.cluster.TokenService#batchRequestParamToken `
> ` TokenResult batchRequestParamToken(Set<Long> ruleIds, int acquireCount, Map<Integer, Object> paramMap);`
> why it is not
> ` TokenResult batchRequestParamToken(Set<Long> ruleIds, int acquireCount, Map<Integer, Collection<Object>> paramMap);`
> The code looks more uniform

The case (single `paramIdx` position with multiple parameters) is rare and we supported this in `requestParamToken` only for legacy logic. But indeed compatibility should be considered. I'll check this later.

> `com.alibaba.csp.sentinel.cluster.request.data.ParamFlowRequestData#params `
> if it should be ordered collection, why it still use Collection type？The caller may be does not know that there is such a requirement

Ehh, I've put the comment in the wrong place... The `params` in `ParamFlowRequestData` belongs to a single `paramIdx` so they don't need to be ordered.

Issue #861, opened 2019-06-24
The passCount in DegradeRule from AtomicLong to LongAdder


wchswchs:
Issue Description
Type: enhancement

Describe what happened (or what feature you want)
Because counting passCount when degrade is a high concurrency scenario, change `passCount` in `DegradeRule` to improve the performance.

wchswchs:
I have contributed.

sczyh30:
Closing the issue as the implementation of circuit breaker will be refactored.

PR #862, opened 2019-06-24
Enhancement/passCount in DegradeRule to LongAdder

wchswchs:
To improve the performance for counting passCount in degrade scenario.

Features #861 

Describe how you did it
Change `passCount` in `DegradeRule` from `AtomicLong` to `LongAdder`

Describe how to verify it
Run ut `DegradeTest` below `compare.slots.block.degrade.passcount` package.
`DegradeTest` below `compare.slots.block.degrade.passcount.atomic` for `AtomicLong` Test.
`DegradeTest` below `compare.slots.block.degrade.passcount.longadder` for `LongAdder` Test.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
@@ -22,12 +22,12 @@ import com.alibaba.csp.sentinel.node.DefaultNode;
 import com.alibaba.csp.sentinel.slots.block.AbstractRule;
 import com.alibaba.csp.sentinel.slots.block.RuleConstant;
 import com.alibaba.csp.sentinel.slots.clusterbuilder.ClusterBuilderSlot;
+import com.alibaba.csp.sentinel.slots.statistic.base.LongAdder;
 
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * <p>
@@ -178,7 +178,7 @@ public class DegradeRule extends AbstractRule {
 
     // Internal implementation (will be deprecated and moved outside).
 
-    private AtomicLong passCount = new AtomicLong(0);
+    private LongAdder passCount = new LongAdder();
     private final AtomicBoolean cut = new AtomicBoolean(false);
 
     @Override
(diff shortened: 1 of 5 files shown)


wchswchs:
I'll add unit tests in these days.

wchswchs:
I have added unit test to compare counting `passCount` between `AtomicLong` and `LongAdder`.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
```diff
@@ -195,12 +195,13 @@ public boolean passCheck(Context context, DefaultNode node, int acquireCount, Ob
...
             if (rt < this.count) {
-                passCount.set(0);
+                passCount.reset();
                 return true;
             }
 
             // Sentinel will degrade the service only if count exceeds.
-            if (passCount.incrementAndGet() < rtSlowRequestAmount) {
+            passCount.increment();
+            if (passCount.sum() < rtSlowRequestAmount) {
```
sczyh30:
```java
passCount.increment();
if (passCount.sum() < rtSlowRequestAmount)
```

The operations are not atomic? It might lead to inaccurate calculation.

sczyh30:
And we need to consider whether it's really necessary to make this change.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
wchswchs: The operations are atomic because `cells` and `base` in `LongAdder` are volatile.

wchswchs:
> And we need to consider whether it's really necessary to make this change.

I think degrade maybe bring so large traffic, so I think this change is necessary.

linlinisme:
```
passCount.increment();
if (passCount.sum() < rtSlowRequestAmount)
```
这里有个问题就是如果当A,B请求同时进来，当A请求passCount + 1, 些时B请求passCount + 1。这时passCount.sum()就是两个相加之后。本来A是可以直接return true的，结果B的+1影响到了A，就return false了，这种逻辑上就有问题了

wchswchs:
> ```
> passCount.increment();
> if (passCount.sum() < rtSlowRequestAmount)
> ```
> 
> The problem here is that if A and B requests come in at the same time, when A requests passCount + 1, then B requests passCount + 1. At this time passCount.sum() is the two additions. Originally, A can directly return true. As a result, +1 of B affects A,AI return false. This is logically problematic.

I see, but in `LongAdder`, I think it has solved the problem by the below code：
image

linlinisme:
> > ```
> > passCount.increment();
> > if (passCount.sum() < rtSlowRequestAmount)
> > ```
> > 
> > 
> > The problem here is that if A and B requests come in at the same time, when A requests passCount + 1, then B requests passCount + 1. At this time passCount.sum() is the two additions. Originally, A can directly return true. As a result, +1 of B affects A,AI return false. This is logically problematic.
> 
> I see, but in `LongAdder`, I think it has solved the problem by the code below：
> image

it is not the key point .....

wchswchs:
Let me see.

wchswchs:
I think it maybe cause thread safety between two threads: one to sum and the other to increment.

wchswchs:
I have just thinked over a lot, it has no problem for this change.

sczyh30:
Closing the deprecated PR. Thanks for contributing!

PR #869, opened 2019-06-27
Add support for managing gateway flow rules and customized API group in Sentinel dashboard

### LINKED ISSUE #699: Support managing gateway flow rules in Sentinel dashboard
## Issue Description

Type: *feature request*

Since 1.6.0, Sentinel has provided the `sentinel-api-gateway-adapter-common` module, which abstracts entity and management for gateway flow rules and customized API group definitions. It's necessary to support managing gateway flow rules and customized API group definitions via dashboard.

Note: since 1.6.0 an `appType` property field is added to identify the type of application (0 by default, gateway is 1). The heartbeat client will carry the `appType` field in the heartbeat message. The dashboard can identify the field and the "gateway rules" page is only available to the gateway services.

Discussions and contributions are welcomed~

cdfive:

Support managing gateway flow rules and customized API group definitions via dashboard.

Fixes #699 

### Describe how you did it

1. Use `appType` in heartbeat message of client to distinguish normal service or gateway service.
   Add `appType` in `AppInfo` and `MachineInfo` class, add two menus for gateway type app in 
  `sidebar.html`.

2. Add `GatewayApiController`, `GatewayFlowRuleController`  and corresponding entity,repository,vo 
in gateway package(controller/gateway,datasource/entity/gateway,repository/gateway,vo/gateway).
   Add fetch and modify method in `SentinelApiClient` to interactive with gateway command handler.
3. Add `api.html`,`flow.html` and js files in corresponding gateway folder.

### Describe how to verify it

1. Start dashboard with VM argument:
`-Dproject.name=sentinel-dashboard -Dcsp.sentinel.dashboard.server=localhost:8080`

2. Start sentinel-demo-spring-cloud-gateway with VM argument:
`-Dproject.name=sc-gateway-demo -Dcsp.sentinel.dashboard.server=localhost:8080 -Dcsp.sentinel.app.type=1`

3. Visit [http://localhost:8090/httpbin/json](http://localhost:8090/httpbin/json) to make sentinel init in 
sc-gateway-demo.

4. View and operate list/add/update/delete API group and gateway flow rules via dashboard pages,
check sc-gateway-demo's api groups or flow rules by visiting http://localhost:8720/gateway/getApiDefinitions,http://localhost:8720/gateway/getRules
Check via UI and view logs to verify.

A simple test class `ApiDefinitionControllerTest` is added, I'll try to improve the unit test cases later.

### Special notes for reviews

1. To make the gateway flow rules in order, the type of rule map in `GatewayRuleManager` class is changed from `ConcurrentHashMap` to `Collections.synchronizedMap(new LinkedHashMap()`
`HashSet` to `LinkedhashSet`.
In this way, when user add or update one rule, then refresh the list page, the order of rules can remain
(in adding order).

2. To support set interval by different unit of second,minute,hour,day via dashboard, a property named `intervalUnit` is added in `GatewayFlowRule` class, and `intervalSec` change to `interval`, the `calIntervalSec` method is added to get the interval second by `interval` and `intervalUnit`.

3. Previously, a variable name in `login.js` isn't a good naming, though it doesn't affect functionality, I change the `LoginService` to `AuthService`, to make the meaning accurate.

4. Besides, I post a blog about how to build the front-end developing environment of sentinel dashboard, which may help users who are in first participation, and add it in awesome-sentinel.md.

PTAL.


sczyh30:
👍

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/rule/GatewayFlowRule.java
```diff
@@ -31,7 +31,8 @@
 
     private int grade = RuleConstant.FLOW_GRADE_QPS;
     private double count;
-    private long intervalSec = 1;
+    private long interval = 1;
```
sczyh30: This could bring breaking changes. In fact, we could just convert the time unit in the dashboard side (convert to seconds), and keep `intervalSec` here.

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/rule/GatewayFlowRule.java
```diff
@@ -52,6 +53,23 @@ public GatewayFlowRule(String resource) {
         this.resource = resource;
     }
 
+    public long calIntervalSec() {
```
sczyh30: Ditto. This logic can be moved to the dashboard.

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/rule/GatewayRuleManager.java
```diff
@@ -46,9 +39,9 @@
     /**
      * Gateway flow rule map: (resource, [rules...])
      */
-    private static final Map<String, Set<GatewayFlowRule>> GATEWAY_RULE_MAP = new ConcurrentHashMap<>();
+    private static final Map<String, Set<GatewayFlowRule>> GATEWAY_RULE_MAP = Collections.synchronizedMap(new LinkedHashMap<String, Set<GatewayFlowRule>>());
```
sczyh30: `Collections.synchronizedMap` could bring bad performance, which is not recommended. 

The sorting could be a problem, but I think it's okay in the dashboard as it's only a sample here. In production we could store the rules in DB and sort by its id.

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/rule/GatewayFlowRule.java
cdfive: I also think about it, but if keep intervalSec here, though we can convert in dashboard side, we can't display the unit which are choosen by user.
e.g.
If user input 2, choose hour unit, convert to 7200 second in gateway rule. 
Then user refresh the list, and edit the rule, we can only get 7200,
choose 7200 second or 2 hour will get the same 7200 in this way.

I found that the `getIntervalSec` only used in serveral place, and change them to `calIntervalSec()`.

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/rule/GatewayRuleManager.java
cdfive: Indeed the performance of `Collections.synchronizedMap` is weaker,  how about we consider that the add or modify rules are not frequent operations, maynot need high concurrency performance.
Though in memery store, user may more care about friendly visibility, the order in list page may be important.

Yes, if we store the rules in DB and sort by its id, there is no such problem.

I'm also wondering why JUC didn't offer `ConcurrentLinkedHashMap`.
Google's `ConcurrentLinkedHashMap` and JDK's `ConcurrentSkipListMap` seems also not suitable for such a circumstance.
Is there any better way?

jasonjoo2010: Or why use it here?
Why doesn't ConcurrentHashMap meet the purpose?

For changeable iterating you can load it into a CopyOnWriteSet before iterating and iterate the CopyOnWriteSet safely.
For modification both to modify in place or to initialize new container and replace the reference shared are safe.

sczyh30: `Collections.synchronizedMap` will also lock the read operations (e.g. `get`), so it might not be a good solution here.

In fact, I don't think the sorting problem should be addressed in SDK (as for the SDK it's not needed till now). It might be more appropriate to resolve this in the dashboard (though in the dashboard without DB it's hard to achieve).

cdfive: @jasonjoo2010  Test add or update one gateway flow rule, when I click save, I found that the order of rules in list page are changed often, this may make user confused. It seems due to the none order of ConcurrentHashMap here. Hope it keeps in order when add or modify.

jasonjoo2010: > @jasonjoo2010 Test add or update one gateway flow rule, when I click save, I found that the order of rules in list page are changed often, this may make user confused. It seems due to the none order of ConcurrentHashMap here. Hope it keeps in order when add or modify.

Oh I see.
I think for me maybe I will fetch the keys and sort them before rendering. And if `Collections.synchronizedMap` could solve it it's better to drop a comment here explaining why use synchronized implementation but not for synchronized reason.

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/rule/GatewayFlowRule.java
cdfive: It has been done. Keep intervalSec in `GatewayFlowRule` and convert in the dashboard side.
In `GatewayFlowRuleEntity` class, two methods `calIntervalSec`,`parseIntervalSec` are added to do convert.

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/rule/GatewayFlowRule.java
cdfive: It has been removed.

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/rule/GatewayRuleManager.java
cdfive: ConcurrentHashMap and HashSet in RuleManager class have been reverted.
I also found that other rule like flow,degrade,system, in their list page, the same problem exists.
It doesn't has much effect on function, only effect the display.

There may be other approaches that we can think about to achieve right sorting.

sentinel-dashboard/src/main/webapp/resources/app/scripts/controllers/gateway/flow.js
```diff
@@ -0,0 +1,232 @@
...
+          getMachineRules();
+          gatewayFlowRuleDialog.close();
+        } else {
+          alert('新增流控规则失败!' + data.msg);
+        }
+      });
+    };
+
+    function saveRule(rule, edit) {
+      GatewayFlowService.saveRule(rule).success(function (data) {
```
sczyh30: Maybe we need to check whether the rule is valid here?

sentinel-dashboard/src/main/webapp/resources/app/views/dialog/gateway/flow-rule-dialog.html
```diff
@@ -0,0 +1,159 @@
...
+								<input type='number' min="0" class="form-control highlight-border" ng-model='currentRule.count' placeholder="请输入" />
+							</div>
+						</div>
+					</div>
+
+					<div class="form-group">
+						<div>
+							<label class="col-sm-2 control-label">间隔</label>
+							<div class="col-sm-3">
+								<input type='number' min="1" class="form-control highlight-border" ng-model='currentRule.interval' placeholder="单位(秒)" />
```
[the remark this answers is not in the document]
cdfive: Meticulous! Yes, it should be updated when user change different interval unit.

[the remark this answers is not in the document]
cdfive: Oh, as the select after the input displayed the interval unit, how about the placeholder here just "请输入"?

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/datasource/entity/gateway/GatewayFlowRuleEntity.java
```diff
@@ -0,0 +1,341 @@
...
+    private String app;
+    private String ip;
+    private Integer port;
+
+    private Date gmtCreate;
+    private Date gmtModified;
+
+    private String resource;
+    private Integer resourceMode;
+
```
sczyh30: Missing `grade` here?

sentinel-dashboard/src/main/webapp/resources/app/scripts/controllers/gateway/flow.js
```diff
@@ -0,0 +1,232 @@
...
+        width: 780,
+        overlay: true,
+        scope: $scope
+      });
+    };
+
+    $scope.addNewRule = function () {
+      var mac = $scope.macInputModel.split(':');
+      $scope.currentRule = {
+        grade: 0,
```
sczyh30: The default grade should be `1` (QPS)

sczyh30: And this field does not take effect due to the missing field of `GatewayRuleEntity`.

sentinel-dashboard/src/main/webapp/resources/app/views/dialog/gateway/flow-rule-dialog.html
```diff
@@ -0,0 +1,159 @@
...
+							<input type="text" name="fieldName" class="form-control highlight-border" placeholder="请输入" ng-model='currentRule.paramItem.fieldName' />
+						</div>
+					</div>
+
+					<div class="form-group" ng-if="currentRule.paramItem != null && (currentRule.paramItem.parseStrategy == 2 || currentRule.paramItem.parseStrategy == 3 || currentRule.paramItem.parseStrategy == 4)">
+						<label class="col-sm-2 control-label">匹配模式</label>
+						<label class="col-sm-5 control-label">
+							<div class="form-control highlight-border" align="center">
+								<input type="radio" value="0" checked ng-model="currentRule.paramItem.matchStrategy" title="精确" />&nbsp;精确&nbsp;&nbsp;
+								<input type="radio" value="1" ng-model="currentRule.paramItem.matchStrategy" title="前缀" />&nbsp;前缀&nbsp&nbsp;
```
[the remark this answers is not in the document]
cdfive: matchStrategy中的前缀已去掉，目前3种matchStrategy  精确：0 子串：1 正则：2
对应的界面和`SentinelGatewayConstants`类都已修改。

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/datasource/entity/gateway/GatewayFlowRuleEntity.java
cdfive: Yes， `grade` propertity has been added.

sentinel-dashboard/src/main/webapp/resources/app/scripts/controllers/gateway/flow.js
cdfive: Default grade has been fixed to `1`(QPS), corresponding grade element in page and entity added.

sentinel-dashboard/src/main/webapp/resources/app/scripts/controllers/gateway/flow.js
cdfive: In 115 line, before calling `addNewRule` and `saveRule` , call `GatewayFlowService.checkRuleValid($scope.currentRule)` to check, and the check method has been improved.

sentinel-dashboard/src/main/webapp/resources/app/views/dialog/gateway/flow-rule-dialog.html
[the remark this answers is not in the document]
cdfive: Fixed, with tip text `统计窗口时间长度` in this placeholder.

sczyh30:
Fabulous! Just one more advice: we need to keep the customized API name unique. We could add validation in the "new API definition" dialog to check whether the current API list contains the API group name.

sentinel-dashboard/src/main/webapp/resources/app/scripts/controllers/gateway/flow.js
sczyh30: 👍

cdfive:
> Fabulous! Just one more advice: we need to keep the customized API name unique. We could add validation in the "new API definition" dialog to check whether the current API list contains the API group name.

Good idea, the customized API name should be unique.

cdfive:
The unique validation for customized API name  has been added, both front and back ends. Since the api name can't be modify when editing, only validate when adding new customized API .

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/SentinelGatewayConstants.java
```diff
@@ -37,9 +37,8 @@
     public static final int URL_MATCH_STRATEGY_REGEX = 2;
 
     public static final int PARAM_MATCH_STRATEGY_EXACT = 0;
-    public static final int PARAM_MATCH_STRATEGY_PREFIX = 1;
```
sczyh30: This might bring compatibility problem so we might need to keep its old value.

sentinel-extension/sentinel-parameter-flow-control/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/param/ParamFlowRuleUtil.java
```diff
@@ -168,7 +163,7 @@ public static void fillExceptionFlowItems(ParamFlowRule rule) {
 
             if (flowRules == null) {
                 // Use hash set here to remove duplicate rules.
-                flowRules = new HashSet<>();
+                flowRules = new LinkedHashSet<>();
```
sczyh30: Is this change necessary?

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/SentinelGatewayConstants.java
cdfive: It has been reverted to keep compatibility.

sentinel-extension/sentinel-parameter-flow-control/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/param/ParamFlowRuleUtil.java
cdfive: To keep rules in order previously, has been reverted.

sentinel-dashboard/src/main/webapp/resources/gulpfile.js
```diff
@@ -111,7 +111,8 @@ gulp.task('clean', function () {
 });
 
 // 总任务
-gulp.task('build', ['clean', 'jshint', 'lib', 'js', 'css']);
+// gulp.task('build', ['clean', 'jshint', 'lib', 'js', 'css']);
+gulp.task('build', ['clean', 'lib', 'js', 'css']);
```
sczyh30: Why is the `jshint` task removed?

cdfive: Oh I'm sorry for commit local modify file,  I'll fix it soon.

sczyh30:
Nice work! Could you please squash and separate your commits according to the corresponding module? For the commit related to the dashboard, it can be prefixed with `dashboard: ` :)

sczyh30:
Thanks for your awesome contribution!

Issue #316, opened 2018-12-19
Slot chain SPI should target slot itself rather than the entire builder


jialianglinjl:
## Issue Description

Type: *feature request*

### Describe what happened (or what feature you want)

Due to the fact that the order between slots in slot chain is essential, currently Sentinel make `SlotChainBuilder` as the slot chain SPI so that users can implement their own builder, where they can arrange the order between slots. But this introduce complexity for users to arrange the builder when there are many external slots from other dependency.

### Describe what you expected to happen

The ideal approach is to make the slot itself as the SPI interface so that users only need to care about their own slot, rather than the global chain. But due to the fact that the order between slots in slot chain is essential, we have to provide a `@Order` annotation for users so that they can arrange the order with the annotation. And the default slots of Sentinel (e.g. `StatisticSlot`) should have their own unique order respectively.

tigerMoon:
Maybe I can have a try. its more like filter chain.

as I see.   
user can define custom `Slot`, and  order in the slot chain,  then register to the `SlotChain`, and `SlotChain.getNext()` indexed with order. 

`NodeSelectorSlot`, `StatisticSlot`, `ClusterBuilderSlot ` are the basic type, and others are custom type.  

we provide SlotRegister that only custom type slot allowed.

PR #871, opened 2019-06-27
 Modify variable names and Optimize methods

yidadi:
Modify variable names and Optimize methods

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slotchain/SlotChainProvider.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slotchain/SlotChainProvider.java
@@ -15,12 +15,9 @@
  */
 package com.alibaba.csp.sentinel.slotchain;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.ServiceLoader;
-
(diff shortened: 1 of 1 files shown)


sczyh30:
Hi, could you please illustrate why these changes are necessary?

yidadi:
This method #resolveSlotChainBuilder just get the first SlotChainBuilder from the list.

sczyh30:
Maybe your change is not compatible with the current design:

> - If there is no user-defined `SlotChainBuilder`, then use the default slot chain builder
> - If user-defined `SlotChainBuilder` exists, the provider will pick the first loaded customized slot chain builder, then cache the builder.


See #145 for details.

yidadi:
Why? Is it necessary to put an object in a list and get only the first one?

sczyh30:
The list was initially designed for different strategies (e.g. pick the first customized, pick the final customized or ordered by `@SpiOrder`) but is not actually used indeed... I think it's okay to remove the list, but support for `@SpiOrder` should also be considered (we could leverage the `SpiLoader` class). For further improvement of the slot SPI, we could refer to #316

sczyh30:
Thanks!

PR #880, opened 2019-07-01
remove the ex.printStackTrace() to avoid OOM and Redundant code

yidadi:
remove the ex.printStackTrace() to avoid OOM and Redundant code


sczyh30:
Some of the exceptions (e.g. failed to initialize) are **fatal errors** and should be printed for warning. And I don't think it's necessary to change it in the demo, as it's not intuitive.

yidadi:
Now that the log output by RecordLog , there is no need to print it out again.

PR #889, opened 2019-07-04
fix  AuthorityRuleEntity and ParamFlowRuleEntity fastjson  serialize problem

linlinisme:

the ParamFlowRuleEntity and AuthorityRuleEntity json serialize is wrong when they use `com.alibaba.fastjson.JSON#toJSONString(java.lang.Object)`. Because the `@JsonIgnore` annotation  is not belong to fastjson, but com.fasterxml.jackson.  They are different  json lib.  So
 `com.alibaba.fastjson.JSON#toJSONString(java.lang.Object)` still json serialize the field though it modified with `@JsonIgnore`。

fix https://github.com/alibaba/Sentinel/issues/948
### Describe how you did it

 add  fastjson's   annotation   `@JSONField(serialize = false)`  at the base of `@JsonIgnore`

### Describe how to verify it
run test cases :`com.alibaba.csp.sentinel.dashboard.datasource.entity.JsonSerializeTest`


sczyh30:
Thanks!

For V2 controllers (rules for the entire application) we may need another good solution. Maybe we could just push the origin `ParamFlowRule` instead of the rule entity (the `ip` and `port` is deprecated in V2 scenarios). Maybe the `RuleRepository` should also be refactored to adapt this.

Issue #881, opened 2019-07-02
remove the tmpLog  dir


linlinisme:
## Issue Description

 *bug report* 

### Describe what happened (or what feature you want)

every time I install Sentinel my /user/home will add several junk files like
```
tmpLogDir1562036016836
tmpLogDir1562036596518
tmpLogDir1562037419434
```
I found it is created by some test cases. 
It can be remove when the cases run over?
It is annoying to see this junk files, because I have to remove it  manually.

sczyh30:
Hi, could you please investigate which test cases generated the temporary directories?

linlinisme:
`com.alibaba.csp.sentinel.RecordLogTest`
```
@Test
    public void testChangeLogBase() {
        String userHome = System.getProperty("user.home");
        String newLogBase = userHome + File.separator + "tmpLogDir" + System.currentTimeMillis();
        System.setProperty(LogBase.LOG_DIR, newLogBase);

        RecordLog.info("testChangeLogBase");
        String logFileName = RecordLog.getLogBaseDir();
        File[] files = new File(logFileName).listFiles();
        assertTrue(files != null && files.length > 0);
    }

    @Test
    public void testLogBaseDir() {
        RecordLog.info("testLogBaseDir");
        assertTrue(RecordLog.getLogBaseDir().startsWith(System.getProperty("user.home")));
    }
```
`com.alibaba.csp.sentinel.slots.logger.EagleEyeLogUtilTest
`
 ```
@Test
    public void testWriteLog() throws Exception {
        EagleEyeLogUtil.log("resourceName", "BlockException", "app1", "origin", 1);

        final File file = new File(RecordLog.getLogBaseDir() + EagleEyeLogUtil.FILE_NAME);
        await().timeout(2, TimeUnit.SECONDS)
            .until(new Callable<File>() {
                @Override
                public File call() throws Exception {
                    return file;
                }
            }, FileMatchers.anExistingFile());
    }

    @Test
    public void testChangeLogBase() throws Exception {
        String userHome = System.getProperty("user.home");
        String newLogBase = userHome + File.separator + "tmpLogDir" + System.currentTimeMillis();
        System.setProperty(LogBase.LOG_DIR, newLogBase);

        EagleEyeLogUtil.log("resourceName", "BlockException", "app1", "origin", 1);

        final File file = new File(RecordLog.getLogBaseDir() + EagleEyeLogUtil.FILE_NAME);
[...]

PR #908, opened 2019-07-12
fix temp file problem

linlinisme:
###Describe what this PR does / why we need it
When install `sentinel` project once, temporary files create  in the `user.home` dir. Time and time the `user.home` become confused. And I have to remove the tmp file manually. Some times I almost delete the whole  `user.dir`. I think the tmp file should not be created or delete.

fix https://github.com/alibaba/Sentinel/issues/881
### Describe how you did it
`com.alibaba.csp.sentinel.slots.logger.EagleEyeLogUtilTest#testChangeLogBase
`
`com.alibaba.csp.sentinel.RecordLogTest#testChangeLogBase
`
 I remove the @Test annotation
 Because change `LogBase` It is not not work when integration Testing
` LogBase.LOG_DIR` can be just static init for once and it will not be changed

And I fix  original verify logic problem.


sczyh30:
Thanks!

PR #914, opened 2019-07-15
Add the feature of excluding unused URLs in CommonFilter (#287)

### LINKED ISSUE #287: CommonFilter provides interface to exclude some unused URLs
## Issue Description

Type: *feature request*

应用中的某些url不需要进行统计，比如spring boot 端点请求信息，能否增加相关的扩展点，可以过滤掉不想统计的url

Request feature: support filtering urls that don't want to be resource in `CommonFilter`.

lym-ifae:
### Why we need it
We need to support filtering urls that don't want to be resource in `CommonFilter`.

Resolves #287 

### Describe how you did it
As there's already a UrlCleaner interface, we could just reuse the interface and improve the logic of unifying the resource name with UrlCleaner. If developers are going to exclude some URLs, they can convert the URLs to the empty string "" in their `UrlCleaner` implementation, and in `CommonFilter` we could check whether the cleaned resource name is empty. If it's empty, then it should be excluded and won't go through the logic of Sentinel `SphU.entry`.

### Describe how to verify it
We can get the statistic node of the resource via `ClusterBuilderSlot.getClusterNode(res)` method, and if the `ClusterNode` of the target URL is absent, then we think it has been excluded.

### DIFF
--- a/sentinel-adapter/sentinel-web-servlet/src/main/java/com/alibaba/csp/sentinel/adapter/servlet/CommonFilter.java
+++ b/sentinel-adapter/sentinel-web-servlet/src/main/java/com/alibaba/csp/sentinel/adapter/servlet/CommonFilter.java
@@ -73,19 +73,19 @@ public class CommonFilter implements Filter {
                 target = urlCleaner.clean(target);
             }
 
-            // Parse the request origin using registered origin parser.
-            String origin = parseOrigin(sRequest);
-
-            ContextUtil.enter(target, origin);
-            entry = SphU.entry(target, EntryType.IN);
-
-
-            // Add method specification if necessary
-            if (httpMethodSpecify) {
-                methodEntry = SphU.entry(sRequest.getMethod().toUpperCase() + COLON + target,
-                        EntryType.IN);
+            // If you intend to exclude some URLs, you can convert the URLs to the empty string ""
+            // in the UrlCleaner implementation.
+            if (target != "") {
+                // Parse the request origin using registered origin parser.
+                String origin = parseOrigin(sRequest);
+                ContextUtil.enter(target, origin);
+                entry = SphU.entry(target, EntryType.IN);
+                // Add method specification if necessary
+                if (httpMethodSpecify) {
+                    methodEntry = SphU.entry(sRequest.getMethod().toUpperCase() + COLON + target,
+                            EntryType.IN);
+                }
             }
-
             chain.doFilter(request, response);
         } catch (BlockException e) {
             HttpServletResponse sResponse = (HttpServletResponse) response;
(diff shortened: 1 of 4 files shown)


sentinel-adapter/sentinel-web-servlet/src/main/java/com/alibaba/csp/sentinel/adapter/servlet/CommonFilter.java
```diff
@@ -73,19 +73,19 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
-            entry = SphU.entry(target, EntryType.IN);
-
-
-            // Add method specification if necessary
-            if (httpMethodSpecify) {
-                methodEntry = SphU.entry(sRequest.getMethod().toUpperCase() + COLON + target,
-                        EntryType.IN);
+            // If you intend to exclude some URLs, you can convert the URLs to the empty string ""
+            // in the UrlCleaner implementation.
+            if (target != "") {
```
sczyh30: Here we could use `StringUtil.isEmpty(target)` instead.

sczyh30:
And could you please also update the document here: https://github.com/alibaba/Sentinel/blob/master/sentinel-adapter/sentinel-web-servlet/README.md

sentinel-adapter/sentinel-web-servlet/src/main/java/com/alibaba/csp/sentinel/adapter/servlet/CommonFilter.java
```diff
@@ -107,7 +107,9 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             if (entry != null) {
                 entry.exit();
             }
-            ContextUtil.exit();
+            if (ContextUtil.getContext() != null) {
```
cdfive: This judgment seems unnecessary, since it has been done inside `ContextUtil.exit()` method.

sczyh30:
@lym-ifae The CI status indicates there's something wrong with your changes: https://travis-ci.org/alibaba/Sentinel/builds/559316648

sentinel-adapter/sentinel-web-servlet/src/main/java/com/alibaba/csp/sentinel/adapter/servlet/CommonFilter.java
```diff
@@ -73,19 +73,19 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
-            entry = SphU.entry(target, EntryType.IN);
-
-
-            // Add method specification if necessary
-            if (httpMethodSpecify) {
-                methodEntry = SphU.entry(sRequest.getMethod().toUpperCase() + COLON + target,
-                        EntryType.IN);
+            // If you intend to exclude some URLs, you can convert the URLs to the empty string ""
+            // in the UrlCleaner implementation.
+            if (StringUtil.isEmpty(target)) {
```
sczyh30: This should be `!StringUtil.isEmpty(target)`.

sczyh30:
Thanks for contributing! 👍

PR #926, opened 2019-07-17
Migrage the console server's specifying logic into transport-common

jasonjoo2010:
Because there are similar implementation in both `simple-http` and `netty-http` submodules on processing console server's address so migrate it into transport-common.

NO

### Special notes for reviews
Should actually run for verifying.


sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
```diff
@@ -48,12 +52,69 @@ public static Long getHeartbeatIntervalMs() {
...
+            if (index == 0) {
+                // skip
+                continue;
+            }
+            String host = ipPortStr;
+            if (index >= 0) {
+                try {
+                    port = Integer.parseInt(ipPortStr.substring(index + 1));
+                    if (port <= 1 || port >= 65535) {
+                        throw new RuntimeException("Port number [" + port + "] over range");
```
sczyh30: Maybe here we could just ignore the invalid item, rather than fail directly. What do you think of it?

jasonjoo2010: The dealing in old logics,
it will be just ignored in `simple-http` while in `netty-http` an exception is generated (in `InetSocketAddress.checkPort()`) but also caught immediately in place, log a warning in record.log, print a stack trace to standard output.

Because the logging of sentinel can not be merged into the biz project's logging so the warning may be ignored. So the malformed `:port` is more like to be eaten and initializing flow will go on without the correct dashboard target.
That's why some guys suffer from no heartbeat / no application registered.

I like the idea of stopping to fix rather than ignore but logging. 
But I will also feel ok on keeping the logic the same as how it acted in `netty-http`.

sczyh30: It's okay for me. I'll review the PR these days.

sczyh30:
Could you please resolve the conflicts?

jasonjoo2010:
> Could you please resolve the conflicts?

They have been resolved.

sczyh30:
Thanks!

PR #922, opened 2019-07-16
Fix the problem that dashboard do not correctly show the SystemRule in CPU usage strategy

Crazy10552:
### Does this pull request fix one issue?
   Fixes #905 

### Describe how you did it
  
-  I add new properties in entity [com.alibaba.csp.sentinel.dashboard.datasource.entity.rule.SystemRuleEntity]

- I update the related js and html about  System rule 

### Describe how to verify it
   
image

### Special notes for reviews
 None

PR #927, opened 2019-07-18
Fix the problem that dashboard do not correctly show the SystemRule in CPU usage strategy

### LINKED ISSUE #905: SystemRule usage
## Issue Description
rule do not work all and the olny rule which works but it do not set as i want 
Type: *bug report* 

> I set five system rule,use zookeeperdatasource,now receive system rule update data
image

>then i use  curl http://127.0.0.1:8720/getRules?type=system to get my rule

>debug find  [SystemRuleManager ](https://github.com/alibaba/Sentinel/blob/master/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemRuleManager.java)  code like that
image
>only first rule can seem to be seen, someone like the first rule,but the only rule become five rule
image

>also dashboard can not show system which i set highestCpuUsage
image

image

### Tell us your environment
dubbo 2.5.3
sentinel version 1.6.2

Crazy10552:
   Fixes #905 

### Describe how you did it
  
-  I add new properties in entity [SystemRuleEntity](https://github.com/alibaba/Sentinel/blob/master/sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/datasource/entity/rule/SystemRuleEntity.java)

- I update the related js and html about  System rule 

### Describe how to verify it
   
image

### Special notes for reviews
 i recreate my branch to slove my commit problem of my git account,the previous pull [pull922](https://github.com/alibaba/Sentinel/pull/922)


sczyh30:
Thanks for contributing! I'll refine the code later.

Issue #631, opened 2019-03-29
Will be support timeout fallback in the future?


px-kelvin:
##Issue Description

Type: feature request

##Describe what happened (or what feature you want)

can‘t change hystrix to sentinel because RT options not suitable our project

### Describe what you expected to happen

@SentinelResource can support timeout fallback is  like timeoutInMilliseconds property of hystrix @HystrixCommand  in the future .  or  can you give me some alternatives

mvphjx:
me too； need a timeout strategy

sczyh30:
See #945 for more discussions.

PR #945, opened 2019-07-29
Add timeout "cut-down" mechanism in @SentinelResource annotation

huang054:

Sometimes the system can't respond in time for some reason, like fast failure, service downgrade, user's request will not block all the time.
Service downgrade with @SentinelResource without timeout, plus timeout service downgrade

### Describe how you did it
Use the get() time parameter of the future to determine whether it times out

### Describe how to verify it
  @SentinelResource(value = "hello", fallback = "helloFallback",timeOut = 1000)

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
@@ -70,6 +70,8 @@ public @interface SentinelResource {
      */
     String defaultFallback() default "";
 
+    int timeOut() default 0;
+
     /**
      * The {@code fallback} is located in the same class with the original method by default.
      * However, if some methods share the same signature and intend to set the same fallback,
--- a/sentinel-extension/sentinel-annotation-aspectj/src/main/java/com/alibaba/csp/sentinel/annotation/aspectj/SentinelResourceAspect.java
+++ b/sentinel-extension/sentinel-annotation-aspectj/src/main/java/com/alibaba/csp/sentinel/annotation/aspectj/SentinelResourceAspect.java
@@ -26,6 +26,10 @@ import org.aspectj.lang.annotation.Aspect;
 import org.aspectj.lang.annotation.Pointcut;
 
 import java.lang.reflect.Method;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 /**
  * Aspect for methods with {@link SentinelResource} annotation.
@@ -40,7 +44,7 @@ public class SentinelResourceAspect extends AbstractSentinelAspectSupport {
     }
 
     @Around("sentinelResourceAnnotationPointcut()")
-    public Object invokeResourceWithSentinel(ProceedingJoinPoint pjp) throws Throwable {
+    public Object invokeResourceWithSentinel(final ProceedingJoinPoint pjp) throws Throwable {
         Method originMethod = resolveMethod(pjp);
 
         SentinelResource annotation = originMethod.getAnnotation(SentinelResource.class);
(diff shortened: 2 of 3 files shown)


huang054:
_
postman
service

_

sczyh30:
Hi, could you please describe what this PR does and why we need it according to the PR template?

huang054:
> 嗨，您能否根据PR模板描述这个PR的作用以及我们为什么需要它？

Sometimes the system can't respond in time for some reason, like fast failure, service downgrade, user's request will not block all the time.

sczyh30:
Thanks for contributing. This kind of "cut-down" mechanism for managed thread pools should be well discussed and designed. Just a managed `SingleThreadExecutor` is not enough.

huang054:
> Thanks for contributing. This kind of "cut-down" mechanism for managed thread pools should be well discussed and designed. Just a managed `SingleThreadExecutor` is not enough.

I am referring to netty to use the cpu core number to set the thread pool size, you mean that the implementation here should reduce the thread hosting use.

0704681032:
the biz logic will be executed in another thread,is it ok?

huang054:
> the biz logic will be executed in another thread,is it ok?


I  test it success  because the future.get() is blocked.

0704681032:
when the biz logic code depend on some thread related attribute such as threadlocal,it will be got somthing wrong



---Original---

sczyh30:
Any thoughts on the design? @CarpenterLee @cdfive @jasonjoo2010 @linlinisme

jasonjoo2010:
> Any thoughts on the design? @CarpenterLee @cdfive @jasonjoo2010 @linlinisme

I am agree with @0704681032 , including:

1. Executing in **async way** is implicit if timeout is set to positive value compared to zero. It may lead to bug.

2. Timeout doesn't really mean the execution's termination. Especially we could use rate limiting/threads rules to protect the logic block. So the timeout doesn't make sense to service provider while it often does on calling side(eg. timeout in client / request / consumer in client-server / request-response / consumer-provider model).

3. If we really want the logic block have such kind of timeout we'd better check every invocation's setting carefully (HttpClient / Service Subscription / SQL execution / etc.) or wrap it using Runnable block in any executor simply (It's not a common solution).


These above are my consideration and if anything goes wrong please let me know

linlinisme:
> > Any thoughts on the design? @CarpenterLee @cdfive @jasonjoo2010 @linlinisme
> 
> I am agree with @0704681032 , including:
> 
> 1. Executing in **async way** is implicit if timeout is set to positive value compared to zero. It may lead to bug.
> 2. Timeout doesn't really mean the execution's termination. Especially we could use rate limiting/threads rules to protect the logic block. So the timeout doesn't make sense to service provider while it often does on calling side(eg. timeout in client / request / consumer in client-server / request-response / consumer-provider model).
> 3. If we really want the logic block have such kind of timeout we'd better check every invocation's setting carefully (HttpClient / Service Subscription / SQL execution / etc.) or wrap it using Runnable block in any executor simply (It's not a common solution).
> 
> These above are my consideration and if anything goes wrong please let me know

My suggestion is:
1. Blocked the main thread and wait for another thread pool execute may not be a appropriate. First, blocked main thread is not recommended, why not consider something like reactor way? Second, submit request to another thread pool will reduced  application performance, especially when the request's response is slow, will increase  system burden. And the Thread Pool thread num is hard to decide,  for the compute task and i/o task 's thread num is different.
2. I have a proposal:   Every request  we create a `time-listener`,  it responsible for time-out notify.  And submit to a `ScheduledThreadPoolExecutor` to execute.  When the  'time-listener' task execute, if the request is not completed, execute time out logic. Otherwise, when request completed we `clean` the` time-listener`. It still just a preliminary idea, every one can give you opinion.

linlinisme:
any progress?

sczyh30:
Closing this due to no further feedback. We could discuss the design in #631 if needed.

Issue #66, opened 2018-08-17
[Feature] Support setting the universal default rule for all resources


evenin:
据了解，目前的规则只能针对某个资源进行设置。希望提供支持全局规则设置，或资源名可通配符表达。
应用场景：所有资源默认错误率达到多少自动降级，有特殊需求的资源另外设置。

CarpenterLee:
很好的建议，对减少用户规则设置的工作量有很大帮助。

我们不支持规则通配符主要有两个考虑：
- 一个是统计效率，模糊匹配会增加计算量；
- 一个是规则的明确性，每一条规则的作用范围应该是明确的，避免误杀其他资源。

**全局默认降级规则**的想法不错，可以再深入讨论一下，比如把全局规则存储在哪里、跟用户的规则是什么关系等等。

evenin:
目前规则与统计插槽（ClusterBuilderSlot）是强关系……针对这样设计，我认为可以以这种方式实现（改动最小）：复制全局规则以初始化某资源规则。经过我初步测试是可行的。以下是我修改DegradeRuleManager.checkDegrade()的代码，供参与

/** 修改支持全局规则 - shenjian **/
	private static final String GLOBAL = "*";

	public static void checkDegrade(ResourceWrapper resource, Context context, DefaultNode node, int count) throws BlockException {
		if (degradeRules == null) {
			return;
		}
		List<DegradeRule> rules = getRules(resource.getName());
		if (rules == null) {
			return;
		}
		for (DegradeRule rule : rules) {
			if (!rule.passCheck(context, node, count)) {
				throw new DegradeException(rule.getLimitApp());
			}
		}
	}

	private static List<DegradeRule> getRules(String resourceName) {
		List<DegradeRule> rules = degradeRules.get(resourceName);
		if (rules != null) {
			return rules;
		}
		rules = degradeRules.get(GLOBAL);
		if (rules == null || rules.isEmpty()) {
			return null;
		}
		// 使用全局规则初始化资源规则
		rules = copyRules(rules, resourceName);
		List<DegradeRule> rules2 = degradeRules.putIfAbsent(resourceName, rules);
		if (rules2 != null && !rules2.equals(rules)) {// 没有添加成功：规则已存在
			return rules2;
		}
		return rules;
	}

	private static List<DegradeRule> copyRules(List<DegradeRule> rules, String newResourceName) {
		List<DegradeRule> list = new ArrayList<DegradeRule>(rules.size());
[...]

PR #956, opened 2019-08-01
Support setting global rule

linlinisme:

Fix: https://github.com/alibaba/Sentinel/issues/66

### Describe how you did it

1. Global rules support both flow and degrade.
2. The flow and degrade global rules have one and only one rule.
3. Global rules are created with specific resources
4. GlobalManger manages them: create, store, update, and retrieve.
5. When the resource performs flow and degrade check, it will get the rule check through GlobalManager in real time.
6. Global rules and resource rules themselves are superimposed on each other. The resource itself is checked first, and the final check is the global rule.

### Describe how to verify it
      Run test cases I offer
### Special notes for reviews
     None


sczyh30:
👍 I'll take a look these days :)

Issue #952, opened 2019-07-31
ConfigUtil should support reading files with customized charset


sczyh30:
## Issue Description

Type: *enhancement*

### Describe what happened (or what feature you want)

The [ConfigUtil](https://github.com/alibaba/Sentinel/blob/master/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java) should support reading files with customized charset, or else the content will be malformed when the properties file contains non-ASCII characters (e.g. Chinese characters).

Maybe we could leverage the `InputStreamReader` and `BufferedReader` class. For example:

```java
try (BufferedReader bufferedReader = Files.newBufferedReader(Paths.get(fileName), Charset.forName(SentinelConfig.charset()))) {
    properties = new Properties();
    properties.load(bufferedReader);
}
```

We could get the charset from `SentinelConfig.charset()` (by default it's `UTF-8`).

### Tell us your environment

Sentinel 1.7.0-SNAPSHOT on `master` branch

PR #961, opened 2019-08-03
ConfigUtil support reading files with customized charset

aq0706:
Fixes [#952](https://github.com/alibaba/Sentinel/issues/952)

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
@@ -17,12 +17,11 @@ package com.alibaba.csp.sentinel.util;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.InputStreamReader;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
@@ -71,7 +70,7 @@ public final class ConfigUtil {
             }
 
             try (BufferedReader bufferedReader =
-                         Files.newBufferedReader(Paths.get(fileName), getCharset())) {
+                         new BufferedReader(new InputStreamReader(new FileInputStream(file), getCharset()))) {
                 properties = new Properties();
                 properties.load(bufferedReader);
             }
@@ -113,7 +112,7 @@ public final class ConfigUtil {
         Properties properties = new Properties();
         for (URL url : list) {
             try (BufferedReader bufferedReader =
-                         new BufferedReader(new InputStreamReader(url.openStream(), getCharset().newDecoder()))) {
+                         new BufferedReader(new InputStreamReader(url.openStream(), getCharset()))) {
                 Properties p = new Properties();
                 p.load(bufferedReader);
                 properties.putAll(p);
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
@@ -19,6 +19,7 @@ import com.alibaba.csp.sentinel.config.SentinelConfig;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.InputStreamReader;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
@@ -114,7 +115,7 @@ public final class ConfigUtil {
         for (URL url : list) {
             try {
                 try (BufferedReader bufferedReader =
-                             Files.newBufferedReader(Paths.get(url.toURI()), Charset.forName(SentinelConfig.charset()))) {
+                             new BufferedReader(new InputStreamReader(url.openStream(), Charset.forName(SentinelConfig.charset())))) {
                     Properties p = new Properties();
                     p.load(bufferedReader);
                     properties.putAll(p);
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
@@ -22,6 +22,7 @@ import java.io.File;
 import java.io.InputStreamReader;
 import java.net.URL;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.ArrayList;
@@ -72,7 +73,7 @@ public final class ConfigUtil {
             }
 
             try (BufferedReader bufferedReader =
-                         Files.newBufferedReader(Paths.get(fileName), Charset.forName(SentinelConfig.charset()))) {
+                         Files.newBufferedReader(Paths.get(fileName), getCharset())) {
                 properties = new Properties();
                 properties.load(bufferedReader);
             }
@@ -114,7 +115,7 @@ public final class ConfigUtil {
         Properties properties = new Properties();
         for (URL url : list) {
             try (BufferedReader bufferedReader =
-                         new BufferedReader(new InputStreamReader(url.openStream(), Charset.forName(SentinelConfig.charset())))) {
+                         new BufferedReader(new InputStreamReader(url.openStream(), getCharset().newDecoder()))) {
                 Properties p = new Properties();
                 p.load(bufferedReader);
                 properties.putAll(p);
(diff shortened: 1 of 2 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
```diff
@@ -108,18 +113,11 @@ private static Properties loadPropertiesFromClasspathFile(String fileName) {
...
-                        p.load(input);
-                        properties.putAll(p);
-                    } finally {
-                        try {
-                            input.close();
-                        } catch (Throwable t) {
-                        }
-                    }
+                try (BufferedReader bufferedReader =
+                             Files.newBufferedReader(Paths.get(url.toURI()), Charset.forName(SentinelConfig.charset()))) {
```
sczyh30: It seems that `Files.newBufferedReader(xxx)` only support `file` URI type. When the properties file is located in the jar (i.e. `jar:file://xxx`), exceptions will be thrown:

```
java.nio.file.FileSystemNotFoundException
        at com.sun.nio.zipfs.ZipFileSystemProvider.getFileSystem(ZipFileSystemProvider.java:171)
        at com.sun.nio.zipfs.ZipFileSystemProvider.getPath(ZipFileSystemProvider.java:157)
        at java.nio.file.Paths.get(Paths.java:143)
```

Maybe we could just wrap the `url.openStream()` with `InputStreamReader` like this:

```java
for (URL url : list) {
    Properties p = new Properties();
    try (InputStreamReader reader = new InputStreamReader(url.openStream(),
        Charset.forName(SentinelConfig.charset()))) {
        p.load(reader);
        properties.putAll(p);
    } catch (Throwable e) {
        e.printStackTrace();
    }
}

```

aq0706: Thanks for your review, I made a stupid mistake and now I am correcting it.

sczyh30: 👍

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
```diff
@@ -108,18 +114,11 @@ private static Properties loadPropertiesFromClasspathFile(String fileName) {
         Properties properties = new Properties();
         for (URL url : list) {
             try {
```
sczyh30: Can we remove the redundant `try`? Just one try-catch is enough.

aq0706: 👌

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
```diff
@@ -107,20 +113,11 @@ private static Properties loadPropertiesFromClasspathFile(String fileName) {
 
         Properties properties = new Properties();
         for (URL url : list) {
-            try {
+            try (BufferedReader bufferedReader =
+                         new BufferedReader(new InputStreamReader(url.openStream(), Charset.forName(SentinelConfig.charset())))) {
```
sczyh30: It seems that there's potential circular static dependency here: `SentinelConfig` -> `SentinelConfigLoader` -> `ConfigUtil` -> `SentinelConfig`, which could lead to initialization failure. We have to ensure that the `SentinelConfig` is not referenced by `ConfigUtil`.

Here we may need to provide a universal charset in `ConfigUtil`:

```java
private static final String CHARSET = System.getProperty("csp.sentinel.charset", "UTF-8");
```

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
```diff
@@ -139,6 +140,12 @@ private static ClassLoader getClassLoader() {
         return classLoader;
     }
 
+    private static Charset getCharset() {
+        // avoid static loop dependencies: SentinelConfig -> SentinelConfigLoader -> ConfigUtil -> SentinelConfig
+        // so not use SentinelConfig.charset()
+        return Charset.forName(System.getProperty(SentinelConfig.CHARSET, StandardCharsets.UTF_8.name()));
```
sczyh30: We should not reference `SentinelConfig` here, or the circular dependency problem will still exist. Maybe we could just copy the constant, or move the constant to elsewhere.

sczyh30:
The CI indicates test failure. Could you please resolve that?
https://travis-ci.org/alibaba/Sentinel/builds/567855168

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
aq0706: `ConfigUtil` only references the constants of `SentinelConfig`, does not trigger the initialization of SentinelConfig, i think no circular dependency here.

aq0706: I understand what you mean now, I think copy constants is the best way.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/ConfigUtil.java
```diff
@@ -107,20 +112,11 @@ private static Properties loadPropertiesFromClasspathFile(String fileName) {
 
         Properties properties = new Properties();
         for (URL url : list) {
-            try {
+            try (BufferedReader bufferedReader =
+                         new BufferedReader(new InputStreamReader(url.openStream(), getCharset().newDecoder()))) {
```
sczyh30: What's the difference between `InputStreamReader(url.openStream(), getCharset().newDecoder())` and `InputStreamReader(url.openStream(), getCharset())`? It seems that the latter one is equivalent to:

```java
new InputStreamReader(url.openStream(),
  getCharset().newDecoder()
     .onMalformedInput(CodingErrorAction.REPLACE)
     .onUnmappableCharacter(CodingErrorAction.REPLACE))
```

aq0706: When `malformedInputAction` or `unmappableCharacterAction` occurs: 
- `InputStreamReader(url.openStream(), getCharset())` will replaced by a special character. 
- `InputStreamReader(url.openStream(), getCharset().newDecoder())` will throws  `MalformedInputException` or `UnmappableCharacterException`

sczyh30: Both have pros and cons. The `REPLACE` action is commonly used, as it won't affect normal attributes.

aq0706: Yes, this way you can continue reading data even if something goes wrong.

sczyh30: So here we may change to `InputStreamReader(url.openStream(), getCharset())`?

aq0706: How do we let the user know that the read property is wrong, without throwing an exception. 

The unknown error attribute is a hidden bomb.

linlinisme: `ConfigUtil.java` may be just a reading tool, Its job is to reading the contents of the file。The correctness of the contents of the file should be guaranteed by the file provider itself, not the reading tool. Later, consider presenting the current configuration on the console and supporting dynamic editing.

sczyh30: > How do we let the user know that the read property is wrong, without throwing an exception.

The malformed attribute might be discovered in the log.

aq0706: Thanks guys, I will change it later.

sentinel-core/src/test/java/com/alibaba/csp/sentinel/util/ConfigUtilTest.java
```diff
@@ -79,6 +81,45 @@ public void testLoadProperties() throws IOException {
...
+    //add Jvm parameter
+    //-Dcsp.sentinel.charset="GBK"
+    //@Test
+    public void testLoadPropertiesWithCustomizedCharset() throws IOException {
 
+        String charset = "GBK";
+
+        File file = null;
+        String dir = "/data/日志/",
+                fileName = "propertiesTest.properties";
```
linlinisme: I think the `dir` had better not included  Chinese character. Because some computer os may not support Chinese.

aq0706: 👍

sczyh30:
Thanks for contributing!

PR #963, opened 2019-08-05
Aspectj remove duplicate check

0704681032:
when codes execute here ,we can make sure that the exception is not belongsTo exceptionsToIgnore and is belongsTo exceptionsToTrace!
so,we do not need check the exception again
instead,we should call the traceException method directly

### DIFF
--- a/sentinel-extension/sentinel-annotation-aspectj/src/main/java/com/alibaba/csp/sentinel/annotation/aspectj/SentinelResourceAspect.java
+++ b/sentinel-extension/sentinel-annotation-aspectj/src/main/java/com/alibaba/csp/sentinel/annotation/aspectj/SentinelResourceAspect.java
@@ -64,7 +64,7 @@ public class SentinelResourceAspect extends AbstractSentinelAspectSupport {
                 throw ex;
             }
             if (exceptionBelongsTo(ex, annotation.exceptionsToTrace())) {
-                traceException(ex, annotation);
+                traceException(ex);
                 return handleFallback(pjp, annotation, ex);
             }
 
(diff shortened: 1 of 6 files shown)


sentinel-extension/sentinel-annotation-aspectj/src/main/java/com/alibaba/csp/sentinel/annotation/aspectj/SentinelResourceAspect.java
```diff
@@ -64,7 +64,7 @@ public Object invokeResourceWithSentinel(ProceedingJoinPoint pjp) throws Throwab
                 throw ex;
             }
             if (exceptionBelongsTo(ex, annotation.exceptionsToTrace())) {
-                traceException(ex, annotation);
+                traceException(ex);//exceptionsToIgnore not include ex &&  exceptionsToTrace include ex
```
sczyh30: Maybe the comment is not needed here?

0704681032: sorry to interrupt you,it is my first pr :( 
i create a new branch,and set git config --global user.email "307517019@qq.com",but it still include the first pr(at that time,i did not set the git config: email),so i cannot pass the cla check

sczyh30: You may need to squash your commits with your correct email (you could leverage `git rebase` with squash mode).

sczyh30: And you could just drop the last commit during your `rebase`.

xjxia:
need do a new pr?

PR #964, opened 2019-08-05
Remove duplicate exception check in SentinelResourceAspect

0704681032:
when codes execute here ,we can make sure that the exception is not belongsTo exceptionsToIgnore and is belongsTo exceptionsToTrace!
so,we do not need check the exception again
instead,we should call the traceException method directly


sczyh30:
Thanks for contributing!

PR #965, opened 2019-08-05
Spring web httpclient adapter

0704681032:
since it is very popular to use resttemplate,i think
spring web httpclient adaper should be provided


sczyh30:
Hi, thanks for contributing. Spring RestTemplate has been supported in [Spring Cloud Alibaba Sentinel](https://github.com/alibaba/spring-cloud-alibaba/wiki/Sentinel#resttemplate-支持). You may refer to https://github.com/alibaba/spring-cloud-alibaba/blob/master/spring-cloud-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/custom/SentinelProtectInterceptor.java

0704681032:
thank you,i got it.but,will you consider separate it out from springcloud alibaba,since some body do not use it



---Original---

sczyh30:
Integrations with Spring components (RestTemplate, FeignClient) are commonly put in Spring Cloud Alibaba (except the reactive components). We've illustrated this in the document :)

PR #966, opened 2019-08-06
support setting default rules for resources

linlinisme:

Fixes https://github.com/alibaba/Sentinel/issues/66
### Describe how you did it

I adopt another design:
the default rules can be created and added in the resource pre-source init while application first time access resource.
I put this op in the old lookProcessChain method, now it rename to lookProcessChainAndInitDefaultRules.
The reason I do this as follows:

**1. resource corresponding default rules can be created one time and only one time.
2. it's thread safe, if we do it at other place we may need extract syn op.
3. setting resource's default rules can be treated as a part of init method.
4. it's created on demand.**

### Describe how to verify it

run test cases.

### Special notes for reviews


pigercc:
```
if (SentinelConfig.isGlobalRuleOpen()) {
                for (Map.Entry<String, Set<DegradeRule>> entry : defaultDegradeRules.entrySet()) {
                    if (degradeRules.containsKey(entry.getKey())) {
                        defaultDegradeRules.remove(entry.getKey());
                    } else {
                        degradeRules.put(entry.getKey(), entry.getValue());
                    }
                }
            }
```
thanks for the code,it is helpful when i also need the global rule,but i think the code has a  bug

the right expectation is:
if I set degrade rule, use mine
else use the global rule

but actual performance:
1. lookProcessChainAndInitDefaultRules init 
``` 
 setRulesForResource(resource, degradeConf.get(resource));
defaultDegradeRules.put(resource, degradeConf.get(resource));
```
2 I set my degrade rule, is work
3 I remove my degrade rule,but now cannot apply the global rule
   because defaultDegradeRules has remove resource
   so degradeRules cannot add the global rule
   ```
  degradeRules.put(entry.getKey(), entry.getValue());
 ```

PR #967, opened 2019-08-06
resolved update flowRules.when between clear and putAll,getRules for

zhuyou1234:
empty.


sczyh30:
Hi, could you please describe what this PR does and why we need it according to the PR template?

zhuyou1234:
> Hi, could you please describe what this PR does and why we need it according to the PR template?

i think is a optimize.in FlowRuleManager,when it update flowRules.between clear and putAll。getRules will be empty。i think it's not reasonable

zhuyou1234:
> Hi, could you please describe what this PR does and why we need it according to the PR template?

Issue #969, opened 2019-08-06
FlowRuleManager  update flowRules.when between clear and putAll,getRules is empty.


zhuyou1234:
orgin:
private static final Map<String, List<FlowRule>> flowRules = new ConcurrentHashMap<String, List<FlowRule>>();

public void configUpdate(List<FlowRule> value) {
            Map<String, List<FlowRule>> rules = FlowRuleUtil.buildFlowRuleMap(value);
            if (rules != null) {
                flowRules.clear();
                flowRules.putAll(rules);
            }
            RecordLog.info("[FlowRuleManager] Flow rules received: " + flowRules);
 }

this should be optimize.

private static Map<String, List<FlowRule>> flowRules = new ConcurrentHashMap<String, List<FlowRule>>();

 @Override
        public void configUpdate(List<FlowRule> value) {
            Map<String, List<FlowRule>> rules = FlowRuleUtil.buildFlowRuleMap(value);
            if (rules != null) {
                flowRules = new ConcurrentHashMap<>(rules);
            }
            RecordLog.info("[FlowRuleManager] Flow rules received: " + flowRules);
        }

sczyh30:
Refer https://github.com/alibaba/Sentinel/issues/1141#issuecomment-828074340

PR #973, opened 2019-08-07
FlowPropertyListener update flowRules,FlowRuleManager sometimes getRu…

zhuyou1234:
see  https://github.com/alibaba/Sentinel/issues/969

FlowRuleManager.FlowPropertyListener.configUpdate,first clear ,then putAll.between this，other thread invoke FlowRuleManager.getRules,return empty。i think is not reasonable。


sczyh30:
This has been improved via 36b162c
Thanks for contributing.

Issue #870, opened 2019-06-27
DataSource integration for Consul


N1cer:
我现在使用的配置中心是consul，但是文档中只找到了zk,nacos,apollo做push模式，那么我该如何做规则的持久化？

sczyh30:
Hi, Consul data-source is not supported yet. Would you like to contribute?

N1cer:
I think I can give it a try.

sczyh30:
Nice, looking forward to your pull request! You could refer to existing implementations like [sentinel-datasource-zookeeper](https://github.com/alibaba/Sentinel/tree/master/sentinel-extension/sentinel-datasource-zookeeper).

sczyh30:
Hi, any progress? :)

N1cer:
so sorry,I can‘t do this.

wavesZh:
I want to try it!

sczyh30:
> I want to try it!

@wavesZh Nice, looking forward to your PR~

sczyh30:
Resolved in #979

PR #979, opened 2019-08-09
DataSource integration for Consul

wavesZh:

add consul data source

 #870

### Describe how you did it

The data source first initial rules from a Consul during initialization.
Then it start a watcher to observe the updates of rule date and update to memory.

Because Consul do not provide http api to watch the update of KV, so i use a long polling and 
blocking queries of  the Consul's feature to deal it. When Querying data by index will blocking until change or timeout. If the index of the current query is larger than before, it means that the data has changed.

### Describe how to verify it
see UT


sentinel-extension/sentinel-datasource-consul/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.pszymczyk.consul</groupId>
+            <artifactId>embedded-consul</artifactId>
+            <version>${consul.process.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
```
cdfive: Here should be with test scope, as it's UT depenency.

wavesZh: resolved

wavesZh:
> Could you please add a `README.md` in the module for some instructions? You may refer to: https://github.com/alibaba/Sentinel/blob/master/sentinel-extension/sentinel-datasource-redis/README.md

👌

sczyh30:
And you could describe how the data source works in the PR description (e.g. it leveraged long polling).

sentinel-extension/sentinel-datasource-consul/src/main/java/com/alibaba/csp/sentinel/datasource/consul/ConsulDataSource.java
```diff
@@ -0,0 +1,125 @@
...
+    private volatile long lastIndex;
+
+    private int watchTimeout;
+
+    private ConsulKVWatcher watcher = new ConsulKVWatcher();
+
+    private ExecutorService watcherService = newSingleThreadExecutor(
+            new NamedThreadFactory("sentinel-consul-ds-update", true));
+
+    public ConsulDataSource(Converter<String, T> parser, String host, int port, String ruleKey, int watchTimeout) {
```
sczyh30: You have to provide the unit of the `watchTimeout` in the comment. It seems that the unit here should be `s` rather than commonly-used `ms`.

sczyh30: And you may also provide a constructor with the default port (8500).

sentinel-extension/sentinel-datasource-consul/src/main/java/com/alibaba/csp/sentinel/datasource/consul/ConsulDataSource.java
```diff
@@ -0,0 +1,125 @@
...
+            RecordLog.warn("[ConsulDataSource] Error when loading initial config", ex);
+        }
+    }
+
+    @Override
+    public String readSource() throws Exception {
+        if (this.client == null) {
+            throw new IllegalStateException("Consul has not been initialized or error occurred");
+        }
+        Response<GetValue> response = getValue(ruleKey);
```
sczyh30: The semantic for `readSource` (and derived `loadConfig`) requires the retrieving operation to return intermediately, rather than waiting. Maybe for `readSource` method, we could just send a normal request.

sentinel-extension/sentinel-datasource-consul/src/main/java/com/alibaba/csp/sentinel/datasource/consul/ConsulDataSource.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private class ConsulKVWatcher implements Runnable {
+        private boolean running = true;
+        @Override
+        public void run() {
+            while (running) {
+                Response<GetValue> response = getValue(ruleKey);
+                if (response == null) {
+                    try {
+                        TimeUnit.MILLISECONDS.sleep(watchTimeout);
```
sczyh30: sleep for seconds?

sczyh30:

wavesZh:
> Nice, thanks for contributing!

Thank you for your advice！

Issue #951, opened 2019-07-31
Support event listener of circuit breaker state change event


huangxfchn:
For example, if sentinel start to switch off, then can call back an interface to notify the application. The same is true when sentinel is out of circuit breaker mode.

sczyh30:
In fact, Sentinel has provided callback support for `StatisticSlot`:

- `ProcessorSlotEntryCallback`: callback when resource entry passed (`onPass`) or has been blocked (`onBlocked`)
- `ProcessorSlotExitCallback`: callback when resource entry successfully exited (`onExit`)

We can register your own callback via `StatisticSlotCallbackRegistry`. Maybe here we can just leverage the `onBlocked` callback to handle the rule trigger alarm. You could get the triggered rule from the `BlockException`.

You may also refer to #418

huangxfchn:
Looks a bit like.

I need to catch `DegradeException`, and determine the `count = = 1` representative to downgrade strategy, but `onBlocked` method should be invoked after the circuit breaker mode, this method was invoked every time during until out of fusing time window. Is that right?

So, what if I knew when to jump out of the circuit breaker mode.

sczyh30:
You mean notification for circuit breaker `CLOSE` event? This hasn't been supported yet, but it seems helpful. Maybe we could provide a listener registry for circuit breaker state change events. We could discuss the design here.

huangxfchn:
Yeah,  circuit breaker `START` and `CLOSE` event.

linlinisme:
I am doing something like that, may be it can assign to me.

sczyh30:
> I am doing something like that, maybe it can assign to me.

Nice, looking forward to your PR~

sczyh30:
The `CircuitBreakerStateChangeObserver` will be available in 1.8.0.

PR #984, opened 2019-08-13
support degrade status change listen

linlinisme:

Fix  https://github.com/alibaba/Sentinel/issues/951

### Describe how you did it

The proposal is based on Observer-module.
When the degrade rule fuse switch  open, Publisher will publish a switch open event.
   ```
  if (cut.compareAndSet(false, true)) {
            ResetTask resetTask = new ResetTask(this);
            pool.schedule(resetTask, timeWindow, TimeUnit.SECONDS);
            try {
                Event<RuleStatusContentWrapper<DegradeRule>> event = new Event<>(new RuleStatusContentWrapper<>(CIRCUIT_BREAK_START, this), DegradeRuleManager.class, System.currentTimeMillis());
                RuleStatusPublisherProvider.getPublisher().asynPublish(event);
            } catch (Exception ex) {
                RecordLog.warn("[RuleStatusPublisherProvider.getPublisher] publish event error", ex);
            }
        }
```
When fuse switch closed, Publisher publish a switch close event.

  ```
        public void run() {
            rule.passCount.set(0);
            rule.cut.set(false);
            try {
                Event<RuleStatusContentWrapper<DegradeRule>> event = new Event<>(new RuleStatusContentWrapper<>(CIRCUIT_BREAKER_CLOSE, rule), DegradeRuleManager.class, System.currentTimeMillis());
                RuleStatusPublisherProvider.getPublisher().asynPublish(event);
            } catch (Exception ex) {
                RecordLog.warn("[RuleStatusPublisherProvider.getPublisher] publish event error", ex);
            }
        }

```
The Subscriber will listen the event and make some  business logic processing.

```
public interface EventSubscriber {
    /**
     * Subscribe the event and make some  business logic processing
     *
     * @param event
     */
    void listen(Event event);
}
```

For convenient extension, the Subscriber support  SPI custom load.

`    private static final ServiceLoader<EventSubscriber> subscribers = ServiceLoader.load(EventSubscriber.class);`

**About the event publish:**
We support asynchronous publish and recommended this way.
Because It can avoid synchronous blocking When the Subscriber are doing some slow operation.

**About the Event design, It Composed of several parts:**
data: the event content, tell us what happen.
source: who produce this event.
time: when the event was produced.

### Describe how to verify it

Run test cases

### Special notes for reviews

I specially add try-catch code block when publish event, to avoid the publish/listen operation's
 un-control exception reflect the normal statistical


sentinel-core/src/main/java/com/alibaba/csp/sentinel/event/subscriber/EventSubscriber.java
```diff
@@ -0,0 +1,31 @@
...
+ * @author lianglin
+ * @since 1.7.0
+ */
+public interface EventSubscriber {
+    /**
+     * Subscribe the event and make some  business logic processing
+     *
+     * @param event
+     */
+    void listen(Event event);
```
sczyh30: For the circuit breaker, maybe it's more clear to be `onOpen` and `onClose`?

linlinisme: If I want to define a general subscription interface, it can be used for more than circuit breaker, such as flow limiting event or other events. Any suggestions?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/event/subscriber/EventSubscriberProvider.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @author lianglin
+ * @since 1.7.0
+ */
+public final class EventSubscriberProvider {
+
+    private static final ServiceLoader<EventSubscriber> subscribers = ServiceLoader.load(EventSubscriber.class);
+
+    public static List<EventSubscriber> provide() {
```
cdfive: It seems that the static `subscribers` loaded by SPI is not used, as the variable name is same.

sentinel-core/src/main/resources/META-INF/services/com.alibaba.csp.sentinel.event.subscriber.EventSubscriber
```diff
@@ -0,0 +1,2 @@
+# Default event subscriber
+com.alibaba.csp.sentinel.event.subscriber.DefaultEventSubscriber
```
cdfive: Two same file in `META-INF/services`, and the package of test case `DefaultEventSubscriberTest` is `com.alibaba.csp.sentinel.event`.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/event/subscriber/EventSubscriberProvider.java
linlinisme: > It seems that the static `subscribers` loaded by SPI is not used, as the variable name is same.

Thanks reminding. They use same names,I don't noticed that.....

PR #995, opened 2019-08-16
Improve RT statistic and exception tracing in Sentinel gRPC adapter

### LINKED ISSUE #289: [Bug] grpc adapter-Tracer.trace() can't success

chenledong:

Improve RT statistic and exception tracing in Sentinel gRPC adapter.

Fixes #289 

### Describe how you did it

- Override `ForwardingClientCall.SimpleForwardingClientCall#cancel` for some cases only call cancel.
  - Tracing exception and exit entry.
  - Add entry check for other cases may call both cancel and onClose.

### Describe how to verify it

Run the test cases and demo.

### Special notes for reviews

None.


sentinel-adapter/sentinel-grpc-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/grpc/SentinelGrpcServerInterceptor.java
```diff
@@ -39,67 +47,63 @@
...
             call.close(FLOW_CONTROL_BLOCK, new Metadata());
             return new ServerCall.Listener<ReqT>() {
             };
         } catch (RuntimeException e) {
-            //catch the RuntimeException startCall throws,
-            // entry is guaranteed to exit
+            // Catch the RuntimeException startCall throws,
+            // entry is guaranteed to exit.
             if (entry != null) {
                 entry.exit();
```
sczyh30: Here we could also trace the exception if the entry is not absent.

sczyh30:
Thanks for contributing!

PR #1004, opened 2019-08-21
sentinel console disable login

### LINKED ISSUE #997: Can sentinel console disable login?
## Issue Description

Type: *feature request*

Turn off login so I can use my console page more easily

Just like the `nacos console`

zhaoyuguang:
由于部分公司自己开发了控制台，不希望被Sentinel的安全认证拦截。因此Sentinel支持定制关闭登陆功能，`application.properties auth.enabled=false`即可完成 不需要登录

主要改造有两点
1：AuthService的实例化类 改为使用auth.enabled变量控制那个实例初始化
2：增加接口以便与模拟登陆。

Resolves #997


sentinel-dashboard/src/main/resources/application.properties
```diff
@@ -12,6 +12,7 @@ logging.pattern.file= %d{yyyy-MM-dd HH:mm:ss} [%thread] %-5level %logger{36} - %
 #auth settings
 auth.filter.exclude-urls=/,/auth/login,/auth/logout,/registry/machine,/version
 auth.filter.exclude-url-suffixes=htm,html,js,css,map,ico,ttf,woff,png
+auth.enabled=false
```
cdfive: How about making the default value true? since for most users, they may need login validation.

zhaoyuguang: Yes, I have modified

sentinel-dashboard/src/main/webapp/resources/app/scripts/directives/header/header.js
```diff
@@ -12,7 +12,20 @@ angular.module('sentinelDashboardApp')
...
-          $state.go('login');
+          AuthService.check().success(function (data) {
+            if (data.code == 0) {
+              $window.localStorage.setItem('session_sentinel_admin', {
+                username: data.data
+              });
+              if (data.data.id == 'FAKE_EMP_ID') {
+                document.getElementById('li-logout').style.display = 'none';
+              } else {
+                document.getElementById('li-logout').style.display = 'block';
```
zhaoyuguang: @cdfive Hi , `document.getElementById('li-logout').style.display` This JS code doesn't look very beautiful, can you guide me, how to use angularJS code this.

sczyh30:
@zhaoyuguang The CI indicates test failure. Could you please resolve it?

sentinel-dashboard/src/main/webapp/resources/app/scripts/directives/header/header.js
cdfive: @zhaixiaoxiang We can use `ng-show` directive to show or hide html element.
e.g. `<li ng-show="showLogout">`
The `showLogout` expression can be associated with a variable `$scope.showLogout` in js.
When `showLogout` is true/false, the element is visible/hide.

In header.js:
```javascript
controller: function ($scope, $state, $window, AuthService) {
        $scope.showLogout = true;
        ...
       if (data.data.id == 'FAKE_EMP_ID') {
           $scope.showLogout = true; // show  
       } else {
           $scope.showLogout = true; // hide
       }
}
```

cdfive: There is a small problem:

`auth.enable=false`

When user first visit the page(has no session data in localStorage), the logout button will be hide.
Then refresh page(press F5), the logout button is shown again.

This is due to that the show/hide code is inside the if judgement, `!$window.localStorage.getItem('session_sentinel_admin')`,
when refresh page(press F5), the localStorage has the session data, the if condition is not satisfied so the logout button isn't hide.

It's need to add some code in else branch to handle this case.

Now the code for putting data into localStorage is: 
```javascript
$window.localStorage.setItem('session_sentinel_admin', {
  username: data.data
});
```
When getting data by using `$window.localStorage.getItem('session_sentinel_admin')`,
The json data can't be retrieved, but '[object Object]' instead. 

Maybe we need to optimize it, 
one way is that use `JSON.stringify(xxx))` and `JSON.parse(xxx)`
to put/get data into/from localStorage.

zhaoyuguang: Thank you very much for your guidance and code review. I will try this design, but you may be @ wrong person. 😆

cdfive: > Thank you very much for your guidance and code review. I will try this design, but you may be @ wrong person.

OMG, I‘m sorry[facepalm].

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/auth/SimpleWebAuthServiceImpl.java
```diff
@@ -25,8 +25,8 @@
  * @author cdfive
  * @since 1.6.0
  */
-@Primary
 @Component
+@ConditionalOnProperty(name="auth.enabled", havingValue="true")
 public class SimpleWebAuthServiceImpl implements AuthService<HttpServletRequest> {
```
cdfive: There is a property `matchIfMissing` in `@ConditionalOnProperty`, default is false, maybe we can set it true here
in order to make it as default when no `auth.enabled` added in application.properties.

Besides, suggest that add a space between `=` for better code formatting.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/controller/AuthController.java
```diff
@@ -77,4 +82,16 @@
         request.getSession().invalidate();
         return Result.ofSuccess(null);
     }
+
+    @RequestMapping(value = "/check")
```
cdfive: Here "/check" and "/logout" `@RequestMapping`, we can optimize them to `@PostMapping`.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/controller/AuthController.java
```diff
@@ -77,4 +82,16 @@
...
+    public Result<?> check(HttpServletRequest request) {
+        AuthService.AuthUser authUser = authService.getAuthUser(request);
+        if (authUser == null) {
+            return Result.ofFail(-1, "Not logged in");
+        }
+        if (authService instanceof FakeAuthServiceImpl || request.getSession().getAttribute(SimpleWebAuthServiceImpl.WEB_SESSION_KEY) == null) {
+            request.getSession().setAttribute(SimpleWebAuthServiceImpl.WEB_SESSION_KEY, authUser);
```
cdfive: I think this judgement can be removed, since if use `FakeAuthServiceImpl`,  no need to create the backend session.
If use `SimpleWebAuthServiceImpl`,  the backend session should be created when login, not in the check method,
and the backend session has its own expire time.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/controller/AuthController.java
zhaoyuguang: Very right!

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/controller/AuthController.java
zhaoyuguang: Very right, after delete this code, I didn't find any problems when `session`
 has no `WEB_SESSION_KEY`.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/auth/SimpleWebAuthServiceImpl.java
zhaoyuguang: Yes

sentinel-dashboard/src/main/webapp/resources/app/scripts/directives/header/header.js
```diff
@@ -11,8 +11,22 @@ angular.module('sentinelDashboardApp')
...
+              $window.localStorage.setItem('session_sentinel_admin', JSON.stringify({
+                username: data.data
+              }));
+            } else {
+              $state.go('login');
+            }
+          });
+        }
+
+        if (JSON.parse($window.localStorage.getItem("session_sentinel_admin")).username.id == 'FAKE_EMP_ID') {
```
zhaoyuguang: Hi @cdfive, I have a problem and need your guidance. 🌹
When first enter the page, the code runs is shown in numerical order
image
So the L26 throw NPE and logout button can not hidden,
So I change code as :
```
if (!$window.localStorage.getItem("session_sentinel_admin")) {
  AuthService.check().success(function (data) {
    if (data.code == 0) {
      $window.localStorage.setItem('session_sentinel_admin', JSON.stringify({
        username: data.data
      }));
      handleLogout($scope, data.data.id)
    } else {
      $state.go('login');
    }
  });
} else {
  handleLogout($scope, JSON.parse($window.localStorage.getItem("session_sentinel_admin")).username.id)
}

function handleLogout($scope, id) {
  if (id == 'FAKE_EMP_ID') {
    $scope.showLogout = false;
  } else {
    $scope.showLogout = true;
  }
}
```
Am I writing the code correctly?

cdfive: @zhaoyuguang Yes, the code is okay!
The reason of NPE is that `AuthService.check()` is a asynchronous call, L26 execute firstly before success callback.

One more suggestion: 
We can use `JSON.stringify({data.data}))` in localStorage, just remove usename, because the data.data represent entire session data.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/auth/SimpleWebAuthServiceImpl.java
```diff
@@ -25,8 +26,9 @@
  * @author cdfive
  * @since 1.6.0
  */
-@Primary
 @Component
+@Primary
+@ConditionalOnProperty(name = "auth.enabled", matchIfMissing = true)
```
cdfive: Suggest remain `havingValue = "true"` , since if remove it, we can set `auth.enabled` to any expression, this is not clear semantic understanding.

In `FakeAuthServiceImpl`, `havingValue = "false"`, without `matchIfMissing = true`.

sentinel-dashboard/src/main/resources/application.properties
```diff
@@ -12,6 +12,7 @@ logging.pattern.file= %d{yyyy-MM-dd HH:mm:ss} [%thread] %-5level %logger{36} - %
 #auth settings
 auth.filter.exclude-urls=/,/auth/login,/auth/logout,/registry/machine,/version
 auth.filter.exclude-url-suffixes=htm,html,js,css,map,ico,ttf,woff,png
+# If auth.enabled=false, Sentinel console disable login
```
cdfive: The comment is good, suggest remain `auth.enabled=true` after the comment line, in order to make it easy for user to understand and modify.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/auth/SimpleWebAuthServiceImpl.java
zhaoyuguang: Conversely, if we add `havingValue = "true"`, auth.enabled is any expression except `"true"`, the Sentinel login will be disable 🤣

> To make the UT `GatewayApiControllerTest` passed, remove the `@ConditionalOnProperty`
> in `FakeAuthServiceImpl` is okay.

I thought so

sentinel-dashboard/src/main/webapp/resources/app/scripts/directives/header/header.js
zhaoyuguang: Thanks for the guidance to me, the unnecessary username has been cleared, So resolve conversation

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/controller/AuthController.java
zhaoyuguang: Thanks for the guidance, I fixed , please code review, if there is no problem, we can resolve this conversation.

sentinel-dashboard/src/main/resources/application.properties
zhaoyuguang: The problem same above

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/auth/SimpleWebAuthServiceImpl.java
cdfive: > Conversely, if we add havingValue = "true", auth.enabled is any expression except "true", the Sentinel login will be disable
Agree with you, it's okay and easy use for users.

sczyh30:
Could you please also execute the `npm run build` for the frontend and update the file `dist/js/app.js`?

zhaoyuguang:
> Could you please also execute the `npm run build` for the frontend and update the file `dist/js/app.js`?

Thanks for the guidance,`app.js` has been added

cdfive:
@zhaoyuguang Nice work! Thanks for your contribution. I think the implement is simple and clever.

sczyh30:
👍

PR #1007, opened 2019-08-21
Adaptive flow control (Suitable for user peak and callback peak)

### LINKED ISSUE #748: [RFC] Adaptive automatic flow control
## Issue Description

Type: *feature request*

It would be perfect if we can introduce intelligent mechanisms that do adaptive flow control automatically based on real-time metrics. Thus the services can react to burst traffic timely and become more resilient. Users might not need to configure rules anymore.

lym-ifae:
Nowadays users have to manually set the complex flow control rules for flow control. It is required for users to have a better understanding of Sentinel's flow control rules. It would be better if we can introduce intelligent mechanisms to do adaptive flow control automatically based on real-time metrics. Not only can the new mechanisms simplify users operations, but it can also make flow control smarter and adjust more quickly.

Fixes #748 

#### User peak
The scene of user peak should not have long delay. Therefore the algorithm for user peak is based on the method of token bucket. 

The adjustion of the token issuance speed refers to:
* The pass ratio which user expect to achieve;
* The RT which user expect to less than(`expectRt`);
* The CPU utilization.

The capacity of token bucket  is adjusted according to the maximum requests and the token issuance rate.

#### Callback peak
The scene of callback peak can have the delay longer. The algorithm is based on the rate limiter which implemented by Sentinel's `RateLimiterController`.

The adjustion of the flow rate refers to:
* The number of timeout requests;
* The CPU utilization.

#### PID control
This method is more suitable for user peak. And the adjustion of PID controller is also based on the token bucket. Users can try this algorithm by `rule.setGrade(RuleConstant.FLOW_ADAPTIVE_PID);`.

We use incremental PID controller to calculate the change of CPU utilization. Our goal is to keep the CPU utilization at a certain value. 

### Describe how to verify it
When it comes to the scene of user peak , see `TokenBucketDemo`;
When it comes to the scene of callback peak, see `AdaptiveRateDemo`.

### DIFF
--- a/sentinel-adapter/sentinel-web-servlet/README.md
+++ b/sentinel-adapter/sentinel-web-servlet/README.md
@@ -51,6 +51,8 @@ block handler (the `UrlBlockHandler` interface) and register to `WebCallbackMana
 The `UrlCleaner` interface is designed for clean and unify the URL resource.
 For REST APIs, you have to clean the URL resource (e.g. `/foo/1` and `/foo/2` -> `/foo/:id`), or
 the amount of context and resources will exceed the threshold.
+The `UrlCleaner` interface can also exclude unused URLs(e.g. `/exclude/1` -> `/exclude/:id` -> `""`).
+The URLs will be filtered and not be resource in this way.
 
 `RequestOriginParser` interface is useful for extracting request origin (e.g. IP or appName from HTTP Header)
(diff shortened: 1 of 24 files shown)


sczyh30:
Nice work. Could you please sync your branch with upstream `master` and resolve the conflicts (just choose *accept theirs*)?

PR #1014, opened 2019-08-26
"@Override" should be used on overriding and implementing methods

zhaoyuguang:
覆写方法，加@Override注解


sczyh30:
Thanks! As these two classes were picked originally from JDK, we usually don't modify these classes :)

Issue #728, opened 2019-05-05
Add version info in dashboard


fangjian0423:
## Issue Description

Type: feature request

### Describe what happened (or what feature you want)

Add version info in the left top of dashboard, like this:

Sentinel 控制台 1.6.0

Then i can know which version i am using in dashboard and don't need to get it with other ways.

sczyh30:
Hi, thanks for your suggestion. Maybe we can add the version info in the index page.

hisenyuan:
hello,i will add sentinel version in the index page.

sczyh30:
> hello,i will add sentinel version in the index page.

Nice, looking forward to your pull request!

hisenyuan:
`sentinel-dashboard/pom.xml`
I think sentinel dashboard version is the same as sentinel-parent, so I'll get it.
Do you agree with me?

sczyh30:
> `sentinel-dashboard/pom.xml`
> I think sentinel dashboard version is the same as sentinel-parent, so I'll get it.
> Do you agree with me?

That's okay.

hisenyuan:
> > `sentinel-dashboard/pom.xml`
> > I think sentinel dashboard version is the same as sentinel-parent, so I'll get it.
> > Do you agree with me?
> 
> That's okay.

https://github.com/alibaba/Sentinel/pull/800

sczyh30:
The backend API (GET `/version`) has been added by @hisenyuan via #800. Contributions for frontend are welcomed!

zhaoyuguang:
Hi will try to finish `the frontend part`

sczyh30:
Frontend resolved by #1015

PR #1015, opened 2019-08-26
Add version info in the left top of dashboard, like this: Sentinel 控制台 1.7.0

zhaoyuguang:
https://github.com/alibaba/Sentinel/issues/728
look like:
image


sentinel-dashboard/src/main/webapp/resources/app/scripts/services/auth_service.js
```diff
@@ -1,6 +1,13 @@
 var app = angular.module('sentinelDashboardApp');
 
 app.service('AuthService', ['$http', function ($http) {
+  this.version = function () {
```
sczyh30: Can we put this API to another service like `VersionService`? It's actually not related to auth.

zhaoyuguang: Thanks for the guidance, that is more clever, I fix it, please review 🌹

zhaoyuguang:
Resolves https://github.com/alibaba/Sentinel/issues/728

PR #1035, opened 2019-09-04
Fix the dashboard privileges check bug

lkxiaolou:
Fix the dashboard privileges check bug, from the interface AuthService, there is a comment on the authTarget function 

> @return if current user has the specific privileges to the target, return true, otherwise return false.

but in the dashboard controllers, there is not check on the authTarget's return value to verify privileges.

NONE

### Describe how you did it
Add checks for every authTarget's return.


lkxiaolou:
and update the fastjson version because of the security, see [https://github.com/alibaba/fastjson/releases](https://github.com/alibaba/fastjson/releases)

Issue #1039, opened 2019-09-09
Sentinel Statistic  should add time stamp to fix statistics window misalignment problem


linlinisme:
## Issue Description

Type: *bug report*

### Describe what happened (or what feature you want)
 Statistic should add time stamp

### Describe what you expected to happen

`StatisticNode.java
`
```
 @Override
    public void addPassRequest(int count) {
        rollingCounterInSecond.addPass(count);
        rollingCounterInMinute.addPass(count);
    }

    @Override
    public void addRtAndSuccess(long rt, int successCount) {
        rollingCounterInSecond.addSuccess(successCount);
        rollingCounterInSecond.addRT(rt);

        rollingCounterInMinute.addSuccess(successCount);
        rollingCounterInMinute.addRT(rt);
    }

    @Override
    public void increaseBlockQps(int count) {
        rollingCounterInSecond.addBlock(count);
        rollingCounterInMinute.addBlock(count);
    }

    @Override
    public void increaseExceptionQps(int count) {
        rollingCounterInSecond.addException(count);
        rollingCounterInMinute.addException(count);
    }
```
**Such as second、 minute multidimensional statistics should be consistent**.  **But now they take the current time of the system independently which may lead to some mistake statistics**. **Part of the statistics from last second(rollingCounterInSecond) is counted to the next second(rollingCounterInMinute)**。In particular, the request is in the last second and very close to the next second or high concurrency. 

And I think that is why  some sentinel users  they  **question   the QPS did not reach the threshold but the  request was limited by  the flow rule**。 This is because the last second's blocked request counted to next second.

ps: rollingCounterInMinute is the datasource fo dashboard statistics 

To fix this bug, I think the multidimensional statistics should based on same time stamp, then  the data can be counted in correct and same time window, no matter wherever and whenever.

xiejiashuai:
I occur the same problem too @sczyh30

PR #1040, opened 2019-09-11
[sentinal-1039] the multidimensional statistics should based on same …

complone:
the multidimensional statistics should based on same time stamp

### Describe how you did it
The request mentioned in #1040  is very close to the next second or high in the last second. I think rt should be monitored for each request when the status is changed.

### Special notes for reviews

the multidimensional statistics should based on same time stamp,the issue is #1039

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/Node.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/Node.java
@@ -164,16 +164,18 @@ public interface Node extends OccupySupport, DebugSupport {
     /**
      * Increase the block count.
      *
+     * @param rt Current timestamp, calculated based on this
      * @param count count to add
      */
-    void increaseBlockQps(int count);
+    void increaseBlockQps(long rt, int count);
 
     /**
      * Add the biz exception count.
      *
+     * @param rt Current timestamp, calculated based on this
      * @param count count to add
      */
-    void increaseExceptionQps(int count);
+    void increaseExceptionQps(long rt, int count);
 
     /**
      * Increase current thread count.
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/Node.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/Node.java
@@ -15,11 +15,11 @@
  */
 package com.alibaba.csp.sentinel.node;
 
-import java.util.Map;
-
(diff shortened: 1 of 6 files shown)


sczyh30:
Could you please illustrate your design in the PR description?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/Node.java
```diff
@@ -164,16 +164,18 @@
...
+    void increaseBlockQps(long rt, int count);
 
     /**
      * Add the biz exception count.
      *
+     * @param rt Current timestamp, calculated based on this
      * @param count count to add
      */
-    void increaseExceptionQps(int count);
+    void increaseExceptionQps(long rt, int count);
```
sczyh30: This could bring breaking changes. Please describe why this is really needed.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/StatisticNode.java
```diff
@@ -251,13 +251,13 @@ public void addRtAndSuccess(long rt, int successCount) {
     }
 
     @Override
-    public void increaseBlockQps(int count) {
+    public void increaseBlockQps(long rt, int count) {
         rollingCounterInSecond.addBlock(count);
```
sczyh30: The `rt` is not used?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/Node.java
complone: The request mentioned in #1039 is very close to the next second or high in the last second. I think rt should be monitored for each request when the status is changed.

sczyh30: > The request mentioned in #1040 is very close to the next second or high in the last second. I think rt should be monitored for each request when the status is changed.

In your implementation, the RT is not actually used?

complone: I will resubmit

sczyh30:
The CI indicates a build failure. Could you please resolve that?

complone:
> The CI indicates a build failure. Could you please resolve that?

Looking at places where you can refer to related functions

sczyh30:
Closing this PR due to unclear design. We may need to discuss this in #1039 first.

PR #1050, opened 2019-09-18
Improve the checking logic of SystemRule and enhance SystemRuleManager

sczyh30:

Improve `SystemRuleManager` and refactor the checking logic of `maxCpuUsage` strategy in SystemRule.

NONE

### Describe how you did it

- The BBR period for maxCpuUsage strategy is deprecated. Now Sentinel will just check the CPU usage value, which could avert surge of CPU usage quickly.
- Improve `SystemRuleManager` and `SystemStatusListener`.
- Add some test cases.

### Describe how to verify it

Run the test cases.

### Special notes for reviews

NONE


fangwentong:
> The BBR period for maxCpuUsage strategy is deprecated. Now Sentinel will just check the CPU usage value, which could avert surge of CPU usage quickly.

CPU usage value updates every seconds,  this change means that if CPU usage exceed threshold, all the traffic in next second will be blocked?

PR #1054, opened 2019-09-20
Add volatile for the stop flag of watcher

wavesZh:
The watcher's stop flag should be volatile


sczyh30:
Thanks for contributing!

Issue #1052, opened 2019-09-19
the order of slot  in SlotChainBuilder


wavesZh:
in `GatewaySlotChainBuilder` 

```java
public ProcessorSlotChain build() {
        ··· ··· 
        chain.addLast(new AuthoritySlot());
        chain.addLast(new SystemSlot());
        chain.addLast(new GatewayFlowSlot());
       ··· ···
}
``` 

in `DefaultSlotChainBuilder`

```java

public ProcessorSlotChain build() {
        ··· ··· 
        chain.addLast(new SystemSlot());
        chain.addLast(new AuthoritySlot());
       ··· ···
}
```

why the order of `SystemSlot` and `AuthoritySlot` is different?

sczyh30:
In fact, the order of the slots in `GatewaySlotChainBuilder` is better than `DefaultSlotChainBuilder`.

sczyh30:
Optimized in #1056

PR #1056, opened 2019-09-20
Change the position of SystemSlot and AuthoritySlot in soltchain

complone:
Change the position of SystemSlot and AuthoritySlot in soltchain

### DIFF
--- a/sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/slot/GatewaySlotChainBuilder.java
+++ b/sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/slot/GatewaySlotChainBuilder.java
@@ -44,8 +44,8 @@ public class GatewaySlotChainBuilder implements SlotChainBuilder {
         chain.addLast(new LogSlot());
         chain.addLast(new StatisticSlot());
         // Rule checking slot
-        chain.addLast(new SystemSlot());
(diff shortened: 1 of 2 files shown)


sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/slot/GatewaySlotChainBuilder.java
```diff
@@ -44,8 +44,8 @@ public ProcessorSlotChain build() {
         chain.addLast(new LogSlot());
         chain.addLast(new StatisticSlot());
         // Rule checking slot
-        chain.addLast(new AuthoritySlot());
         chain.addLast(new SystemSlot());
+        chain.addLast(new AuthoritySlot());
```
sczyh30: In fact, the order of the slots in the `GatewaySlotChainBuilder` is more appropriate. Maybe we could update the `DefaultSlotChainBuilder`.

sczyh30:
Related to #1052

sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/slot/GatewaySlotChainBuilder.java
complone: Okay, it's been corrected.

sczyh30:
Thanks for contributing!

PR #1111, opened 2019-10-23
Add init parameter to support unifying web context name in Sentinel Web CommonFilter

zhaoyuguang:
resolve https://github.com/alibaba/Sentinel/issues/1024

CommonFilter.java 增加可以控制context是统一名称或是为url的自定义配置参数


sczyh30:
The CI indicates test failure: https://travis-ci.org/alibaba/Sentinel/builds/601795740?utm_source=github_status&utm_medium=notification

Could you please resolve it?

sentinel-adapter/sentinel-web-servlet/src/main/java/com/alibaba/csp/sentinel/adapter/servlet/CommonFilter.java
```diff
@@ -78,7 +87,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             if (!StringUtil.isEmpty(target)) {
                 // Parse the request origin using registered origin parser.
                 String origin = parseOrigin(sRequest);
-                ContextUtil.enter(WebServletConfig.WEB_SERVLET_CONTEXT_NAME, origin);
+                ContextUtil.enter(webContextUnify ? WebServletConfig.WEB_SERVLET_CONTEXT_NAME : target, origin);
```
sczyh30: How about enabling `webContextUnify` by default? In most circumstances, users will not use flow control by web entrance, but removing this could reduce memory footprint.

zhaoyuguang: Enabling `webContextUnify` by `registration.addInitParameter(CommonFilter.WEB_CONTEXT_UNIFY, "false");`
Sorry? I do not understand what you mean.
抱歉，宿何老师 我没有get到你的意思

sczyh30: Sorry, I'd seen the wrong part. Ignore my comment...

sentinel-adapter/sentinel-web-servlet/src/main/java/com/alibaba/csp/sentinel/adapter/servlet/CommonFilter.java
```diff
@@ -42,18 +42,27 @@
...
+    public final static String WEB_CONTEXT_UNIFY = "WEB_CONTEXT_UNIFY";
     private final static String COLON = ":";
     private boolean httpMethodSpecify = false;
+    private boolean webContextUnify = true;
 
     @Override
     public void init(FilterConfig filterConfig) {
         httpMethodSpecify = Boolean.parseBoolean(filterConfig.getInitParameter(HTTP_METHOD_SPECIFY));
+        if (filterConfig.getInitParameter(WEB_CONTEXT_UNIFY) != null) {
+            webContextUnify = Boolean.parseBoolean(filterConfig.getInitParameter(WEB_CONTEXT_UNIFY));
```
sczyh30: What if an error occurs when `Boolean.parseBoolean(xxx)`? Maybe we need to handle the exception here?

zhaoyuguang: I thought so before, but I later found out that
The data type stored by filterConfig initParameter is <String, String>,
parseBoolean's method is like this
```
     Public static boolean parseBoolean(String s) {
         Return ((s != null) && s.equalsIgnoreCase("true"));
     }
```
After adding `filterConfig.getInitParameter(WEB_CONTEXT_UNIFY) != null`, I found it difficult to cause the parseBoolean method to be an error.

sczyh30: 👍

sczyh30:
Thanks for contributing!

PR #1140, opened 2019-11-06
treat it as root node when parentId is empty string

lhl4546:
One node should also be treated as root node when its parentId is an empty string. because FetchJsonTreeCommandHandler.java use fastjson to serialize tree node data, if WriteNullStringAsEmpty option in fastjson is set then null string value will be output as empty string, which will convert parentId of root node  to empty string, so this real root node can not be treated as root node.

### DIFF
--- a/sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/domain/ResourceTreeNode.java
+++ b/sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/domain/ResourceTreeNode.java
@@ -56,7 +56,7 @@ public class ResourceTreeNode {
             ResourceTreeNode node = fromNodeVo(vo);
             map.put(node.id, node);
             // real root
-            if (node.parentId == null) {
+            if (node.parentId == null || node.parentId.isEmpty()) {
                 root = node;
             } else if (map.containsKey(node.parentId)) {
                 map.get(node.parentId).children.add(node);


sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/domain/ResourceTreeNode.java
```diff
@@ -56,7 +56,7 @@ public static ResourceTreeNode fromNodeVoList(List<NodeVo> nodeVos) {
             ResourceTreeNode node = fromNodeVo(vo);
             map.put(node.id, node);
             // real root
-            if (node.parentId == null) {
+            if (node.parentId == null || node.parentId.isEmpty()) {
```
sczyh30: Maybe this can be simplified as `StringUtil.isEmpty(node.parentId)`.

sczyh30:
Could you please sign the CLA [here](https://cla-assistant.io/alibaba/Sentinel?pullRequest=1140)?

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/domain/ResourceTreeNode.java
lhl4546: yeah, but I think keep domain clean will be better

sczyh30:
Thanks for contributing!

PR #1148, opened 2019-11-07
Fix the bug that the Zuul adapter does not exit the entry with parameters

wavesZh:

[热点参数限流#基本使用](https://github.com/alibaba/Sentinel/wiki/%E7%83%AD%E7%82%B9%E5%8F%82%E6%95%B0%E9%99%90%E6%B5%81#%E5%9F%BA%E6%9C%AC%E4%BD%BF%E7%94%A8)
> 注意：若 entry 的时候传入了热点参数，那么 exit 的时候也一定要带上对应的参数（exit(count, args)），否则可能会有统计错误

NONE

### Describe how you did it

put holder which contains entry and params into Deque and exit with param


sentinel-adapter/sentinel-zuul-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/zuul/filters/EntryHolder.java
```diff
@@ -0,0 +1,41 @@
...
+import com.alibaba.csp.sentinel.Entry;
+
+/**
+ * @author wavesZh
+ */
+class EntryHolder {
+
+    private Entry entry;
+
+    private  Object[] params;
```
sczyh30: The fields could be final

sczyh30:
Thanks for contributing!

PR #1155, opened 2019-11-13
Fix type parameter consistency of cluster node command

agensi:


sczyh30:
Thanks for contributing!

Issue #1169, opened 2019-11-20
Bug when handling the end separator of the customized log directory in LogBase


sczyh30:
## Issue Description

Type: *bug report*

### Describe what happened (or what feature you want)

Sentinel supports user-defined customized log directory via the `csp.sentinel.log.dir` property item. When the given dir is not ended with the path separator (e.g. `/` in Linux), Sentinel will append the path separator to the given dir. This was broken due to a small mistake since 1.7.0:

https://github.com/alibaba/Sentinel/blob/d45ceb06dcc61356adc920772e142b81c1114f19/sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/LogBase.java#L97-L99

which should be:

```java
// Append the separator if needed
logBaseDir = addSeparator(logBaseDir);
```

### Describe what you expected to happen

Handle the separator correctly.

### Tell us your environment

Sentinel 1.7.0

zhaoyuguang:
I will finish it

PR #1172, opened 2019-11-21
Bug fix when handling the end separator of the customized log directory in LogBase

zhaoyuguang:
resolve https://github.com/alibaba/Sentinel/issues/1169


sczyh30:
Thanks!

Issue #1191, opened 2019-12-06
Support configuring project.name via the properties file


sczyh30:
## Issue Description

Type: *feature request*

### Describe what happened (or what feature you want)

Since [1.7.0](https://github.com/alibaba/Sentinel/releases/tag/1.7.0), Sentinel has supported configuration via the properties file (except the `project.name` property). The `project.name` property is not supported due to the legacy logic (compatible with the legacy default path `~/logs/csp/${appName}.properties`), which is rarely used. So we may deprecate the legacy logic in the next few versions and then support configuring `project.name` via the properties file.

Further discussions are welcomed!

sczyh30:
Also refer to #804

wavesZh:
I will finish it

sczyh30:
Good. Compatibility and implicit class dependency should be considered carefully.

PR #1192, opened 2019-12-07
Support configuring project.name via the properties file #1191

wavesZh:

Support configuring project.name via the properties file #1191

### Describe how you did it

add `csp.sentinel.project.name` . For Compatibility,  `AppNameUtil#getAppName()` and `SentinelConfig#getAppName` both can return correct value.

### Special notes for reviews


linlinisme:
@wavesZh Hey buddy, two things need to be noticed. First, consider compatibility with older versions. Second, consider the `project.name` when user don't manual explicit setting, we should not leave it empty. I think you can refer to the `SentinelConfig#resolveAppType`, migrate the `resolveAppName` from  `AppNameUtil` to `SentinelConfig`.  And `AppNameUtil` should not be used any more. What do you think ? @sczyh30

sczyh30:
Resolved in #1412 as it keeps compatibility. Thanks for contributing and looking forward more! 👍

PR #1204, opened 2019-12-16
SystemRule calculate process cpu usage to support application running in container

tianhaowhu:
SystemRule currently use the host's cpu usage, so it doesn't support the application running in the container such as docker, etc. This PR leverage both the process cpu usage and host's cpu usage to support application running in container.

NONE

### Describe how you did it

In the SystemStatusListener, I leverage JMX to get processCpuTime, processUpTime and cpuCores(the number of cpu cores, and it is the number of cpu cores which assigned to the container if application running in container).

Then the process cpu usage can be calculated as below:
processCpuUsage := delta(processCpuTime) / delta(processUpTime) / cpuCores

After that, we can use the max value between host's cpu usage and processCpuUsage to check the system status

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemStatusListener.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemStatusListener.java
@@ -16,6 +16,8 @@
 package com.alibaba.csp.sentinel.slots.system;
 
 import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
+import java.util.concurrent.TimeUnit;
 
 import com.alibaba.csp.sentinel.Constants;
 import com.alibaba.csp.sentinel.log.RecordLog;
@@ -33,6 +35,9 @@ public class SystemStatusListener implements Runnable {
 
     volatile String reason = StringUtil.EMPTY;
 
+    volatile long processCpuTime = 0;
+    volatile long processUpTime = 0;
+
     public double getSystemAverageLoad() {
         return currentLoad;
     }
@@ -46,6 +51,7 @@ public class SystemStatusListener implements Runnable {
         try {
             OperatingSystemMXBean osBean = ManagementFactory.getPlatformMXBean(OperatingSystemMXBean.class);
             currentLoad = osBean.getSystemLoadAverage();
+
             /*
              * Java Doc copied from {@link OperatingSystemMXBean#getSystemCpuLoad()}:</br>
              * Returns the "recent cpu usage" for the whole system. This value is a double in the [0.0,1.0] interval.
@@ -54,7 +60,21 @@ public class SystemStatusListener implements Runnable {
              * observed. All values between 0.0 and 1.0 are possible depending of the activities going on in the
              * system. If the system recent cpu usage is not available, the method returns a negative value.
              */
-            currentCpuUsage = osBean.getSystemCpuLoad();
+            double systemCpuUsage = osBean.getSystemCpuLoad();
+
+            // calculate process cpu usage to support application running in container environment
+            RuntimeMXBean runtimeBean = ManagementFactory.getPlatformMXBean(RuntimeMXBean.class);
+            long newProcessCpuTime = osBean.getProcessCpuTime();
+            long newProcessUpTime = runtimeBean.getUptime();
+            int cpuCores = osBean.getAvailableProcessors();
+            long processCpuTimeDiffInMs = TimeUnit.NANOSECONDS
+                    .toMillis(newProcessCpuTime - processCpuTime);
+            long processUpTimeDiffInMs = newProcessUpTime - processUpTime;
+            double processCpuUsage = (double) processCpuTimeDiffInMs / processUpTimeDiffInMs / cpuCores;
+            processCpuTime = newProcessCpuTime;
+            processUpTime = newProcessUpTime;
+
+            currentCpuUsage = Math.max(processCpuUsage, systemCpuUsage);
 
             if (currentLoad > SystemRuleManager.getSystemLoadThreshold()) {
                 writeSystemStatusLog();


sczyh30:
Hi, thanks for contributing. Could you please [sign the CLA here](https://cla-assistant.io/alibaba/Sentinel?pullRequest=1204)? And please make sure the email of your commits match your GitHub email. You may refer to the instruction [here](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user), or rebase your commits with the correct email.

---

感谢贡献，请将 commit 对应的 email 调整成与 GitHub 的 email 相匹配并 [确认一下 CLA](https://cla-assistant.io/alibaba/Sentinel?pullRequest=1204)。

sczyh30:
And could you please illustrate your design (and how you did it) in the PR description?

tianhaowhu:
> And could you please illustrate your design (and how you did it) in the PR description?

sure

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemStatusListener.java
```diff
@@ -54,7 +60,21 @@ public void run() {
              * observed. All values between 0.0 and 1.0 are possible depending of the activities going on in the
              * system. If the system recent cpu usage is not available, the method returns a negative value.
              */
-            currentCpuUsage = osBean.getSystemCpuLoad();
+            double systemCpuUsage = osBean.getSystemCpuLoad();
+
+            // calculate process cpu usage to support application running in container environment
+            RuntimeMXBean runtimeBean = ManagementFactory.getPlatformMXBean(RuntimeMXBean.class);
+            long newProcessCpuTime = osBean.getProcessCpuTime();
+            long newProcessUpTime = runtimeBean.getUptime();
```
jasonjoo2010: What's the difference between `uptime` and `systemtime` ? Avoid unexpected changing of system time?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemStatusListener.java
```diff
@@ -54,7 +60,21 @@ public void run() {
...
+            int cpuCores = osBean.getAvailableProcessors();
+            long processCpuTimeDiffInMs = TimeUnit.NANOSECONDS
+                    .toMillis(newProcessCpuTime - processCpuTime);
+            long processUpTimeDiffInMs = newProcessUpTime - processUpTime;
+            double processCpuUsage = (double) processCpuTimeDiffInMs / processUpTimeDiffInMs / cpuCores;
+            processCpuTime = newProcessCpuTime;
+            processUpTime = newProcessUpTime;
+
+            currentCpuUsage = Math.max(processCpuUsage, systemCpuUsage);
```
jasonjoo2010: And i think if we calculate the cpu usage by ourselves we can abandon `systemCpuUsage` totally.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemStatusListener.java
tianhaowhu: uptime won't be affected by unexpected changing of system time, the mechanism is similar to System.nanoTime()

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemStatusListener.java
tianhaowhu: They are both needed. The application performance could be affected if low processCpuUsage but high systemCpuUsage.

jasonjoo2010: > They are both needed. The application performance could be affected if low `processCpuUsage` but high `systemCpuUsage`.

In my logic in the scenario you mentioned slow down current application may not make things better and it also can use other dimension like `global avgRT` indicating whether should be limited traffic? After all the rules are written for applications.

eg.  the system cpu load keeps on about 0.8 but process use only 0.1

So what's your opinion?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemStatusListener.java
jasonjoo2010: > `uptime` won't be affected by unexpected changing of system time, the mechanism is similar to `System.nanoTime()`

Yeah it's okay for me if no portable issue on RuntimeMXBean

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemStatusListener.java
tianhaowhu: > > They are both needed. The application performance could be affected if low `processCpuUsage` but high `systemCpuUsage`.
> 
> In my logic in the scenario you mentioned slow down current application may not make things better and it also can use other dimension like `global avgRT` indicating whether should be limited traffic? After all the rules are written for applications.
> 
> eg. the system cpu load keeps on about 0.8 but process use only 0.1
> 
> So what's your opinion?

`systemCpuUsage` is appliable for physical machine which might running multiple applications, on the other hand, `processCpuUsage` is for docker environment which running single application. so I use the max value between them to compatiable for both physical and docker environments.

of course, there are many other dimensions, but I think we can keep `systemCpuUsage` and leave the options open to adopters.

jasonjoo2010: Okay i don't insist on it and we can keep it for backward compatibility.

jasonjoo2010:
Thanks for contribution and look forward for more!

brucelwl:
为什么不直接用double processCpuLoad = osBean.getProcessCpuLoad();获取呢?

PR #1208, opened 2019-12-19
Lhl4546 patch 1

lhl4546:
an appropriate map capacity can reduce resize


sczyh30:
Thanks for contributing!

PR #1245, opened 2020-01-12
Use Dubbo common constants.

echooymxq:
Use Dubbo‘s common constants, keep unified.


sczyh30:
Thanks for contributing!

PR #1247, opened 2020-01-13
Fix hardcoded file separator

### LINKED ISSUE #1239: Hardcoded file separator
https://github.com/alibaba/Sentinel/blob/fe23d5c11f2e14ab8da72aa26490e74d4f7eddf3/sentinel-demo/sentinel-demo-dynamic-file-rule/src/main/java/com/alibaba/csp/sentinel/demo/file/rule/FileDataSourceInit.java#L50

The path split symbol problem is different for different operating systems. Hard-coded file separators should not be used. Instead, use a platform-independent API provided by the language library,such as java.io.File.separator or java.lang.System.getProperty("file.separator")

xue8:
fix #1239

### Describe how you did it
`File.separator` look here. [Class File](https://docs.oracle.com/javase/8/docs/api/java/io/File.html)


sczyh30:
Thanks for contributing!

PR #1294, opened 2020-02-19
add volatile keyword for singleton object.

mantuliu:
add volatile keyword for singleton object.


sczyh30:
Thanks for contributing!

Issue #1031, opened 2019-09-02
I think there is something wrong with InMemoryMetricsRepository


shepherdviolet:
Today we are doing some testing. We found that only 1-3 minutes of data can be displayed in the real-time monitoring page (实时监控页). Especially when stress testing, only the last minute can be displayed. But the document shows that this page can display the last five minutes.

So I read the source code, I think this logic was originally designed to display the last five minutes, until the [PR #488](https://github.com/alibaba/Sentinel/pull/488) merged.

I am not sure that this PR have some problems, maybe the contributor has some reasonable considerations. But it does evict data earlier, before 5 minutes elapsed.

Class 'InMemoryMetricsRepository' uses ConcurrentLinkedHashMap to store metric data, and set capacity to MAX_METRIC_LIVE_TIME_MS (5 minutes), than define the weight of each element as 'currentTime - sampleTime'. I don't think this can achieve logic that only keeps the data for 5 minutes. The 'weight' should not be used like this. 

In this case, older data is heavier and takes up more space. If the timestamp of element is exactly equal to 'currentTime - 1000 * 60 * 5' , its weight is equal to MAX_METRIC_LIVE_TIME_MS. Fortunately, the weight is only calculated when the element is added. XD.
If the data is pulled from the client in a timely manner, the problem will be less serious.

The weight of each element is equal to one time unit (1000ms ?), will it be better? Although it will take up a lot of memory...

shepherdviolet:
Today, we provide our own optimized version, for reference only. we rewrites InMemoryMetricsRepository (-28%) and MetricFetcher (-35%). Two measures can reduce memory consumption by 63%. And can store 5 minutes of data correctly. But there is a problem with this implementation. The time unit of the statistical sample must be seconds. So just for reference! I am not considering doing PR for the time being.

Our commit: 
https://github.com/shepherdviolet/Sentinel/commit/f2f7c2d5b7d08230fb37198674fc3caec3934d77

Memory usage comparison==============================================

When dashboard first start--------------------------------------------------------------------

Eden Space:
   used     = 28319160 (27.00725555419922MB)
PS Old Generation
   used     = 27016016 (25.764480590820312MB)

After stress testing (Before optimization, Official version 1.6.3)-----------------------------

Eden Space:
   used     = 18968336 (18.089614868164062MB)
PS Old Generation
   used     = 494753768 (471.8339614868164MB)

num     #instances         #bytes  class name
----------------------------------------------
[...]

PR #1299, opened 2020-02-23
Fix InMemoryMetricsRepository can't keep  the last five minutes metrics data problem

linlinisme:

fix  https://github.com/alibaba/Sentinel/issues/1031  https://github.com/alibaba/Sentinel/issues/1174

### Describe how you did it

Rewrite  `java.util.LinkedHashMap#removeEldestEntry` make sure only keep data in lastest five minutes.
Use `ReentrantReadWriteLock` to avoid `ConcurrentModificationException` problem

### Describe how to verify it

com.alibaba.csp.sentinel.dashboard.repository.metric.InMemoryMetricsRepositoryTest


linlinisme:
> I have reviewed the changes and `synchronized` has been replaced with `ReadWriteLock` to gain better reading performance.
> 
> But I have a question that why time in milliseconds from dataset posted from clients is used as key?
> Is it necessary to cluster them into seconds for less objects with longer lifetime? Maybe memory pressure could be reduced i think.
> 
> Just a proposal and let's discuss more.
> 
> @sczyh30 @linlinisme

I think  cluster into seconds is accepted.  And I think we should agree on issues as soon as possible @jasonjoo2010 @sczyh30 .

jasonjoo2010:
> > I have reviewed the changes and `synchronized` has been replaced with `ReadWriteLock` to gain better reading performance.
> > But I have a question that why time in milliseconds from dataset posted from clients is used as key?
> > Is it necessary to cluster them into seconds for less objects with longer lifetime? Maybe memory pressure could be reduced i think.
> > Just a proposal and let's discuss more.
> > @sczyh30 @linlinisme
> 
> I think cluster into seconds is accepted. And I think we should agree on issues as soon as possible @jasonjoo2010 @sczyh30 .

I totally agree with you and how about we make the improvement and go on quickly. I think i can take charge this time so we will indeed get everything ASAP.

sentinel-dashboard/src/test/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepositoryTest.java
```diff
@@ -95,47 +127,41 @@ public void testExpireMetric() {
         inMemoryMetricsRepository.save(entry);
 
         List<MetricEntity> list = inMemoryMetricsRepository.queryByAppAndResourceBetween(
-            DEFAULT_EXPIRE_APP, DEFAULT_RESOURCE, now - 2 * EXPIRE_TIME, now + EXPIRE_TIME);
+                "testExpire", "testResource", now - EXPIRE_TIME, now);
```
cdfive: Why change the range here？I think previous `2 * EXPIRE_TIME` is better, since it proves that the `expireEntry` has not exist in repository now.

sczyh30:
> > I have reviewed the changes and `synchronized` has been replaced with `ReadWriteLock` to gain better reading performance.
> > But I have a question that why time in milliseconds from dataset posted from clients is used as key?
> > Is it necessary to cluster them into seconds for less objects with longer lifetime? Maybe memory pressure could be reduced i think.
> > Just a proposal and let's discuss more.
> > @sczyh30 @linlinisme
> 
> I think cluster into seconds is accepted. And I think we should agree on issues as soon as possible @jasonjoo2010 @sczyh30 .

+1 for aggregating by seconds.

cdfive:
+1 for aggregating by seconds.

sentinel-dashboard/src/test/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepositoryTest.java
linlinisme: > Why change the range here？I think previous `2 * EXPIRE_TIME` is better, since it proves that the `expireEntry` has not exist in repository now.

In this case what want to be proved is  entry which  exceed `EXPIRE_TIME` will be eliminate not the `2 * EXPIRE_TIME`.  `2 * EXPIRE_TIME` can pass the test case ， but not strictly.

cdfive: The last assertion has proved the case, and the new range here is more clearer semantically and strictly. I agree, it's ok.

linlinisme:
> LGTM

ok, I merge this git pull. And open anther issue about `aggregating by seconds`.

PR #1336, opened 2020-03-12
Highly performance for write metric

yangyiweigege:

to most enhance write metrics performance

#1174
Describe how you did it

seperate lock by app
Describe how to verify it

i have verified in our sentinenl producer enviroment six month.
Special notes for reviews

### DIFF
--- a/sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
+++ b/sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
@@ -53,7 +53,7 @@ public class InMemoryMetricsRepository implements MetricsRepository<MetricEntity
 
     @Override
     public void save(MetricEntity entity, String app) {
-        allMetrics.get(app).computeIfAbsent(entity.getResource(), e -> new LinkedHashMap<Long, MetricEntity>(100) {
+        allMetrics.computeIfAbsent(app, e -> new HashMap<>(16)).computeIfAbsent(entity.getResource(), e -> new LinkedHashMap<Long, MetricEntity>(100) {
             @Override
             protected boolean removeEldestEntry(Map.Entry<Long, MetricEntity> eldest) {
                 return eldest.getKey() < System.currentTimeMillis() - MAX_METRIC_LIVE_TIME_MS;
--- a/sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
+++ b/sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/MetricsRepository.java
```diff
@@ -30,14 +30,14 @@
...
-    void save(T metric);
+    void save(T metric, String app);
 
     /**
      * Save all metrics to the storage repository.
      *
      * @param metrics metrics to save
      */
-    void saveAll(Iterable<T> metrics);
+    void saveAll(Iterable<T> metrics, String app);
```
cdfive: In T, actually `MetricEntity` class, there is a `app` field, it seems that we needn't change the method signature.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
```diff
@@ -46,46 +47,68 @@
...
-        readWriteLock.writeLock().lock();
+
+        if (appLockMap.get(app) == null) { //DCL to the same app concurrent request
+
+            synchronized (this) {
+                if (appLockMap.get(app) == null) {
+                    appLockMap.put(app, new ReentrantReadWriteLock());
+                    allMetrics.put(app, new HashMap<>(256));
+                    lockAndWriteData(appLockMap.get(app), app, metrics);
+                } else { //
```
linlinisme: why synchronized `InMemoryMetricsRepository`  here ?  it seems synchronized  `appLockMap` will be better. And                     `allMetrics.put(app, new HashMap<>(256));` do duplicate operation which also do in `save` method

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
```diff
@@ -46,46 +47,68 @@
...
-        } finally {
-            readWriteLock.writeLock().unlock();
-        }
 
+    @Override
+    public void save(MetricEntity entity, String app) {
+        allMetrics.computeIfAbsent(app, e -> new HashMap<>(16)).computeIfAbsent(entity.getResource(), e -> new LinkedHashMap<Long, MetricEntity>(100) {
+            @Override
+            protected boolean removeEldestEntry(Map.Entry<Long, MetricEntity> eldest) {
+                return eldest.getKey() < System.currentTimeMillis() - MAX_METRIC_LIVE_TIME_MS;
```
linlinisme: `TimeUtil.currentTimeMillis()`  is a optimized version  for `System.currentTimeMillis()` , so here `TimeUtil.currentTimeMillis()` may be better

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
```diff
@@ -46,46 +47,68 @@
...
         } finally {
-            readWriteLock.writeLock().unlock();
+            readWriteLock.writeLock().unlock(); //release write lock
         }
     }
 
     @Override
     public List<MetricEntity> queryByAppAndResourceBetween(String app, String resource,
                                                            long startTime, long endTime) {
+        if (appLockMap.get(app) == null) {
```
linlinisme: It felt strange here

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
yangyiweigege: 1.first question. synchronize is avoid to repeat init the app-map。 it only affect when the request is first arrival
2.about duplicate operation. only to passed the unit test. have no actually meanings.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
yangyiweigege: if  the applocakMap can't get the lock.it means have no datas of the app(app-lock only init when the app first write request arrive).so  i return immediately。for example .a app named testAPP. if can't get the lock of testApp.means testAPP don't report any data.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
yangyiweigege: the synchonized  is only affect the app init.for example. we have 40 app.   the synchorinze only affect in 40 app firest report data.after init. after 40 locks generate, all code where arrival else.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
linlinisme: lock is used to control resource access, no equals to has data nor not.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
linlinisme: init a lock blocked all the  other invoker’s call is not a highly performance design.  Another optional design is defined another lock  which control the `app lock` init not the `InMemoryMetricsRepository` itself.  And we should keep the code clean and simple, don't  repeat your self.

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
yangyiweigege: ...is to avoid  NPE.read data need to add readLock.so i must check the applock when i need add readlock .

yangyiweigege:
@linlinisme  i will change the System.getCurrentTime to Timeunit.TimeUtil.currentTimeMillis().

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
yangyiweigege: 1.about duplicate operation.i will remove the duplication code in save method and update the unit test.
2.about the synchronize in init.   it  only  invoked in init period.  i think is could accepted. maybe i should init lock  in the app register ?

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/repository/metric/InMemoryMetricsRepository.java
linlinisme: read data throw NPE? 

 ```
  Map<String, LinkedHashMap<Long, MetricEntity>> resourceMap = allMetrics.get(app);
        if (resourceMap == null) {
            return results;
        }
        LinkedHashMap<Long, MetricEntity> metricsMap = resourceMap.get(resource);
        if (metricsMap == null) {
            return results;
        }
```

The origin design had deal with it.

yangyiweigege: 是这样的。就是因为我通过app名字分离了 读写锁。假如我要读取某个APP的数据 我肯定是需要获取这个app的读写锁的。 但是app的读写锁 初始化  是在写入的时候。 这里需要判断 这个app的读写锁是否生成。如果不判断。直接拿来用 就会空指针异常了。

sczyh30:
Any further feedback?

sczyh30:
Any more discussions regarding this PR?

Issue #757, opened 2019-05-13
Logging SPI for various logging libraries


shxz130:
the log is bad implement by  write String to assign log file，cannot we use slf4j or other implements such as we support event(EventBus or SpringApplicationEvent) to log，
is it more scalable？zhe sentinel‘s user can feel comfortable。

such as：
  EventBus.post(LogEvent log);

new  Listener(){
   doLog(){
    log.info();
  }
}

sczyh30:
Sentinel is a fundamental component so `sentinel-core` should not depend on any third-party logging libraries (which might bring dependency conflicit).

Maybe we can make the logging interface as SPI and provide different implementations (based on JUL or slf4j). Discussions are welcomed.

shxz130:
i'm every agree to support making the logging interface as SPI and provide different implementations

Warkeeper:
I think Logging SPI is pretty important since I find it's impossible to control the size and the count of log file "sentinel-block.log","sentinel-record.log" and "command-center.log".(They are all hard coded in codes)

sczyh30:
> I think Logging SPI is pretty important since I find it's impossible to control the size and the count of log file "sentinel-block.log", "sentinel-record.log" and "command-center.log".(They are all hardcoded in codes)

This can be improved. We might need more configuration items to control the different logs (e.g. record log, metric log and block log). For logging SPI we need more discussion (whether to support and how to support). @cdfive @jasonjoo2010 @linlinisme Any suggestions?

jasonjoo2010:
Indeed it's a meaningful feature for most users. We can provide it as an SPI protocol and make several implementations in separated modules including slf4j/log4j/jcl logging system forwarder, standard console output, etc.

Certainly it will also enable users to implement it themselves.

goudai:
能否支持采用slf4j做了日志记录框架API？

sczyh30:
Since the logger SPI has been introduced via #1265 and #1338, we may implement extension modules for popular logging libraries. Contributions are welcomed!

Umbrella issue: #1295

PR #1338, opened 2020-03-14
Polish logging SPI related code and add general JUL adapter for Logger SPI

sczyh30:

Polish logging SPI related code and add general JUL adapter for Logger SPI.

Related to #757

### Describe how you did it

* Move the legacy JUL methods from `LogBase` to `BaseJulLogger`.
* Add a `JavaLoggingAdapter` as the general JUL adapter for the Logger SPI, which makes it convenient to use (as the default logger).
* Add `LoggerSpiProvider` to resolve Logger SPI.
* Polish the logback demo.

### Describe how to verify it

Run the test cases and demo.

### Special notes for reviews

Be careful with circular class dependency and error handling in static block.


sczyh30:
Some of the log contents have not updated to the universal placeholder `{}` yet. I'll submit a new PR to do this.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/RecordLog.java
```diff
@@ -15,80 +15,74 @@
...
  *
  * @author youji.zj
+ * @author Eric Zhao
  */
-public class RecordLog extends LogBase {
-    private static com.alibaba.csp.sentinel.log.Logger log = null;
+public class RecordLog {
+
+    public static final String RECORD_LOGGER_NAME = "sentinelRecordLogger";
+    public static final String DEFAULT_RECORD_LOG_FILENAME = "sentinel-record.log";
```
cdfive: The name of final variables are not consistent with `CommandCenterLog` class, as they are in context of `RecordLog`, maybe the same as `RecordLog`'s `LOGGER_NAME`、`DEFAULT_LOG_FILENAME` is better? Seems `DEFAULT_LOG_FILENAME` can be `LOG_FILENAME`.

sentinel-demo/sentinel-demo-log-logback/pom.xml
```diff
@@ -12,12 +12,16 @@
     <artifactId>sentinel-demo-log-logback</artifactId>
 
     <dependencies>
+        <dependency>
+            <groupId>com.alibaba.csp</groupId>
+            <artifactId>sentinel-transport-common</artifactId>
+        </dependency>
+
```
cdfive: The dependency seems not used in demo, it is redundant

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/RecordLog.java
sczyh30: Good idea. I've polished the code, please check.

sentinel-demo/sentinel-demo-log-logback/pom.xml
sczyh30: Actually I'm planning to move the `CommandCenterLog` to `sentinel-transport-common`, as it's only used in transport command centers. This will be done in another PR, but the dependency could be reserved :)

sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/RecordLog.java
cdfive: OK, got it. Now `LogBase` is not a parent class, maybe another class name is better, like `LogConfig` or sth else. Besides, I found that the config parameter which can be customized by user is scattered in different classes(by global seaching "csp.sentinel." in sentinel-core module), if it is better put them into one config class, to make config clearly and convenient to found. And this need careful consideration and discussion, and I think maybe we can do it in the future.

sentinel-demo/sentinel-demo-log-logback/pom.xml
cdfive: OK, got it. Now LogBase is not a parent class, maybe another class name is better, like LogConfig or sth else. Besides, I found that the config parameter which can be customized by user is scattered in different classes(by global seaching "csp.sentinel." in sentinel-core module), if it is better put them into one config class, to make config clearly and convenient to found. And this need careful consideration and discussion, and I think maybe we can do it in the future.

sczyh30: Good idea. We may improve this later, but be careful of the compatibility. We could create an umbrella issue to discuss it :)

PR #1353, opened 2020-03-22
Fixes exit context when request is blocked in springwebmvc adaptor

### LINKED ISSUE #1350: springmvc adaptor didn't exit context when request is blocked
## Issue Description

Add one flow rule with callerA,  if one request from callerA is blocked,
there are some chances that the next requests from other caller will be still blocked,
since the context didn't exit and the web server(e.g. tomcat) may reuse thread.

Request from other caller shoud be passed,
and the context should be exit when one request is completed.

I fount it in `SentinelSpringMvcIntegrationTest#testOriginParser`, the third case with 
"/foo/3" url and no header, and wondering why it expects blocked.

Reproduce it in demo:
1.  Modify the `InterceptorConfig#Line62` in `sentinel-demo-spring-webmvc`
  `config.setOriginParser(request -> request.getHeader("S-user"));` to
  `config.setOriginParser(request -> request.getParameter("S-user"));`

2. Start web demo and dashboard
 
3. Add a flow rule for resource `GET:/hello`, with caller "aaa" and threshold "0"

4. Visit `/host:port/hello?S-user=aaa`, and it will be blocked by Sentinel,
   then visit `/host:port/hello?S-user=bbb` or `/host:port/hello` for many times, they will be blocked occasionally.

cdfive:

When a request is blocked, the current context hasn't exited, and it may affect other request,
since the context is `ThreadLocal` and web server will usually reuse thread.

Fixes #1350 

### Describe how you did it

When `AbstractSentinelInterceptor#preHandle` catch `BlockException`, the function return false
or throws Exception in it, then `afterCompletion` will not be executed, so context can't be exited.

Now `ContextUtil.exit();` is added in in catch Block to exit the context.

### Describe how to verify it

Run test case `SentinelSpringMvcIntegrationTest#testOriginParser`.

### Special notes for reviews

`ContextUtil.exit();` is added in front of `handleBlockException(request, response, e);`
since if add it after, when `handleBlockException` throws Exception, exit can't be executed.
another way is to use `try finally`, not very sure which is better, maybe it's concise.

In test case, small typo in comment has been fixed and add more comment.

PTAL.


sentinel-adapter/sentinel-spring-webmvc-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/spring/webmvc/AbstractSentinelInterceptor.java
```diff
@@ -67,6 +67,7 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
             }
             return true;
         } catch (BlockException e) {
+            ContextUtil.exit();
```
sczyh30: Maybe a `try-catch` here is better? Users may need the context in the block handler.

cdfive: Good idea, change to `try-finally`, which makes same effect as catch the Exeception of `handleBlockException` method and rethrow it, and it's suitable and concise here, please check.

sczyh30:
Thanks!

PR #1355, opened 2020-03-23
Regroup packages related to logging into a separate module

jasonjoo2010:
Follow-up to #1344 
Place packages related to logging into `sentinel-logging` because there may be several loggers to introduce in future.


sczyh30:
Thanks!

PR #1357, opened 2020-03-23
Improve standard output content in LogBase

zhaoyuguang:
resolve https://github.com/alibaba/Sentinel/issues/1348


sczyh30:
Thanks!

PR #411, opened 2019-01-12
Make the slot as the SPI interface and add a @Order annotation for define slot order

cdfive:
### Describe what this PR does / why we need it

Make it easier to add custorm slot and arrange its order for users.

### Does this pull request fix one issue?

Fixes #316 

### Describe how you did it

1.Add a `@Order` annotation and define the order of default slots from -9000 to -1000
NodeSelectorSlot: -9000
ClusterBuilderSlot: -8000
LogSlot: -7000
StatisticSlot: -6000
ParamFlowSlot: -5000
SystemSlot: -4000
AuthoritySlot: -3000
FlowSlot: -2000
DegradeSlot: -1000

2.Add a `loadOrderedInstanceList(Class<T>)` method in `SpiLoader`,
to Load the ordered instance list by SPI

3.Modify the `DefaultSlotChainBuilder` 
new XxxSlot changed to using `SpiLoader.loadOrderedInstanceList(ProcessorSlot.class);`

### Describe how to verify it

Add unit test

### Special notes for reviews

Please review its feasibility in this way, 
if it is okay, maybe the `InitOrder` can also use `Order` annotation instead, for similar logic,
the `InitOrder` and `InitExecutor` didn't modified at present.

PR #1365, opened 2020-03-27
Set default log level to INFO of JDK logging and improve log info and comment in SpiLoader

cdfive:

Relate to #411 
Improve the log, comment and test case of `SpiLoader`.

NONE

### Describe how you did it

Rename the method `loadDifferentInstanceListSorted` to `loadPrototypeInstanceListSorted` in `SpiLoader`, using `RecordLog.debug` in it instead of `RecordLog.info`, and use `{}` to print log.

Use a static block to log the ProcessorSlot info loaded by `SpiLoader` in `DefaultSlotChainBuilder`, which is in order to facilitate log troubleshooting.

### Describe how to verify it

Run test case; Start dashboard, view log file to check.

### Special notes for reviews

When using `RecordLog.debug` in `SpiLoader#loadPrototypeInstanceListSorted`, I found that the debug log messages are also printed in `sentinel-record.log` file. Feeling surprised, debug and check finally find one line in `BaseJulLogger#makeLoggingHandler`, which seems to be the reason.
`heliumRecordLog.setLevel(Level.ALL);`

As default log level is usually INFO, so I changed it to `heliumRecordLog.setLevel(Level.INFO);` and it works as expected.

PTAL.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/SpiLoader.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/SpiLoader.java
@@ -60,7 +60,7 @@ public final class SpiLoader {
                 return null;
             }
         } catch (Throwable t) {
-            RecordLog.warn("[SpiLoader] ERROR: loadFirstInstance failed", t);
+            RecordLog.error("[SpiLoader] ERROR: loadFirstInstance failed", t);
             t.printStackTrace();
             return null;
         }
@@ -95,7 +95,7 @@ public final class SpiLoader {
             }
             return defaultClass.newInstance();
         } catch (Throwable t) {
-            RecordLog.warn("[SpiLoader] ERROR: loadFirstInstanceOrDefault failed", t);
+            RecordLog.error("[SpiLoader] ERROR: loadFirstInstanceOrDefault failed", t);
             t.printStackTrace();
             return null;
         }
@@ -124,15 +124,15 @@ public final class SpiLoader {
             SpiOrderWrapper<T> w = null;
             for (T spi : serviceLoader) {
                 int order = SpiOrderResolver.resolveOrder(spi);
-                RecordLog.info("[SpiLoader] Found {} SPI: {} with order " + order, clazz.getSimpleName(),
-                    spi.getClass().getCanonicalName());
+                RecordLog.info("[SpiLoader] Found {} SPI: {} with order {}", clazz.getSimpleName(),
+                    spi.getClass().getCanonicalName(), order);
                 if (w == null || order < w.order) {
                     w = new SpiOrderWrapper<>(order, spi);
                 }
             }
             return w == null ? null : w.spi;
         } catch (Throwable t) {
-            RecordLog.warn("[SpiLoader] ERROR: loadHighestPriorityInstance failed", t);
+            RecordLog.error("[SpiLoader] ERROR: loadHighestPriorityInstance failed", t);
             t.printStackTrace();
             return null;
         }
@@ -167,7 +167,7 @@ public final class SpiLoader {
             }
             return list;
         } catch (Throwable t) {
-            RecordLog.warn("[SpiLoader] ERROR: loadInstanceList failed", t);
+            RecordLog.error("[SpiLoader] ERROR: loadInstanceList failed", t);
             t.printStackTrace();
             return new ArrayList<>();
         }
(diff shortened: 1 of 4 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/SpiLoader.java
```diff
@@ -219,16 +233,16 @@
...
                 // Since SPI is lazy initialized in ServiceLoader, we use online sort algorithm here.
                 SpiOrderResolver.insertSorted(orderWrappers, spi, order);
-                RecordLog.info("[SpiLoader] Found {0} SPI: {1} with order " + order, clazz.getSimpleName(),
-                        spi.getClass().getCanonicalName());
+                RecordLog.debug("[SpiLoader] Found {} SPI: {} with order {}", clazz.getSimpleName(),
+                        spi.getClass().getCanonicalName(), order);
             }
             List<T> list = new ArrayList<>();
             for (int i = 0; i < orderWrappers.size(); i++) {
                 list.add(i, orderWrappers.get(i).spi);
```
jasonjoo2010: Although these codes are not changed in this PR but i should point out there are two ways to optimize a little, just FYI:

1. `list` should have an initial capacity specified because we know it(Same to `orderWrappers`)
2. It's an in order copying so `add(T obj)` is enough and will make sense.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/SpiLoader.java
```diff
@@ -219,16 +233,16 @@
...
             }
             return list;
         } catch (Throwable t) {
-            RecordLog.warn("[SpiLoader] ERROR: loadDifferentInstanceListSorted failed", t);
+            RecordLog.warn("[SpiLoader] ERROR: loadPrototypeInstanceListSorted failed", t);
```
jasonjoo2010: `warn` and `ERROR` are a little confused(Although it's also not included in this PR actually).
And after several iterations enhancing logging of Sentinel is it still necessary to use `printStackTrace()` directly? We just make a decision and go.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/SpiLoader.java
cdfive: Good idea! It has been improved, please check.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/SpiLoader.java
cdfive: Yes, in Exception catch block, use `ERROR` is better, it has been fixed, please check.

> And after several iterations enhancing logging of Sentinel is it still necessary to use printStackTrace() directly? We just make a decision and go.

Use `printStackTrace()` directly doesn't look good indeed, maybe `printStackTrace()` is useful for users to view directly, since `RecordLog` log the message info file by default,
so I haven't remove it, and waiting for your confirm.I also wonder that if we could throw `RuntimeException` in the catch block. 
BTW, the catch block seems to be never happen or extremely small probability.

jasonjoo2010: > Yes, in Exception catch block, use `ERROR` is better, it has been fixed, please check.
> 
> > And after several iterations enhancing logging of Sentinel is it still necessary to use printStackTrace() directly? We just make a decision and go.
> 
> Use `printStackTrace()` directly doesn't look good indeed, maybe `printStackTrace()` is useful for users to view directly, since `RecordLog` log the message info file by default, so I haven't removed it, and waiting for your confirm.I also wonder that if we could throw `RuntimeException` in the catch block.
> BTW, the catch block seems to be never happen or extremely small probability.

Yeah actually i have no position here and we could keep it then

jasonjoo2010:
@sczyh30 And do you mind the design placing a static block in `DefaultSlotChainBuilder` only for logging?

If not it might be ok then.

sczyh30:
> @sczyh30 And do you mind the design placing a static block in `DefaultSlotChainBuilder` only for logging?
> 
> If not it might be ok then.

It's indeed confusing. Maybe we could just keep it in 1.7.2 and improve it in the next version (with the design of stateless slots).

cdfive:
@sczyh30 @jasonjoo2010 How about removing this static block. 
Since #411 has been merged, using `RecordLog.info`,may have many logs.

jasonjoo2010:
LGTM

jasonjoo2010:
@sczyh30 Do you have any further concerned now?

sczyh30:
Thanks!

PR #1367, opened 2020-03-28
控制台集群限流规则配置页面改进-集群规则隐藏高级配置项

zhaoyuguang:
Related to (2)(3) in #1311


cdfive:
How about threshold type of thread, do we need to hide it when in clusterMode？ @sczyh30 @zhaoyuguang

zhaoyuguang:
> How about threshold type of thread, do we need to hide it when in clusterMode？ @sczyh30 @zhaoyuguang

恩 issue 内有提及

sczyh30:
Thanks for contributing!

PR #1372, opened 2020-03-31
Clear useless data in ParameterMetric for removed rules

wavesZh:

we should clear metric for nonexistent param rule for some reasons:

1. Reduce memory footprint.
2. The service runs for a while then readd the rule after deleting the rule, if happening outbreak request，we expected that blocked, but in fact that will pass, because token bucket algorithm in ParamFlowChecker#passDefaultLocalCheck: 

```java
long toAddCount = (passTime * tokenCount) / (rule.getDurationInSec() * 1000);
long newQps = toAddCount + restQps > maxCount ? (maxCount - acquireCount) : (restQps + toAddCount - acquireCount);
```

NONE

### Describe how you did it

clear when loading rules.

### Describe how to verify it

unit test


sczyh30:
Actually it's not thread-safe when accessing the counter map. We may need to improve it later (using copy-on-write mechanism).

wavesZh:
I will improve it in this pr.

sczyh30:
> I will improve it in this pr.

Just open another PR to improve it (and it could be better to create an issue :)

sczyh30:
Thanks for contributing!

Issue #1322, opened 2020-03-06
新增API网关流控规则多条件匹配


jy2156121:
## Issue Description

Type:  *feature request*

### Describe what happened (or what feature you want)
一个规则目前只能针对一个条件进行设置
比如：Remote Host、Client IP、Header、URL参数、Cookie中的一个条件进行设置

### Describe what you expected to happen
一个规则可以设置多个条件的AND操作，比如Header的参数 p1=xx1 并且 p2=xx2 并且  Remote Host的参数 p3=xx3

### How to reproduce it (as minimally and precisely as possible)

1. 
2. 
3.

wavesZh:
I want to try it !

sczyh30:
> I want to try it !

Nice. You may present your design here and we could discuss with the community :)

wavesZh:
@sczyh30 

Because the gateway‘s rate processed by the converted value of the hotspot parameter condition, so the key is to convert the compound AND condition into the corresponding value in the `GatewayParamParser # parseParameterFor` method.
I plan to add `ParseStrategy` to handle compound AND conditions, and add paramItems in` GatewayParamFlowItem` to store compound AND parameter conditions.

----------------------------------------------------------

由于网关限流是通过热点参数条件转换后的value处理的，所以关键在 `GatewayParamParser#parseParameterFor` 方法中将复合AND条件转换成对应的value。
我打算新增 `ParseStrategy` 用于处理复合AND条件，并在 `GatewayParamFlowItem` 中新增paramItems 存放复合AND参数条件。

wavesZh:
@sczyh30 Can you give me some advice?

PR #1376, opened 2020-04-02
Add gateway composite pararm 

wavesZh:

#1322 

None

### Describe how you did it

Add a new parseStrategy to deal with composite param scene. 

### Describe how to verify it

unit test


sentinel-adapter/sentinel-api-gateway-adapter-common/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/common/rule/CompositeParam.java
```diff
@@ -0,0 +1,104 @@
+package com.alibaba.csp.sentinel.adapter.gateway.common.rule;
```
sczyh30, line 1: Could you please add the license header?

wavesZh: fixed

PR #1377, opened 2020-04-02
Force  api to fail if an incompitible fastjson found

jasonjoo2010:
Because structure in json format cannot be decoded correctly with `fastjson-1.2.11` or earlier we will initiate an error message if we found that.

So many include the proposal #736

### DIFF
--- a/sentinel-core/src/test/java/com/alibaba/csp/sentinel/util/VersionUtilTest.java
+++ b/sentinel-core/src/test/java/com/alibaba/csp/sentinel/util/VersionUtilTest.java
@@ -15,6 +15,8 @@
  */
 package com.alibaba.csp.sentinel.util;
 
+import static org.junit.Assert.*;
+
 import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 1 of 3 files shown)


sentinel-core/src/test/java/com/alibaba/csp/sentinel/util/VersionUtilTest.java
```diff
@@ -27,4 +29,19 @@ public void testGetDefaultVersion() {
...
+        assertEquals(0x0102ff04, VersionUtil.fromVersionString("1.2.255.4"));
+        assertEquals(0xffffffff, VersionUtil.fromVersionString("255.255.255.255"));
+        assertEquals(0, VersionUtil.fromVersionString("1.255.256.0"));
+        assertEquals(0, VersionUtil.fromVersionString("1.2."));
+        assertEquals(0x01000000, VersionUtil.fromVersionString("1"));
+        assertEquals(0x01020000, VersionUtil.fromVersionString("1.2"));
+        assertEquals(0, VersionUtil.fromVersionString("test"));
+        assertEquals(0, VersionUtil.fromVersionString("1.2.3-"));
+        assertEquals(0, VersionUtil.fromVersionString("1.2.3b"));
+    }
```
cdfive, line 46: I found that `fastjson` has versions like `1.2.60.sec09`
[https://mvnrepository.com/artifact/com.alibaba/fastjson](https://mvnrepository.com/artifact/com.alibaba/fastjson)

Do we need to handle with that...

sczyh30:
Thanks! We may conduct further discussions to eradicate the problem.

PR #1420, opened 2020-04-21
Refactor the mechanism of recording error (on completed)

### LINKED ISSUE #1419: Biz exceptions should be recorded on entry completed
## Issue Description

Type: *enhancement*

Biz exceptions should be recorded on entry completed (i.e. `exit`), rather than immediately after calling `Tracer.traceEntry(ex, e)`, so that the statistic data could be more accurate.

sczyh30:

Refactor the mechanism of recording error (record on completed).

Resolves #1419 

### Describe how you did it

- Refactor for `Entry`: Separate block exception (`blockError`) and biz exception (`error`). Use `setError(ex)` to mark error for this invocation.
- Refactor for `StatisticSlot`: record the biz exception on entry completed (if `error` is present)
- Polish Tracer with `entry.setError(ex)` mechanism.
- Polish `MetricExitCallback#exit` with the new error tracing mechanism.

### Describe how to verify it

Run the test cases.

### Special notes for reviews

The upcoming refactor of circuit breaking depends on the mechanism.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/Entry.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/Entry.java
@@ -15,6 +15,7 @@
  */
 package com.alibaba.csp.sentinel;
 
+import com.alibaba.csp.sentinel.slots.block.BlockException;
 import com.alibaba.csp.sentinel.util.TimeUtil;
 import com.alibaba.csp.sentinel.context.ContextUtil;
 import com.alibaba.csp.sentinel.node.Node;
@@ -44,6 +45,7 @@ import com.alibaba.csp.sentinel.context.Context;
  * @author qinan.qn
  * @author jialiang.linjl
  * @author leyou(lihao)
+ * @author Eric Zhao
  * @see SphU
  * @see Context
  * @see ContextUtil
@@ -52,18 +54,23 @@ public abstract class Entry implements AutoCloseable {
 
     private static final Object[] OBJECTS0 = new Object[0];
 
-    private long createTime;
+    private final long createTimestamp;
+    private long completeTimestamp;
+
     private Node curNode;
     /**
      * {@link Node} of the specific origin, Usually the origin is the Service Consumer.
      */
     private Node originNode;
+
     private Throwable error;
-    protected ResourceWrapper resourceWrapper;
+    private BlockException blockError;
+
+    protected final ResourceWrapper resourceWrapper;
 
     public Entry(ResourceWrapper resourceWrapper) {
         this.resourceWrapper = resourceWrapper;
-        this.createTime = TimeUtil.currentTimeMillis();
+        this.createTimestamp = TimeUtil.currentTimeMillis();
     }
 
     public ResourceWrapper getResourceWrapper() {
@@ -119,8 +126,17 @@ public abstract class Entry implements AutoCloseable {
      */
     public abstract Node getLastNode();
 
-    public long getCreateTime() {
-        return createTime;
+    public long getCreateTimestamp() {
+        return createTimestamp;
+    }
+
+    public long getCompleteTimestamp() {
+        return completeTimestamp;
+    }
+
+    public Entry setCompleteTimestamp(long completeTimestamp) {
+        this.completeTimestamp = completeTimestamp;
+        return this;
     }
 
     public Node getCurNode() {
@@ -129,6 +145,15 @@ public abstract class Entry implements AutoCloseable {
 
     public void setCurNode(Node node) {
         this.curNode = node;
+    }
+
+    public BlockException getBlockError() {
+        return blockError;
+    }
+
+    public Entry setBlockError(BlockException blockError) {
+        this.blockError = blockError;
+        return this;
     }
 
     public Throwable getError() {
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/Tracer.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/Tracer.java
@@ -18,11 +18,7 @@ package com.alibaba.csp.sentinel;
 import com.alibaba.csp.sentinel.context.Context;
 import com.alibaba.csp.sentinel.context.ContextUtil;
 import com.alibaba.csp.sentinel.context.NullContext;
-import com.alibaba.csp.sentinel.metric.extension.MetricExtensionProvider;
-import com.alibaba.csp.sentinel.node.ClusterNode;
-import com.alibaba.csp.sentinel.node.DefaultNode;
 import com.alibaba.csp.sentinel.slots.block.BlockException;
-import com.alibaba.csp.sentinel.metric.extension.MetricExtension;
 import com.alibaba.csp.sentinel.util.AssertUtil;
 
 /**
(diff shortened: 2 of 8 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/statistic/StatisticSlot.java
```diff
@@ -171,4 +150,16 @@ public void exit(Context context, ResourceWrapper resourceWrapper, int count, Ob
...
+        if (node == null) {
+            return;
+        }
+        node.addRtAndSuccess(rt, batchCount);
+        node.decreaseThreadNum();
+
+        if (error != null && !(error instanceof BlockException)) {
+            node.increaseExceptionQps(batchCount);
+        }
+    }
```
cdfive, line 164: This `recordCompleteFor` makes it more clear to record for related node. 
In `StatisticSlot#exit` of previous version, there is a judgement for max rt: 
```java
int maxStatisticRt = SentinelConfig.statisticMaxRt();
if (rt > maxStatisticRt) {
    rt = maxStatisticRt;
}
```
Is it still needed here?
I found that in `MetricBucket`
```java
public void addRT(long rt) {
    add(MetricEvent.RT, rt);
    // Not thread-safe, but it's okay.
    if (rt < minRt) {
        minRt = rt;
    }
}
```
If a RT which is greater than default maxRT, it will still be added with actual value,
not quite sure about it, and the minRT seems not the same meaning as judgement of maxRT.

sczyh30: > This `recordCompleteFor` makes it more clear to record for related node.
> In `StatisticSlot#exit` of previous version, there is a judgement for max rt:
> 
> ```java
> int maxStatisticRt = SentinelConfig.statisticMaxRt();
> if (rt > maxStatisticRt) {
>     rt = maxStatisticRt;
> }
> ```
> 
> Is it still needed here or some other place?

It's not needed anymore. The maxStatisticRt was designed to avoid affecting the avgRt by extremely-slow requests, which could undermine the effect of RT-based circuit breaking. Now we'll record the original response time (so that the monitoring data could be more accurate). The RT-based circuit breaking strategy will be refactored in the incoming PRs so that it will not suffer from high avgRt.

> I found that in `MetricBucket`
> 
> ```java
> public void addRT(long rt) {
>     add(MetricEvent.RT, rt);
>     // Not thread-safe, but it's okay.
>     if (rt < minRt) {
>         minRt = rt;
>     }
> }
> ```
> 
> If a RT which is greater than default maxRT, it will still be added with actual value,
> not quite sure about it, and the minRT seems not the same meaning as judgement of maxRT.

It doesn't matter for minRt, which is used for adaptive system protection during the BBR period. The maxStatisticRt could be deprecated later.

cdfive: Got it, thanks!

sentinel-core/src/main/java/com/alibaba/csp/sentinel/Entry.java
```diff
@@ -52,14 +54,19 @@
 
     private static final Object[] OBJECTS0 = new Object[0];
 
-    private long createTime;
+    private final long createTime;
+    private long completeTimestamp;
```
cdfive, line 58: How about keeping the same suffix of the two fields?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/Tracer.java
```diff
@@ -39,87 +35,81 @@
...
-        if (entry == null || entry.getCurNode() == null) {
+
+        if (context == null || context instanceof NullContext) {
             return;
         }
-
-        DefaultNode curNode = (DefaultNode)entry.getCurNode();
-        traceExceptionToNode(e, count, entry, curNode);
+        traceEntryInternal(e, context.getCurEntry());
     }
```
cdfive, line 92: The parameter `count` now is unused, do we have no scenes that need to use it? 
As `@Deprecated` is marked, maybe the caller `SentinelReactorSubscriber#L150` and `ReactorSphU#L60` use the newer method is better.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/Tracer.java
```diff
@@ -39,87 +35,81 @@
...
-        ClusterNode clusterNode = curNode.getClusterNode();
-        if (clusterNode == null) {
+    private static void traceEntryInternal(/*@NeedToTrace*/ Throwable e, Entry entry) {
+        if (entry == null) {
             return;
         }
-        clusterNode.trace(t, count);
+
+        entry.setError(e);
     }
```
cdfive, line 113: `clusterNode.trace(t, count);` is removed here, and is it right understanding that the `increaseExceptionQps` now is in the `StatisticSlot#recordCompleteFor` method. But the operation of `cluserNode` seems missing, is it affects?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/Tracer.java
sczyh30: As it's related to other adapter modules, it might be better to make changes in a new PR.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/Tracer.java
sczyh30: It's included in `recordCompleteFor` method (for both DefaultNode and its ClusterNode)

sentinel-core/src/main/java/com/alibaba/csp/sentinel/Entry.java
sczyh30: Good idea. I've updated `createTime` to `createTimestamp`.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/Tracer.java
cdfive: Okay, I found it in `DefaultNode#increaseExceptionQps`.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/statistic/StatisticSlot.java
```diff
@@ -115,52 +115,31 @@ public void entry(Context context, ResourceWrapper resourceWrapper, DefaultNode
...
-            node.increaseExceptionQps(count);
-            if (context.getCurEntry().getOriginNode() != null) {
-                context.getCurEntry().getOriginNode().increaseExceptionQps(count);
-            }
-
-            if (resourceWrapper.getEntryType() == EntryType.IN) {
-                Constants.ENTRY_NODE.increaseExceptionQps(count);
-            }
             throw e;
         }
```
cdfive, line 122: I found that with the new design, we needn't increaseExceptionQps here(Unexpected error), which looks nicer.
By the way, `context.getCurEntry().setError(e)` looks equivalent to `Tracer.traceContext(e, context)`, and the difference is that traceContext has `shouldTrace` judgement.
Since almost never happens here, both are okay, right?

PR #1429, opened 2020-04-25
If the lock acquisition fails, an IllegalMonitorStateException will be thrown.

Yanghf123:
…e thrown.

If the lock acquisition fails, an IllegalMonitorStateException will be thrown.

NONE

### Describe how to verify it
Run the test cases.


sczyh30:
Thanks for contributing!

Issue #1416, opened 2020-04-20
sentinel-apache-dubbo-adapter Entry leak cause full GC


ecyuan:
sentinel-apache-dubbo-adapter-1.7 cause full gc.

When dubbo request grow, full gc regularly.

Reason:
https://github.com/alibaba/Sentinel/blob/release-1.7/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboConsumerFilter.java#L68

https://github.com/alibaba/Sentinel/blob/release-1.7/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java#L70

Provider and Consumer use the same key to store CtEntry.
A = client
B = server dubbo provider
C = server dubbo provider
For example:  A --> call B (call C)

When onResponse
https://github.com/alibaba/Sentinel/blob/release-1.7/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/BaseSentinelDubboFilter.java#L65

C remove DubboUtils.DUBBO_INTERFACE_ENTRY_KEY first, then B can't find the key, so CtEntry object create by SentinelDubboProviderFilter escape. so heap will grow.

sczyh30:
Thanks for reporting! It's indeed a bug. Would you like to contribute a PR to fix it? You may refer to the [contribution guidelines](https://github.com/alibaba/Sentinel/blob/master/CONTRIBUTING.md).

linlinisme:
Assign to me, I know this guy @ecyuan

PR #1431, opened 2020-04-25
[1416]  fix sentinel-apache-dubbo-adapter full GC problem

linlinisme:

fix https://github.com/alibaba/Sentinel/issues/1416

### Describe how you did it

The reason for  Full GC is  **CtEntry**   stored in **RpcContext** can't be clear correctly when application process a nested dubbo-rpc. For example: [A(client)] -> [B(provider) -> C(provider)].

 As we know that, one-time dubbo-rpc in sentinel will create two CtEntry: **interfaceEntry**, **methodEntry**. According to design, at the beginning of  rpc  we create refer-CtEntries and  put them it Dubbo#RpcContext, clear the Dubbo#RpcContext's refer-CtEntries when finised. It does well when just do a direct rpc. But while lauch a  nested dubbo-rpc **[B(provider) -> C(provider)]**), The B(provider)'s CtEntries will not be clear correctly. Because B and C use the fixed-same key for store CtEntries. C  remove the key firstly, B will not find the key ,skip the clear logic   lead to  ` ContextUtil.exit()`  take no effect  the key point lead to Full GC.

Store Logic
```
interfaceEntry = SphU.entry(interfaceResourceName, ResourceTypeConstants.COMMON_RPC, EntryType.IN);
rpcContext.set(DubboUtils.DUBBO_INTERFACE_ENTRY_KEY, interfaceEntry);
methodEntry = SphU.entry(methodResourceName, ResourceTypeConstants.COMMON_RPC, EntryType.IN, invocation.getArguments());
rpcContext.set(DubboUtils.DUBBO_METHOD_ENTRY_KEY, methodEntry);
```

Clear Logic
```
Entry interfaceEntry = (Entry) RpcContext.getContext().get(DubboUtils.DUBBO_INTERFACE_ENTRY_KEY);
Entry methodEntry = (Entry) RpcContext.getContext().get(DubboUtils.DUBBO_METHOD_ENTRY_KEY);
  if (methodEntry != null) {
            Tracer.traceEntry(throwable, methodEntry);
            methodEntry.exit();
            RpcContext.getContext().remove(DubboUtils.DUBBO_METHOD_ENTRY_KEY);
 }
 if (interfaceEntry != null) {
            Tracer.traceEntry(throwable, interfaceEntry);
            interfaceEntry.exit();
            RpcContext.getContext().remove(DubboUtils.DUBBO_INTERFACE_ENTRY_KEY);
  }
```

Sentinel ContextUtil is a Thread-Local Design, Dubbo use thread-pool manager dubbo-rpc thread.
if ` ContextUtil.exit()` don't take effect, when new Request come in the _**context**_  the  more entries will refer which can't be collected by  gc.  Over time the full gc will occr.

And  the solution is also simple: just assign different key to different request.

### DIFF
--- a/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/BaseSentinelDubboFilter.java
+++ b/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/BaseSentinelDubboFilter.java
@@ -81,7 +81,6 @@ public abstract class BaseSentinelDubboFilter extends ListenableFilter {
             }
             RpcContext.getContext().remove(methodResourceName);
         }
-        URL url = invoker.getUrl();
         if (CommonConstants.PROVIDER_SIDE.equals(invoker.getUrl().getParameter(CommonConstants.SIDE_KEY))) {
             ContextUtil.exit();
         }
--- a/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/BaseSentinelDubboFilter.java
+++ b/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/BaseSentinelDubboFilter.java
@@ -20,7 +20,6 @@ import com.alibaba.csp.sentinel.Entry;
 import com.alibaba.csp.sentinel.Tracer;
 import com.alibaba.csp.sentinel.adapter.dubbo.config.DubboConfig;
 import com.alibaba.csp.sentinel.context.ContextUtil;
-import org.apache.dubbo.common.URL;
 import org.apache.dubbo.common.constants.CommonConstants;
 import org.apache.dubbo.rpc.Invocation;
 import org.apache.dubbo.rpc.Invoker;
@@ -39,7 +38,47 @@ public abstract class BaseSentinelDubboFilter extends ListenableFilter {
         this.listener = new SentinelDubboListener();
     }
 
-    static class SentinelDubboListener implements Listener {
+
+    private void traceAndExit(Throwable throwable, Invoker invoker, Invocation invocation) {
+        String methodResourceName = getMethodName(invoker, invocation);
+        Entry[] entries = (Entry[]) RpcContext.getContext().get(methodResourceName);
+        if (entries != null) {
+            Entry interfaceEntry = entries[0];
+            Entry methodEntry = entries[1];
+            if (methodEntry != null) {
+                Tracer.traceEntry(throwable, methodEntry);
+                methodEntry.exit();
+            }
+            if (interfaceEntry != null) {
+                Tracer.traceEntry(throwable, interfaceEntry);
+                interfaceEntry.exit();
+            }
+            RpcContext.getContext().remove(methodResourceName);
+        }
+        if (CommonConstants.PROVIDER_SIDE.equals(invoker.getUrl().getParameter(CommonConstants.SIDE_KEY))) {
+            ContextUtil.exit();
+        }
+    }
+
+    /**
+     * Get method name of dubbo rpc
+     *
+     * @param invoker
+     * @param invocation
+     * @return
+     */
+    abstract String getMethodName(Invoker invoker, Invocation invocation);
+
+    /**
+     * Get interface name of dubbo rpc
+     *
+     * @param invoker
+     * @return
+     */
+    abstract String getInterfaceName(Invoker invoker);
+
+
+    private class SentinelDubboListener implements Listener {
 
         public void onResponse(Result appResponse, Invoker<?> invoker, Invocation invocation) {
             onSuccess(appResponse, invoker, invocation);
@@ -63,26 +102,5 @@ public abstract class BaseSentinelDubboFilter extends ListenableFilter {
             traceAndExit(t, invoker, invocation);
         }
 
-    }
-
-    static void traceAndExit(Throwable throwable, Invoker invoker, Invocation invocation) {
-        String methodResourceName = DubboUtils.getResourceName(invoker, invocation, DubboConfig.getDubboConsumerPrefix());
-        Entry[] entries = (Entry[]) RpcContext.getContext().get(methodResourceName);
-        if (entries != null) {
-            Entry interfaceEntry = entries[0];
-            Entry methodEntry = entries[1];
-            if (methodEntry != null) {
-                Tracer.traceEntry(throwable, methodEntry);
-                methodEntry.exit();
-            }
-            if (interfaceEntry != null) {
-                Tracer.traceEntry(throwable, interfaceEntry);
-                interfaceEntry.exit();
-            }
-            RpcContext.getContext().remove(methodResourceName);
-        }
-        if (CommonConstants.PROVIDER_SIDE.equals(invoker.getUrl().getParameter(CommonConstants.SIDE_KEY))) {
-            ContextUtil.exit();
-        }
     }
 }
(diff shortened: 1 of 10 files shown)


sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java
```diff
@@ -61,15 +61,13 @@ public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcExcept
...
+            String interfaceResourceName = DubboUtils.getInterfaceName(invoker);
             // Only need to create entrance context at provider side, as context will take effect
             // at entrance of invocation chain only (for inbound traffic).
             ContextUtil.enter(methodResourceName, application);
             interfaceEntry = SphU.entry(interfaceResourceName, ResourceTypeConstants.COMMON_RPC, EntryType.IN);
-            rpcContext.set(DubboUtils.DUBBO_INTERFACE_ENTRY_KEY, interfaceEntry);
             methodEntry = SphU.entry(methodResourceName, ResourceTypeConstants.COMMON_RPC, EntryType.IN, invocation.getArguments());
-            rpcContext.set(DubboUtils.DUBBO_METHOD_ENTRY_KEY, methodEntry);
+            rpcContext.set(methodResourceName, new Entry[]{interfaceEntry, methodEntry});
             return invoker.invoke(invocation);
```
cdfive, line 71: It's good adding a `DubboUtils#getInterfaceName` for resue.
Now the interfaceResource and methodResource key are merged into one key, but the key name is methodResourceName, is it better using a fixed key? @sczyh30 How about your opinion?

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/BaseSentinelDubboFilter.java
```diff
@@ -38,43 +42,46 @@ public BaseSentinelDubboFilter() {
...
-            methodEntry.exit();
-            RpcContext.getContext().remove(DubboUtils.DUBBO_METHOD_ENTRY_KEY);
-        }
-        if (interfaceEntry != null) {
-            Tracer.traceEntry(throwable, interfaceEntry);
-            interfaceEntry.exit();
-            RpcContext.getContext().remove(DubboUtils.DUBBO_INTERFACE_ENTRY_KEY);
+    static void traceAndExit(Throwable throwable, Invoker invoker, Invocation invocation) {
+        String methodResourceName = DubboUtils.getResourceName(invoker, invocation, DubboConfig.getDubboConsumerPrefix());
+        Entry[] entries = (Entry[]) RpcContext.getContext().get(methodResourceName);
```
cdfive, line 70: As both `SentinelDubboProviderFilter` and `SentinelDubboConsumerFilter` share this method, ` DubboConfig.getDubboConsumerPrefix()` may not correct, when providerPrefix, consumerPrefix have different settings.
If use methodResourceName as key, here we may need to differentiate provider and consumer.

linlinisme: emm.. I don't notice this. I will modify it later.

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java
```diff
@@ -60,16 +70,14 @@ public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcExcept
...
+            String methodResourceName = getMethodName(invoker, invocation);
+            String interfaceResourceName = getInterfaceName(invoker);
             // Only need to create entrance context at provider side, as context will take effect
             // at entrance of invocation chain only (for inbound traffic).
             ContextUtil.enter(methodResourceName, application);
             interfaceEntry = SphU.entry(interfaceResourceName, ResourceTypeConstants.COMMON_RPC, EntryType.IN);
-            rpcContext.set(DubboUtils.DUBBO_INTERFACE_ENTRY_KEY, interfaceEntry);
             methodEntry = SphU.entry(methodResourceName, ResourceTypeConstants.COMMON_RPC, EntryType.IN, invocation.getArguments());
-            rpcContext.set(DubboUtils.DUBBO_METHOD_ENTRY_KEY, methodEntry);
+            rpcContext.set(methodResourceName, new Entry[]{interfaceEntry, methodEntry});
```
sczyh30, line 80: This may cause leak of `interfaceEntry` when the `methodEntry` is blocked. Refer https://github.com/alibaba/Sentinel/pull/1124#discussion_r354810059

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/BaseSentinelDubboFilter.java
```diff
@@ -35,47 +38,69 @@ public BaseSentinelDubboFilter() {
         this.listener = new SentinelDubboListener();
     }
 
-    static class SentinelDubboListener implements Listener {
+
```
cdfive, line 41: It's better to keep only one blank line here.
Also Line#78,Line#80.
Line#104 can be removed for better and uniform formatting.

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java
cdfive: Sorry for wrong idea, fixed key is incorrect here, since a method may have more than one consumer call with sync or async. I'm agree with adding the Sentinel prefix for identification.

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java
cdfive: +1, when `methodEntry` is blocked, the `rpcContext#set` cann't be executed, and the value in `rpcContext` with key of `methodResourceName` is null. We may need to keep the same as the previous commit, separate with two different key.

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboConsumerFilter.java
```diff
@@ -52,33 +55,87 @@ public SentinelDubboConsumerFilter() {
...
+                interfaceEntry.exit();
             }
-            rpcContext.set(DubboUtils.DUBBO_METHOD_ENTRY_KEY, methodEntry);
-            return invoker.invoke(invocation);
+        }
+    }
+
+
+    private Result asyncInvoke(Invoker<?> invoker, Invocation invocation) {
+        Stack<Entry> entryStack = new Stack<>();
```
sczyh30, line 111: How about using `Deque` instead to avoid bad performance? `Stack` is derived from `Vector`, whose operations are wrapped with `synchronized`.

jasonjoo2010: Oh why not using `final` here? Any performance related reason?

Lin-Liang: final can improve the performance highly?

jasonjoo2010: > final can improve the performance highly?

Not really.
Talk to performance in time efficiency i should say there isn't obvious difference. It's a suggestion aiming:

* No extra data structure was introduced.
* Look simpler
* Implicit reference was introduced for `Stack` or `Dequeue` compared to `final`. They are similar in this view.

sczyh30:
Nice work. Thanks!

PR #1437, opened 2020-04-26
Update resolving logic of project name and polish SentinelConfig

sczyh30:

Update resolving logic of project name and polish SentinelConfig

Polish for #1191 

### Describe how you did it

- Update resolving logic of the project name. Now the order of resolving the name is: `system env (CSP_SENTINEL_APP_NAME) > property (csp.sentinel.app.name) > property (project.name) > main class (or jar) name`
- Polish related code in `SentinelConfig`

### Describe how to verify it

Run the demo and add relevant attributes.

### Special notes for reviews

NONE

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfig.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfig.java
@@ -42,9 +42,13 @@ public final class SentinelConfig {
     public static final int APP_TYPE_COMMON = 0;
 
     private static final Map<String, String> props = new ConcurrentHashMap<>();
-    private static int appType = APP_TYPE_COMMON;
 
-    public static final String APP_TYPE = "csp.sentinel.app.type";
+    private static int appType = APP_TYPE_COMMON;
+    private static String appName = "";
+
+    public static final String PROJECT_NAME_PROP_KEY = "project.name";
+    public static final String APP_NAME_PROP_KEY = "csp.sentinel.app.name";
+    public static final String APP_TYPE_PROP_KEY = "csp.sentinel.app.type";
     public static final String CHARSET = "csp.sentinel.charset";
     public static final String SINGLE_METRIC_FILE_SIZE = "csp.sentinel.metric.file.single.size";
     public static final String TOTAL_METRIC_FILE_COUNT = "csp.sentinel.metric.file.total.count";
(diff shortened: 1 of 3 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfig.java
```diff
@@ -42,37 +42,33 @@
...
-    public static final String SUN_JAVA_COMMAND = "sun.java.command";
-    private static final String JAR_SUFFIX_LOWER = ".jar";
-    private static final String JAR_SUFFIX_UPPER = ".JAR";
-
-    public static final int DEFAULT_STATISTIC_MAX_RT = 4900;
+    public static final String DEFAULT_CHARSET = "UTF-8";
+    public static final long DEFAULT_SINGLE_METRIC_FILE_SIZE = 1024 * 1024 * 50;
+    public static final int DEFAULT_TOTAL_METRIC_FILE_COUNT = 6;
+    public static final int DEFAULT_COLD_FACTOR = 3;
+    public static final int DEFAULT_STATISTIC_MAX_RT = 5000;
```
cdfive, line 63: Hhh, so now the max RT is 5000, and not the classic value 4900?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfig.java
```diff
@@ -242,32 +237,54 @@ public static int statisticMaxRt() {
...
-
-        if (props.get(APP_NAME) != null) {
-            appName = props.get(APP_NAME);
+    private static void resolveAppName() {
+        // Priority: system env -> csp.sentinel.app.name -> project.name -> main class (or jar) name
+        String envKey = toEnvKey(APP_NAME_PROP_KEY);
+        String n = System.getenv(envKey);
+        if (!StringUtil.isBlank(n)) {
+            appName = n;
+            RecordLog.info("App name resolved from system env {}: {}", envKey, appName);
```
cdfive, line 246: The log info doesn't has prefix `[SentinelConfig]` like Line#71, and Line#71 move to inside `resolveAppType` method looks better. It's okay I think and we can do optimize them in the future.

sczyh30: > The log info doesn't has prefix `[SentinelConfig]` like Line#71

Actually prefix is not strictly required if the log content is concrete enough :)

> Line#71 move to inside `resolveAppType` method looks better. It's okay I think and we can do optimize them in the future.

Good idea. We may improve this later.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfig.java
sczyh30: The magic 4900 often makes users confusing, so I've changed it to 5000. The attribute might be deprecated in future versions.

PR #1443, opened 2020-04-29
Fix the bug that cookie of the dashboard may have conflict with web applications under the same domain

zhaoyuguang:
resolve https://github.com/alibaba/Sentinel/issues/1434

### DIFF
--- a/sentinel-dashboard/src/main/resources/application.properties
+++ b/sentinel-dashboard/src/main/resources/application.properties
@@ -3,6 +3,9 @@ spring.http.encoding.force=true
 spring.http.encoding.charset=UTF-8
 spring.http.encoding.enabled=true
 
+#cookie name setting
+server.servlet.session.cookie.name=${SENTINEL_COOKIE_NAME:sentinel}
+
 #logging settings
 logging.level.org.springframework.web=INFO
 logging.file=${user.home}/logs/csp/sentinel-dashboard.log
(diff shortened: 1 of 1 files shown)


sentinel-dashboard/src/main/resources/application.properties
```diff
@@ -3,6 +3,9 @@ spring.http.encoding.force=true
 spring.http.encoding.charset=UTF-8
 spring.http.encoding.enabled=true
 
+#cookie name setting
+server.servlet.session.cookie.name=${SENTINEL_COOKIE_NAME:sentinel}
+
```
cdfive, line 8: Using placeholders, the default value is sentinel, and we can change it 
by VM argument `-DSENTINEL_COOKIE_NAME=xxx`. 
Now the argument name `SENTINEL_COOKIE_NAME` is uppercase, since others such as`auth.username` and `auth.password`.. are lower case with dot separate, they are not unified.

It seems that only use `sentinel` or `sentinel_cookie`、`sentinel_dashboard_cookie` is enough, since we can change it by `-Dserver.servlet.session.cookie.name=xxx`.

cdfive: Above are some suggestions, sorry for choosing dyslexia myself, and I think ${SENTINEL_COOKIE_NAME:sentinel} is also acceptable.
How about your opinion? @zhaoyuguang @sczyh30

zhaoyuguang: 仔细想下 `${SENTINEL_COOKIE_NAME:`的确是多余的，把默认的cookie key改成了`sentinel_dashboard_cookie`这个看起来比较清晰 @cdfive

sczyh30:
Thanks for contributing!

PR #1454, opened 2020-05-04
Update method name of AbstractSentinelInterceptor in Spring Web adapter

researchlive:
The old name is a little bit unclear. So in fact in this method, it is tracking from the entry to exit in order to catch the error messages. But the old name makes people may think that this method is tracking the exception happening process. Just want to make it clear


sczyh30:
Could you please resolve the CI failure?

sczyh30:
@researchlive any feedback?

sczyh30:
Closing this PR due to no feedback. Anyway thanks for contributing.

Issue #1032, opened 2019-09-03
Wrong Circuit Breaker  Implementation


haoel:
After check the documentation and quickly read through the source code, I think I need raise this issue seriously.

- Only has Open and Close status, no Half-Open Status, really broken implementation. Please follow [Martin Flow's design document](https://martinfowler.com/bliki/CircuitBreaker.html) to do the implementation. otherwise, do not named circuit breaker, this could mislead people.

- Using the response time to check the health of upstream is good, but using the average is bad. Please read the article "[Why Averages Suck and Percentiles are Great](https://www.dynatrace.com/news/blog/why-averages-suck-and-percentiles-are-great/) "

- There are three types, Avge RT, Exception  Ratio, & Exception Count, the first two types are "per second", the third one is "per minutes", this is the inconsistent design. and if the throughput is low, it could make the wrong decision.

--Hao

Issue #308, opened 2018-12-17
[RFC] Changing statistic time window in degrade rule dynamically


CarpenterLee:
## Issue Description

Type: *feature request*

### Describe what feature you want

Now Sentinel uses sliding window algorithm to statistic metrics, and time window is globally the same, which is usually 1 or 2 seconds. This time window fits well for the flow controlling and shaping, because flow controlling primary based on QPS, a short time window means more sensitive.

But as for degradation(Circuit Breaking), multi statistic time interval is needed, since degradation is based on average RT, and exception count or exception ratio, 1 or 2 seconds may be too short in low traffic.

### Describe what you expected to happen

Changing time window in degrade rule dynamically.

The main challenge to set time window interval dynamically for each degrade rule is that flow rule and degrade rule use the same statistic node, changing the interval will lead interaction effect.

One solution is to use different statistic node for flow rule and degrade rule. Current statistic structure is use for flow controlling only, and creating new statistic node for degrade rule, every time the time interval is changed, statistic node will be recreated. The statistic nodes for degrade rule are created lazily.

Frankzhaopku:
This issue has seen no activity in 15 days, it has been automatically closed for house-keeping purpose.

westonlv:
现在降级是支持的啊，DegradeRule.setTimeWindow可以设置。
流控本身就针对qps的，没必要了吧

CalvinKirs:
Let me finish it, so the discussion on this topic (https://github.com/alibaba/Sentinel/issues/759) is unnecessary. @CarpenterLee

CalvinKirs:
Because of the statistics, recounting is unavoidable, but we can introduce a new variable and measure ms as the base unit. For example, the time window of the Circuit is 5ms, then this new variable is used to calculate how many times have passed. 1ms @CarpenterLee

Issue #56, opened 2018-08-10
Separate rule action from rule definition


sczyh30:
## Issue Description

Type: *enhancement*

### Describe what feature you want

Separate rule action (`passCheck`) from rule definition (`Rule` entity).

- [x] FlowRule
- [x] DegradeRule (a temporary solution is to abstract a `CircuitBreaker`)
- [x] AuthorityRule
- [x] ParamFlowRule

sczyh30:
Finally resolved in #1490 after two years... Cheers! 🍺

jasonjoo2010:
> Finally resolved in #1490 after two years... Cheers! 🍺

😂

PR #1490, opened 2020-05-18
Refactor degrade hierarchy with new circuit breaker mechanism and improve strategy

### LINKED ISSUE #1421: Add circuit breaker abstraction and separate it from rule definition
## Issue Description

Type: *feature request*

Add circuit breaker abstraction and separate checking logic (i.e. CircuitBreaker) from legacy `DegradeRule` definition.

Also related to #56

sczyh30:

Refactor legacy degrade hierarchy with new circuit breaker mechanism and improve strategy.

Resolves #1421, #1032, #951, #154, #308, #56

### Describe how you did it

* Add `CircuitBreaker` abstraction (with half-open state) and add circuit breaker state change event observer support. Now Sentinel follows the canonical circuit breaker pattern (with some improvements).
* Make statistics of each rule dependent (to support **arbitrary statistic interval**).
* Improve RTT-based circuit breaking strategy (avg RT → slow request ratio).
* Add simple "probe" mechanism (aka. half-open).
* Refactor mechanism of metric recording and state change handling for circuit breakers: record RT and error when requests have completed (i.e. `onExit`, based on #1420).

### Describe how to verify it

Run the test cases and demo.

### Special notes for reviews

This PR contains internal breaking changes (and some behavioral changes for RT-based circuit breaking).


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRuleManager.java
```diff
@@ -146,88 +142,128 @@ public static boolean setRulesForResource(String resourceName, Set<DegradeRule>
...
-                Set<DegradeRule> ruleSet = newRuleMap.get(identity);
-                if (ruleSet == null) {
-                    ruleSet = new HashSet<>();
-                    newRuleMap.put(identity, ruleSet);
+                CircuitBreaker cb = getExistingSameCbOrNew(rule);
+                if (cb == null) {
+                    RecordLog.warn("[DegradeRuleManager] Unknown circuit breaking strategy, ignoring: " + rule);
+                    continue;
                 }
-                ruleSet.add(rule);
```
cdfive, line 201: Inside `getExistingSameCbOrNew` method, the design of reuse the circuit breaker if the rule remains unchanged.is smart. I tried with `DegradeRuleManager.loadRules` in different places, it works. A small question is that, if using `DegradeRuleManager.loadRules(rules)` first time, and rules have two same `DegradeRule` , there will be two `CircuitBreaker` created, since in the first time, the static `circuitBreakers` variable is null.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeSlot.java
```diff
@@ -15,30 +15,73 @@
...
+            fireExit(context, r, count, args);
+            return;
+        }
+        List<CircuitBreaker> circuitBreakers = DegradeRuleManager.getCircuitBreakers(r.getName());
+        if (circuitBreakers == null || circuitBreakers.isEmpty()) {
+            fireExit(context, r, count, args);
+            return;
+        }
+
+        if (curEntry.getBlockError() == null) {
```
cdfive, line 73: This judgement seems can be removed since Line#63.

CodingSinger:
where is the  proposal for this pr?

sczyh30:
> where is the proposal for this pr?

I'll link the proposal here soon.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
```diff
@@ -0,0 +1,147 @@
...
+        // Template implementation.
+        if (currentState.get() == State.CLOSED) {
+            return true;
+        }
+        if (currentState.get() == State.OPEN) {
+            // For half-open state we allow a request for trial.
+            return retryTimeoutArrived() && fromOpenToHalfOpen();
+        }
+        return false;
+    }
```
wavesZh, line 74: If thare are two breakers and curState is open, one breaker tryPass and change state to half open, but other breaker tryPass return false, it is right that no request allow. but in below requqests will no allow pass because two breakers  can't both be true.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
```diff
@@ -0,0 +1,147 @@
...
+import com.alibaba.csp.sentinel.util.TimeUtil;
+
+/**
+ * @author Eric Zhao
+ * @since 1.8.0
+ */
+public abstract class AbstractCircuitBreaker implements CircuitBreaker {
+
+    protected final DegradeRule rule;
+    protected final int retryTimeoutMs;
```
wavesZh, line 32: I think "attempt/try" is better than "retry"

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRuleManager.java
sczyh30: Yes, it can be improved.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeSlot.java
sczyh30: It's here in case the blockError was modified.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
```diff
@@ -52,47 +45,45 @@
...
-    private int grade = RuleConstant.DEGRADE_GRADE_RT;
+    private int timeWindow;
 
     /**
-     * Minimum number of consecutive slow requests that can trigger RT circuit breaking.
+     * Minimum number of requests (in an active statistic time span) that can trigger circuit breaking.
      *
      * @since 1.7.0
      */
-    private int rtSlowRequestAmount = RuleConstant.DEGRADE_DEFAULT_SLOW_REQUEST_AMOUNT;
```
jasonjoo2010, line 88: Here is a missing property in new implementation. Is that acceptable in compatibility?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeSlot.java
```diff
@@ -15,30 +15,73 @@
...
+            if (!cb.tryPass()) {
+                throw new DegradeException(cb.getRule().getLimitApp(), cb.getRule());
+            }
+        }
+    }
+
     @Override
-    public void exit(Context context, ResourceWrapper resourceWrapper, int count, Object... args) {
-        fireExit(context, resourceWrapper, count, args);
+    public void exit(Context context, ResourceWrapper r, int count, Object... args) {
```
jasonjoo2010, line 61: And here's a problem that if something went wrong which may cause RT down to more than 10s (100ms for normal) . Our breaker will cut the traffic after 10s. Maybe we could improve it (or as a new issue and do it in future) that unfinished entries could be recorded as `statIntervalMs` when slidingCounter is reset.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
sczyh30: Maybe "recoveryTimeout"?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/CircuitBreakerStateChangeObserver.java
```diff
@@ -0,0 +1,50 @@
...
+    void onTransformToClosed(CircuitBreaker.State prev, DegradeRule rule);
+
+    /**
+     * Observer method triggered when circuit breaker state transformed to {@code OPEN}.
+     *
+     * @param prev          previous state of the circuit breaker
+     * @param rule          associated rule
+     * @param snapshotValue triggered value on circuit breaker opens
+     */
+    void onTransformToOpen(CircuitBreaker.State prev, DegradeRule rule, double snapshotValue);
```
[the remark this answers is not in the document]
sczyh30: Good idea.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/CircuitBreaker.java
```diff
@@ -0,0 +1,72 @@
...
+    void onRequestComplete(long rt, Throwable error);
+
+    /**
+     * Circuit breaker state.
+     */
+    enum State {
+        /**
+         * In {@code OPEN} state, all requests will be rejected until the next retry time point.
+         */
+        OPEN,
```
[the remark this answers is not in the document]
sczyh30: `OPEN`, `CLOSED` and `HALF-OPEN` are terminologies of [circuit breaker pattern](http://martinfowler.com/bliki/CircuitBreaker.html), so maybe keeping it is better.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -0,0 +1,46 @@
...
+package com.alibaba.csp.sentinel.slots.block.degrade.circuitbreaker;
+
+/**
+ * @author Eric Zhao
+ * @since 1.8.0
+ */
+public enum CircuitBreakerStrategy {
+
+    /**
+     * Circuit breaker opens when slow request ratio exceeds the threshold.
```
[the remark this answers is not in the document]
sczyh30: I'll add additional descriptions to avoid ambiguity.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeSlot.java
sczyh30: Yes, it's necessary to improve the scenarios regarding slow in-flight requests. See #1405

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
sczyh30: It could bring breaking changes (and break the SEMVER spec) indeed. In convention, we should mark it deprecated and remove it until 2.x. But actually it's rarely used and often regarded as a "hidden" attribute (not even appeared in the dashboard)...

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeSlot.java
jasonjoo2010: Oh they are not the same issue i think.

How to calculate a more reasonable RT is one thing while how to make a more reasonable degradation is quite another thing. The primary reason is that the calculating window is separated from general RT, isn't?

And sure we can improve it later but it's more important than making it more reasonable calculating RT. Maybe another time window is required to make it under tracing.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
jasonjoo2010: Yeah you're right. But i just afraid the users who integrating manually suffering broken compatibility.😂 We can make decision here whether doing it carefully this time. Decide and record it and it would be ok.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/CircuitBreaker.java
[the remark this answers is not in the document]
jasonjoo2010: Totally agree.

More descriptions would be more friendly for beginners.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRuleManager.java
```diff
@@ -98,12 +87,19 @@ public static boolean hasConfig(String resource) {
      */
     public static List<DegradeRule> getRules() {
         List<DegradeRule> rules = new ArrayList<>();
-        for (Map.Entry<String, Set<DegradeRule>> entry : degradeRules.entrySet()) {
-            rules.addAll(entry.getValue());
+        for (Map.Entry<String, List<CircuitBreaker>> entry : circuitBreakers.entrySet()) {
+            for (CircuitBreaker cb : entry.getValue()) {
```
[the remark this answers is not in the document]
sczyh30: I'll fix it.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRuleManager.java
```diff
@@ -146,88 +142,128 @@ public static boolean setRulesForResource(String resourceName, Set<DegradeRule>
...
     }
 
     private static class RulePropertyListener implements PropertyListener<List<DegradeRule>> {
 
+        private synchronized void reloadFrom(List<DegradeRule> list) {
+            Map<String, List<CircuitBreaker>> cbs = buildCircuitBreakers(list);
+            Map<String, Set<DegradeRule>> rm = new HashMap<>(cbs.size());
+
+            for (Map.Entry<String, List<CircuitBreaker>> e : cbs.entrySet()) {
+                if (e.getValue() == null || e.getValue().isEmpty()) {
```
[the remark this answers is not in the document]
sczyh30: Changed to assertion in case of bugs.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeRule.java
sczyh30: IMHO we could drop it (the legacy bad design) and add notes in relevant documents :)

sczyh30:
@jasonjoo2010 I've updated the PR, please check.

Note: please use "rebase and merge" for this PR instead of "squash and merge".

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
wavesZh: @sczyh30  Any idea for the scene？Was I wrong？

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
wavesZh: “nextRetryTimestamp” is same above .

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
jasonjoo2010: I think it's ok for this scenario.

B1(open) & B2(open)   -> no request at all
B1(open -> half) & B2(open)   -> no request at all, but B1 may transform to closed leaving B2 alone, that is:
B1(closed) & B2(open)

wavesZh: Sorry, i cannot find how to change half-open to close.
Method `fromHalfOpenToClose` is invoked in `DegradeSlot#exist` when no BlockingException. But if B1(open -> half) & B2(open), the follow requests check by `CircuitBreaker#tryPass` must return false then throw DegradeException.

jasonjoo2010: Requests is blocked by `tryPass()` in abstract parent and the state of breakers are maintained in different implementations when requests complete (post-request).

So they are not collided.

wavesZh: 可能我说的不清楚。
假如此时 B1(open) & B2(open)，B1,B2的 recoveryTimeoutMs 不一样，下一个请求导致B1(half)，而B2由于还未达到 timeout 时间，仍保持 B2(open)，故 tryPass 不通过进而抛出 BlockException. 在 DegradeSlot#exit 中仅对非 BlockException 的情况维护 Breaker 状态，故此时 B1(half) & B2(open).

```java
public void exit(Context context, ResourceWrapper r, int count, Object... args) {
        Entry curEntry = context.getCurEntry();
        if (curEntry.getBlockError() != null) {
            fireExit(context, r, count, args);
            return;
        }
        ...
        if (curEntry.getBlockError() == null) {
          ...
          for (CircuitBreaker circuitBreaker : circuitBreakers) {
                circuitBreaker.onRequestComplete(rt, error);
            }
        }
       ...
}
```java

后面的请求，由于 B1 不符合 [open, close] 状态，tryPass 必然为 false 又抛 BlockException，进入循环。

> B1(open -> half) & B2(open) -> no request at all, but B1 may transform to closed leaving B2 alone, that is:
B1(closed) & B2(open)

这个我没找到在什么哪里有个逻辑实现。

Comment (jasonjoo2010) [reply sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java:74]: 噢我懂你关注的点了，不过这个场景中的处理有问题吗？
B1/B2同时对资源做处理（例如一个针对RT一个针对ErrorRate），这时一个关断一个从关断进入半恢复状态，那么单个的状态是不是有及时恢复，理论上没这么重要，因为其中之一关断则资源关断，两者较长恢复时间者进入半开则两者恢复通路，对于不同管控指标同一资源的断路器这里貌似不存在问题。是这样吗？

Comment (wavesZh) [reply sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java:74]: 不对吧。
> 后面的请求，由于 B1 不符合 [open, close] 状态，tryPass 必然为 false 又抛 BlockException，进入循环

就算 B2 可以进入半开，但是 B1(half) 也会因为上面的判断导致抛出 BlockException，后续在 exit 方法中也不会尝试更改断路器的状态了

Comment (jasonjoo2010) [reply sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java:74]: 在tryPass中，

open -> false
closed -> true
half -> retryTimeout && halfToOpen

这里并不是说非open非close一定false的，是我理解错了吗？

Comment (wavesZh) [reply sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java:74]:
```java
public boolean tryPass() {
        // Template implementation.
        if (currentState.get() == State.CLOSED) {
            return true;
        }
        if (currentState.get() == State.OPEN) {
            // For half-open state we allow a request for trial.
            return retryTimeoutArrived() && fromOpenToHalfOpen();
        }
        return false;
    }
```
我已经自我怀疑了🤨😢  是我理解错了吗？？

jasonjoo2010: > ```java
> ```java
> public boolean tryPass() {
>         // Template implementation.
>         if (currentState.get() == State.CLOSED) {
>             return true;
>         }
>         if (currentState.get() == State.OPEN) {
>             // For half-open state we allow a request for trial.
>             return retryTimeoutArrived() && fromOpenToHalfOpen();
>         }
>         return false;
>     }
> ```
> 
> 
> 我已经自我怀疑了🤨😢  是我理解错了吗？？
> ```

emmm.........仔细想了下，你考虑的问题的确存在，而且貌似还存在其它限流规则的问题，因为这里并没有判断是不是DegradeException，待我下午拿Test测一下

jasonjoo2010: @sczyh30 Hi, Eric

Here're some boundaries which leads to new issues while we should discuss more how to make the recovery logic better:

1. Multiple degrading rules for same resource(Especially after introducing prefix/postfix matching)
2. Another rule after degrading rule in half-open state(Though there's no such kind of rule but it's possible in future)

These will lead to the incorrect workflow:

entry:   R0(open -> half) -> R1(block)
exit: XXXBlockException -> skip the detection for stalling half state

For straightforward patch i think the `onComplete` method is intentioned to be call after a successfully transforming into half. We should make the exiting logic more simple and clear.

sczyh30: @jasonjoo2010 @wavesZh Yes, it's a fatal bug indeed and should be resolved immediately. However we need a better temporary workaround for this (for the half-open case). I'll merge this PR first and we may discuss it in a new issue. It's a common problem for all rules that have their own metrics and rely on `onComplete`.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/DegradeSlot.java
sczyh30: We could open a new issue to discuss it.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
sczyh30: See #1638

PR #1491, opened 2020-05-20
support param flow for complex param

CodingSinger:
In some scenarios, we may need to limit the flow of a field of a complex object, but the current version does not support it. Therefore, we provide an interface that allows to construct and return a value as the limit object.


sczyh30:
Thanks for contributing!

PR #1496, opened 2020-05-21
provide a predicate to check whether the throwable should be traced

pleasecheckhere2016:
提供了一种可以通过predicate判断该异常是否应该被trace的方法。

### Describe how you did it

```java
    @Test
    public void setExceptionPredicate() {
        Predicate<Throwable> throwablePredicate = new Predicate<Throwable>() {
            @Override
            public boolean test(Throwable throwable) {
                if (throwable instanceof TraceException) {
                    return true;
                } else if (throwable instanceof IgnoreException) {
                    return false;
                }
                return false;
            }
        };
        Tracer.setExceptionPredicate(throwablePredicate);
        Assert.assertTrue(Tracer.shouldTrace(new TraceException()));
        Assert.assertFalse(Tracer.shouldTrace(new IgnoreException()));
    }
```


sczyh30:
Good design. Thanks for contributing and look forward to more!

PR #1532, opened 2020-06-08
Dubbo adapter: entry exit with param

wavesZh:

If entry with param, then exit should be better to take the param.


sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboConsumerFilter.java
```diff
@@ -98,7 +98,7 @@ private Result syncInvoke(Invoker<?> invoker, Invocation invocation) {
             throw e;
         } finally {
             if (methodEntry != null) {
-                methodEntry.exit();
+                methodEntry.exit(1, invocation.getArguments());
```
cdfive, line 101: Very careful code! How about also update `SentinelDubboConsumerFilter` in `sentinel-dubbo-adaptor` module to keep them in line, though they are for different version of dubbo.

wavesZh: ConsumerFilter in `sentinel-dubbo-adaptor` don't take param when entry. Should I add param when entry and exit?

cdfive: Yes, it does't take param when entry currently, users may feel confused and suggest them to be consistent.

sczyh30:
Thanks! We may evaluate the memory footprint later, as this may yield a lot of `EntryHolder`.

wavesZh:
> Thanks! We may evaluate the memory footprint later, as this may yield a lot of `EntryHolder`.

I ignore that.. But If it is need to think about,  `LinkedList<EntryHolder> queue` maybe be not required too?

PR #1544, opened 2020-06-13
[bugfix]FLowRule class equals() function is wrong

yunfeiyanggzq:
Signed-off-by: yunfeiyanggzq <yunfeiyang@buaa.edu.cn>

The previous code forgot to compare `controller`,so even the controller is different,the ruleFlow is equal

### Special notes for reviews
I am an asoc participant. I will be mainly responsible for the cluster flow control. I will supplement the missing unit tests while reading the relative code.Thanks for your review and help!!!


yunfeiyanggzq:
after communicating with  sczyh30,it is not a bug ,so I close this pr

Issue #1252, opened 2020-01-14
When blocked in Dubbo adapter, Sentinel should wrap SentinelRpcException with Result rather than directly throw it out


lesline:
## when block exception happen, sentinel return RpcException exception，not SentinelRpcException exception

Type: bug report

### Describe problem:
problem code
```
com.alibaba.csp.sentinel.adapter.dubbo.fallback.DefaultDubboFallback
public class DefaultDubboFallback implements DubboFallback {
    @Override
    public Result handle(Invoker<?> invoker, Invocation invocation, BlockException ex) {
        // Just wrap and throw the exception.
        throw new SentinelRpcException(ex);
    }
}
```
should change to :

```
public class DefaultDubboFallback implements DubboFallback {

    @Override
    public Result handle(Invoker<?> invoker, Invocation invocation, BlockException ex) {
        // Just wrap and throw the exception.
        RpcResult result = new RpcResult();
        result.setException(new SentinelRpcException(ex));
        return result;
    }
}
```

### Tell us your environment
    <version>1.7.2-SNAPSHOT</version>

sczyh30:
Good idea. Would you like to contribute a PR to improve it?

lesline:
ok

lesline:
ok

sczyh30:
PRs are welcomed for Apache Dubbo 2.7.x adapter!

cdfive:
Maybe relate to #374, #1065, #1017, when using `DefaultDubboFallback`, we need to be careful about `Dubbo ExceptionFilter`, `Serialize/Deserialize` and `JDK vension`.

wzg923:
#1065

sczyh30:
Already resolved in 1.8.x.

PR #1572, opened 2020-06-29
Polish Dubbo 2.6.x adapter

sczyh30:

Polish Dubbo 2.6.x adapter.

Related to #1252 

### Describe how you did it

- Unify Dubbo callback registry (for fallback and origin parser) into `DubboAdapterGlobalConfig`
- Polish default fallback implementation (wrap exception with `RpcResult` rather than directly throw it out)

### Describe how to verify it

Run the test cases and demo.

### Special notes for reviews

This PR contains some breaking changes. Please pay attention.


sentinel-adapter/sentinel-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java
```diff
@@ -56,19 +53,20 @@ public SentinelDubboProviderFilter() {
...
-            String resourceName = getResourceName(invoker, invocation, DubboConfig.getDubboProviderPrefix());
+            String methodResourceName = getMethodResourceName(invoker, invocation,
+                DubboAdapterGlobalConfig.getDubboProviderPrefix());
             String interfaceName = invoker.getInterface().getName();
-            ContextUtil.enter(resourceName, origin);
+            ContextUtil.enter(methodResourceName, origin);
             interfaceEntry = SphU.entry(interfaceName, ResourceTypeConstants.COMMON_RPC, EntryType.IN);
-            methodEntry = SphU.entry(resourceName, ResourceTypeConstants.COMMON_RPC,
+            methodEntry = SphU.entry(methodResourceName, ResourceTypeConstants.COMMON_RPC,
                 EntryType.IN, invocation.getArguments());
```
cdfive, line 70: If provider prefix is configured, the methodResourceName has a prefix which is marked for provider, it's useful when searching with prefix in dashboard menu(one scene is that only want to view all provider resources), but the interfaceResourceName has no prefix. How about also adding prefix for interface resource? 
Maybe `getInterfaceResourceName` can be added in `AbstractDubboFilter`.Same  for consumer Filter.

Besides, is there a need to distinguish between interface and method via another prefix, a little tangled. 
@sczyh30 @jasonjoo2010

sczyh30: Yes, the interface resource name may be appended by the same prefix (if set).

sentinel-adapter/sentinel-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/AbstractDubboFilter.java
```diff
@@ -43,15 +43,24 @@ protected String getResourceName(Invoker<?> invoker, Invocation invocation) {
...
 
+    protected String getInterfaceName(Invoker<?> invoker) {
+        return invoker.getInterface().getName();
+    }
 
+    protected String getInterfaceName(Invoker<?> invoker, String prefix) {
+        if (StringUtil.isBlank(prefix)) {
+            return getInterfaceName(invoker);
+        }
+        return prefix + getInterfaceName(invoker);
```
cdfive, line 64: It seems that normally when overloading methods, the method which has less parameters call the method which has more. This is the opposite, but it is okay.

cdfive: Besides, I found that dubbo-adapter of 2.7.x has a `useGroupAndVersion` setting.
`String getMethodResourceName(Invoker<?> invoker, Invocation invocation, Boolean useGroupAndVersion)`.
`DubboConfig.getDubboInterfaceGroupAndVersionEnabled(), csp.sentinel.dubbo.interface.group.version.enabled`
Do we need to add this in 2.6.x?

Another setting is `csp.sentinel.dubbo.trace.biz.exception.enabled` in 2.7.x, and it is unused.

sczyh30: > Besides, I found that dubbo-adapter of 2.7.x has a `useGroupAndVersion` setting.
> `String getMethodResourceName(Invoker<?> invoker, Invocation invocation, Boolean useGroupAndVersion)`.
> `DubboConfig.getDubboInterfaceGroupAndVersionEnabled(), csp.sentinel.dubbo.interface.group.version.enabled`
> Do we need to add this in 2.6.x?

Yes, we could add this in the future. We may open another issue to address this.

> Another setting is `csp.sentinel.dubbo.trace.biz.exception.enabled` in 2.7.x, and it is unused.

Actually I don't think it's useful.

cdfive: OK, I agree with you. IMHO the prefix is more useful, maybe enough at present.

sczyh30: > OK, I agree with you. IMHO the prefix is more useful, maybe enough at present.

The group and version info can be useful. This could be contributed by community later.

cdfive: Thanks for detail description, group and version are useful indeed, got it : )

PR #1581, opened 2020-07-02
update the slot spi demo

yunfeiyanggzq:
update the slot spi demo

### Describe how to verify it
the output is as following
```
INFO: Sentinel log output type is: file
INFO: Sentinel log charset is: utf-8
INFO: Sentinel log base directory is: C:\Users\m1881\logs\csp\
INFO: Sentinel log name use pid is: false
Current context: sentinel_default_context
Current entry resource: abc
Exiting for entry on DemoSlot: abc

Process finished with exit code 0

```

### DIFF
--- a/sentinel-demo/sentinel-demo-slot-chain-spi/src/main/java/com/alibaba/csp/sentinel/demo/slot/DemoSlotChainBuilder.java
+++ b/sentinel-demo/sentinel-demo-slot-chain-spi/src/main/java/com/alibaba/csp/sentinel/demo/slot/DemoSlotChainBuilder.java
@@ -20,17 +20,11 @@ import com.alibaba.csp.sentinel.slotchain.SlotChainBuilder;
 import com.alibaba.csp.sentinel.slots.DefaultSlotChainBuilder;
 
 /**
- * An example slot chain builder. To activate this slot chain builder,
- * add the class name to corresponding SPI file in `resource/META-INF/services` directory.
- *
  * @author Eric Zhao
+ *
+ * @deprecated since 1.7.2, we can use @SpiOrder(-3500) to adjust the order of {@link DemoSlot},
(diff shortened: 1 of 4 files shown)


yunfeiyanggzq:
please review, @sczyh30.Thanks!

sentinel-demo/sentinel-demo-slot-chain-spi/src/main/java/com/alibaba/csp/sentinel/demo/slot/DemoSlotChainBuilder.java
```diff
@@ -20,17 +20,11 @@
 import com.alibaba.csp.sentinel.slots.DefaultSlotChainBuilder;
 
 /**
- * An example slot chain builder. To activate this slot chain builder,
- * add the class name to corresponding SPI file in `resource/META-INF/services` directory.
- *
  * @author Eric Zhao
+ *
+ * @deprecated since 1.7.2, we can use @SpiOrder(-3500) to adjust the order of {@link DemoSlot},
+ * this class is reserved for compatibility with older versions.
```
cdfive, line 26: If has no annotation, the order is `SpiOrder.LOWEST_PRECEDENCE` and the slot will be at the end of slot chain.
How about adding `@SpiOrder(-3500)` in `DemoSlot` class?  
As the comment mentioned, the `DemoSlot` will be after `SystemSlot` and before `FlowSlot`,  and the demo is more friendly for new users.

yunfeiyanggzq: good idea,done

sczyh30:
Thanks for contributing!

PR #1588, opened 2020-07-06
需要增加请求charset非空判断，避免charset.name()空指针异常#1403 2

JiangZian:
需要增加请求charset非空判断，避免charset.name()空指针异常#1403 2

 just carry the message in IllegalArgumentException.


JiangZian:
close

Issue #1403, opened 2020-04-14
Some potential NPE bugs


JulyChen728:
Hi all, our bug scanner has reported some NPE bugs.

1.The first one is at [FileInJarReadableDataSource.java#L99](https://github.com/alibaba/Sentinel/blob/master/sentinel-extension/sentinel-datasource-extension/src/main/java/com/alibaba/csp/sentinel/datasource/FileInJarReadableDataSource.java#L99) since the variable `inputStream` may be `null`.

2.The second one is at [SimpleHttpClient.java#L168](https://github.com/alibaba/Sentinel/tree/master/sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/SimpleHttpClient.java#L168) cause the variable `charset` may be `null`. Since the variable `charset` is checked for `null` at [SimpleHttpClient.java#L100](https://github.com/alibaba/Sentinel/tree/master/sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/SimpleHttpClient.java#L100), a NPE bug may take place when the call chain is [SimpleHttpClient.java#L111](https://github.com/alibaba/Sentinel/tree/master/sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/SimpleHttpClient.java#L111) => [SimpleHttpClient.java#L168](https://github.com/alibaba/Sentinel/tree/master/sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/SimpleHttpClient.java#L168).

[...]

PR #1589, opened 2020-07-06
The request charset non-null judgment needs to be added to avoid the charset.name() NPE #1403 2

JiangZian:
The request charset non-null judgment needs to be added to avoid the charset.name() NPE #1403 2


sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/SimpleHttpClient.java
```diff
@@ -156,6 +156,9 @@ private String getStatusLine(RequestMethod type, String requestPath) {
      * @return encoded request parameters, or empty string ("") if no parameters are provided
      */
     private String encodeRequestParams(Map<String, String> paramsMap, Charset charset) {
+        if (charset == null){
+            throw new IllegalArgumentException( "charset is not allowed to be null");
+        }
```
sczyh30, line 161: Could you please reformat your code?

JiangZian: 好的

sczyh30:
Thanks for contributing!

Issue #1109, opened 2019-10-23
getServletPath() always returns empty when matching URL in Zuul API gateway flow control


fusc2018:
Issue Description: context.getRequest().getServletPath() always empty

type: bug

Class :PrefixRoutePathMatcher.class

Method: test

Content:

```java
    @Override
    public boolean test(RequestContext context) {
        String path = context.getRequest().getServletPath();
        if (canMatch) {
            return pathMatcher.match(pattern, path);
        }
        return false;
    }
```

### How to reproduce it (as minimally and precisely as possible)

```java
  @Override
    public boolean test(RequestContext context) {
        String path = context.getRequest().getRequestURI();
        if (canMatch) {
            return pathMatcher.match(pattern, path);
        }
        return false;
    }
```

sczyh30:
Could you please give more information (e.g. the path you're visiting, and maybe your Zuul Servlet config)?

fusc2018:
First of all, the zuul version I used is 2.1.1.
ZuulServletConfig configuration is as follows

zuul:
  servlet-path: /
  ignore-security-headers: false

The sentinel introduced by pom is as follows
               <dependency>
			<groupId>com.alibaba.csp</groupId>
			<artifactId>sentinel-zuul-adapter</artifactId>
			<version>1.6.0</version>
		</dependency>

###################################
SentinelConfig  is as follows

@Configuration
public class SentinelZuulConfig {

    @Bean
    public SentinelZuulPreFilter sentinelZuulPreFilter(){
        return new SentinelZuulPreFilter(0);
    }

    @Bean
    public SentinelZuulPostFilter sentinelZuulPostFilter(){
        return new SentinelZuulPostFilter(4);
    }

    @Bean
    public SentinelZuulErrorFilter sentinelZuulErrorFilter(){
        return new SentinelZuulErrorFilter();
    }
}

##########################################

GatewayApiRuleConfig 

@Configuration
public class GatewayRulerConfig {
    private static final int URL_MATCH_STRATEGY_EXACT = 0;
    private static final int URL_MATCH_STRATEGY_PREFIX = 1;
    private static final int URL_MATCH_STRATEGY_REGEX = 2;

    @PostConstruct
    public void init(){
        initCustomizedApis();
        initGatewayRules();

    }


    private void initCustomizedApis() {
        Set<ApiDefinition> definitions = new HashSet<>();
        ApiDefinition api1 = new ApiDefinition("baobao_api")
[...]

PR #1605, opened 2020-07-15
Solve the problem of prefix and suffix matching #1109

JiangZian:
Solve the problem of prefix and suffix matching #1109

### DIFF
--- a/sentinel-adapter/sentinel-zuul-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/zuul/api/route/RegexRoutePathMatcher.java
+++ b/sentinel-adapter/sentinel-zuul-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/zuul/api/route/RegexRoutePathMatcher.java
@@ -16,9 +16,9 @@
 package com.alibaba.csp.sentinel.adapter.gateway.zuul.api.route;
 
 import com.alibaba.csp.sentinel.util.AssertUtil;
+import com.alibaba.csp.sentinel.util.StringUtil;
 import com.alibaba.csp.sentinel.util.function.Predicate;
 import com.netflix.zuul.context.RequestContext;
-import com.alibaba.csp.sentinel.util.StringUtil;
 
 import javax.servlet.http.HttpServletRequest;
 import java.util.regex.Pattern;
--- a/sentinel-adapter/sentinel-zuul-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/zuul/api/route/RegexRoutePathMatcher.java
+++ b/sentinel-adapter/sentinel-zuul-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/zuul/api/route/RegexRoutePathMatcher.java
@@ -15,12 +15,13 @@
  */
 package com.alibaba.csp.sentinel.adapter.gateway.zuul.api.route;
 
-import java.util.regex.Pattern;
-
 import com.alibaba.csp.sentinel.util.AssertUtil;
 import com.alibaba.csp.sentinel.util.function.Predicate;
-
 import com.netflix.zuul.context.RequestContext;
+import org.apache.commons.lang.StringUtils;
+
+import javax.servlet.http.HttpServletRequest;
+import java.util.regex.Pattern;
 
 /**
  * @author Eric Zhao
@@ -39,7 +40,13 @@ public class RegexRoutePathMatcher implements Predicate<RequestContext> {
 
     @Override
     public boolean test(RequestContext context) {
-        String path = context.getRequest().getServletPath();
+        //Solve the problem of prefix and suffix matching
+        HttpServletRequest request = context.getRequest();
+        String path = request.getServletPath();
+        String pathInfo = request.getPathInfo();
+        if (StringUtils.length(pathInfo) > 0) {
+            path = path + pathInfo;
+        }
         return regex.matcher(path).matches();
     }
 
--- a/sentinel-adapter/sentinel-zuul-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/zuul/api/route/RegexRoutePathMatcher.java
+++ b/sentinel-adapter/sentinel-zuul-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/zuul/api/route/RegexRoutePathMatcher.java
@@ -16,7 +16,6 @@
 package com.alibaba.csp.sentinel.adapter.gateway.zuul.api.route;
 
 import com.alibaba.csp.sentinel.util.AssertUtil;
-import com.alibaba.csp.sentinel.util.StringUtil;
 import com.alibaba.csp.sentinel.util.function.Predicate;
 import com.netflix.zuul.context.RequestContext;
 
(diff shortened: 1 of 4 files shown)


sentinel-adapter/sentinel-zuul-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/zuul/api/route/RegexRoutePathMatcher.java
```diff
@@ -39,7 +40,13 @@ public RegexRoutePathMatcher(String pattern) {
 
     @Override
     public boolean test(RequestContext context) {
-        String path = context.getRequest().getServletPath();
+        //Solve the problem of prefix and suffix matching
+        HttpServletRequest request = context.getRequest();
+        String path = request.getServletPath();
+        String pathInfo = request.getPathInfo();
+        if (StringUtils.length(pathInfo) > 0) {
+            path = path + pathInfo;
```
lfzCarlosC, line 48: Could there be a case like pathInfo is not trimmed or pathInfo is full of spaces? Isn't apache common StringUtils `isBlank` better?

JiangZian: Ok, this piece is really my incomplete consideration, thank you for correcting, and it has been fixed

sczyh30: Maybe using the embedded `com.alibaba.csp.sentinel.util.StringUtil` is enough.

sczyh30:
Could you please provide a set of test cases regarding how to verify your changes? (e.g. when contextPath=xxx, urlPattern=yyy, the extracted URL is zzz)

sentinel-adapter/sentinel-zuul-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/gateway/zuul/api/route/RegexRoutePathMatcher.java
JiangZian: Ok, this piece is really my incomplete consideration, thank you for correcting, and it has been fixed

JiangZian:
@jasonjoo2010 Could you please review it for me

jasonjoo2010 (commented):
Hi, @JiangZian 

ServletPath() and PathInfo() may have different behaviors between different implementations.
But for `getRequestURI` does it better?
The pair of methods, `getRequestURI` and `getRequestURL`, are more promised. Certainly in this scenario `getRequestURI` should be the one equivalent.

jasonjoo2010:
And besides it's not a good idea in this case to create a mock servlet request object manually.
A more reliable implementation could be used instead like `MockHttpServletRequest` in `spring-test` which can be constructed from specified URI.

sczyh30:
> And besides it's not a good idea in this case to create a mock servlet request object manually.
> A more reliable implementation could be used instead like `MockHttpServletRequest` in `spring-test` which can be constructed from specified URI.

+1 for this. A manually-mocked Servlet request may not be reliable.

JiangZian:
> And besides it's not a good idea in this case to create a mock servlet request object manually.
> A more reliable implementation could be used instead like `MockHttpServletRequest` in `spring-test` which can be constructed from specified URI.

Thanks for your review, I'll try the MockHttpServletRequest method

JiangZian:
> Hi, @JiangZian
> 
> ServletPath() and PathInfo() may have different behaviors between different implementations.
> But for `getRequestURI` does it do better?
> The pair of methods, `getRequestURI` and `getRequestURL`, are more promised. Certainly in this scenario `getRequestURI` should be the one equivalent.

Sorry, I don't quite understand this paragraph. Can you explain The meaning of "The pair of methods, getRequestURI and getRequestURL"

jasonjoo2010:
> > Hi, @JiangZian
> > ServletPath() and PathInfo() may have different behaviors between different implementations.
> > But for `getRequestURI` does it do better?
> > The pair of methods, `getRequestURI` and `getRequestURL`, are more promised. Certainly in this scenario `getRequestURI` should be the one equivalent.
> 
> Sorry, I don't quite understand this paragraph. Can you explain The meaning of "The pair of methods, getRequestURI and getRequestURL"

It could be referred by java doc from javax.servlet-api.
getRequestURI would return "/test/demo" while getRequestURL would return "http://test.com/test/demo?a=3".

sczyh30:
FYI: https://github.com/alibaba/Sentinel/issues/1109#issuecomment-563082969

JiangZian:
> > > Hi, @JiangZian
> > > ServletPath() and PathInfo() may have different behaviors between different implementations.
> > > But for `getRequestURI` does it do better?
> > > The pair of methods, `getRequestURI` and `getRequestURL`, are more promised. Certainly in this scenario `getRequestURI` should be the one equivalent.
> > 
> > 
> > Sorry, I don't quite understand this paragraph. Can you explain The meaning of "The pair of methods, getRequestURI and getRequestURL"
> 
> It could be referred by java doc from javax.servlet-api.
> getRequestURI would return "/test/demo" while getRequestURL would return "http://test.com/test/demo?a=3".

I have implemented a solution, do you see if this solution has a way out?

sentinel-adapter/sentinel-zuul-adapter/src/test/java/com/alibaba/csp/sentinel/adapter/gateway/zuul/route/SentinelZuulRouteTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Test
+    public void testPrefixRoutePathMatche() {
+        PrefixRoutePathMatcher prefixRoutePathMatcher = new PrefixRoutePathMatcher("/servicea/????.jsp");
+        Assert.assertTrue(prefixRoutePathMatcher.test(requestContext));
+    }
+
+    @Test
+    public void testRegexRoutePathMatcher() {
+        RegexRoutePathMatcher regexRoutePathMatcher = new RegexRoutePathMatcher("/servicea[^\\s]+.jsp");
```
jasonjoo2010, line 46: The  pattern `/service[^\s]+.jsp` seems lack of considered.

`\.jsp` should be better than `.jsp`.
And `[^\s]` which indicates non-space subsequence i don't quite understand here because no space is allowed in URI as we know.

Talk to unit test it's better to have both positive and negative cases. So if you familiar with this could you improve it? Otherwise we can improve it later.

jasonjoo2010 (changes requested):
Btw, have you run it in zuul and does it act as you expect? Just a confirmation.

sentinel-adapter/sentinel-zuul-adapter/src/test/java/com/alibaba/csp/sentinel/adapter/gateway/zuul/route/SentinelZuulRouteTest.java
JiangZian: Thank you very much for your Suggestions, but these are my lack of consideration. Now I have resubmitted them

JiangZian: @jasonjoo2010 Do you have time to check my logic

jasonjoo2010 (approved):
Sorry for the latency due to a mountain climbing.
Unit tests don't look like as expected but I think we should fix the bug first. UT can be improved in future.

LGTM

sczyh30:
Thanks for contributing! We may improve the UT later. More PRs are welcomed.

PR #1621, opened 2020-07-21
provide add single rule for xRuleManager

### LINKED ISSUE #1614: problem about FlowRuleManager & DegradeRuleManager put single rule
## Issue Description

Type:  *feature request*

FlowRuleManager & DegradeRuleManager provide `configUpdate()`  method to update Rules by implements PropertyListener. But, `configUpdate()` like `configLoad()`, both of them `clear()` ConcurrentHashMap first, then `putAll()` Rules. Why dont provide a method to support add a new Rule into Map?  This is necessary in the scenario where rule need to be added dynamically.

provide a method like: (eg. in DegradeRuleManager)
```
    public static boolean addRule(DegradeRule rule) {
        if  (null ==rule ) {
            return false;
        }
        return degradeRules.put(rule.getResource(), rule);
    }
```

1. 
2. 
3.

Lovnx:
In a specific scenario, which Rules are not controlled and updated by using Sentinel dashboard, only initialize it and update it by code. you have to add a new Rule only like this below:
```
            List<DegradeRule> rules = DegradeRuleManager.getRules();
            rules.add(rule);
            DegradeRuleManager.loadRules(rules);
```
It may cause concurrency problem.

Fixes #1614 

### Describe how you did it
provide a new method to add single Rule.

### Describe how to verify it
I provide some test case with this PR.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
@@ -15,9 +15,7 @@
  */
 package com.alibaba.csp.sentinel.slots.block.flow;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -90,6 +88,27 @@ public class FlowRuleManager {
     }
 
     /**
+     * add a single FlowRule
+     *
+     * @param rule
+     * @return void
+     */
+    public static void addRule(FlowRule rule){
+        if (isValidRule(rule)) {
+            String resource = rule.getResource();
+            List<FlowRule> oldSet = flowRules.get(resource);
+            if (null != oldSet) {
+                oldSet.add(rule);
+                flowRules.put(resource, oldSet);
+                return;
+            }
+            List<FlowRule> newSet = new ArrayList<>();
+            newSet.add(rule);
+            flowRules.put(resource, newSet);
+        }
+    }
+
+    /**
      * Load {@link FlowRule}s, former rules will be replaced.
      *
      * @param rules new rules to load.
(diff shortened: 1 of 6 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
```diff
@@ -89,6 +87,27 @@ public static void register2Property(SentinelProperty<List<FlowRule>> property)
...
+            String resource = rule.getResource();
+            List<FlowRule> oldSet = flowRules.get(resource);
+            if (null != oldSet) {
+                oldSet.add(rule);
+                flowRules.put(resource, oldSet);
+                return;
+            }
+            List<FlowRule> newSet = new ArrayList<>();
+            newSet.add(rule);
+            flowRules.put(resource, newSet);
```
sczyh30, line 107: All changes regarding rules SHOULD be made via the related `SentinelProperty` (and the rule change listener), as there may be additional logic. It's not allowed to modify the rule map directly.

sentinel-core/src/test/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManagerTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    public void testAddRule() {
+        FlowRule rule1 = new FlowRule("newRule");
+        rule1.setGrade(RuleConstant.FLOW_GRADE_QPS);
+        rule1.setCount(1);
+        rule1.setLimitApp("default");
+        rule1.setStrategy(RuleConstant.STRATEGY_DIRECT);
+        FlowRuleManager.addRule(rule1);
+        assertTrue(FlowRuleManager.hasConfig("newRule"));
```
sczyh30, line 41: Though the "rule" has been added to the rule map, it won't take effect (as its traffic shaping controller was not created).

sczyh30:
> It may cause concurrency problem.

This PR does not fix the concurrency problem.

sczyh30:
Any feedback?

PR #1645, opened 2020-08-04
Refactor the flow of request to fix #1638 for circuit breakers

### LINKED ISSUE #1638: [BUG] Circuit breaker will remain half-open state forever when the request is blocked by upcoming rules
## Issue Description

Type: *bug report*

The circuit breaker won't recover from half-open state when the request is actually blocked by upcoming rules. For example, there are two degrade rules: R1(circuit breaker state=OPEN, recoveryTimeout=10s) and R2(circuit breaker state=OPEN, recoveryTimeout=20s)

There may be circumstances when R1 has reached the recovery timepoint but R2 has not. If a request comes, there will be transformation: **R1(OPEN → HALF-OPEN), R2(OPEN)**

This request will be allowed by R1 but rejected by R2, thus finally blocked. The invocation won't actually occur, so it will **NEVER** complete. For state transformation from HALF-OPEN to OPEN/CLOSED, it should happen only when invocation completes, so for R1 the associated circuit breaker state will be HALF-OPEN forever. Actually this could happen when there are any *blocked rules* after R1 (not only degrade rules).

This is a fatal bug and should be carefully resolved. We may need a temporary workaround for the half-open case in 1.8.0, then improve the overall design later. Discussions are welcomed.

Original discussions can be found in https://github.com/alibaba/Sentinel/pull/1490#discussion_r462079959

jasonjoo2010:

Current implementation of circuit breakers will fire a bug when do examination in half-open state. The examination will be interrupted if the detecting request was blocked by other breakers or slots.

For details please refer to https://github.com/alibaba/Sentinel/pull/1490#discussion_r462079959

#1638

### Describe how you did it

For initial concepts we can illustrate design of circuit breakers, including below scenarios:

Scenarios

And a new design could be shown as:

Flow

In `afterRequestPassed` we:

* Do calculations needed by specific breaker
* Determine the new state(closed -> open, half-open -> open/closed)

In `Exit Handlers` we:

* Execute extra blocks bond to specific entry(Which we should do only if necessary because it will cost extra memory)

Thus, we recycle failed detecting attempt in handler bond to specific entry. `Handlers` also enable users or developers of adapter doing necessary extra work.

### Describe how to verify it

A new unit test can verify it which will fail in previous design: `CircuitBreakingIntegrationTest.testMultipleHalfOpenedBreaders`

### Special notes for reviews

No.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
@@ -17,10 +17,16 @@ package com.alibaba.csp.sentinel.slots.block.degrade.circuitbreaker;
 
 import java.util.concurrent.atomic.AtomicReference;
 
+import com.alibaba.csp.sentinel.Entry;
+import com.alibaba.csp.sentinel.context.Context;
+import com.alibaba.csp.sentinel.slotchain.ResourceWrapper;
 import com.alibaba.csp.sentinel.slots.block.degrade.DegradeRule;
 import com.alibaba.csp.sentinel.slots.block.degrade.DegradeRuleManager;
+import com.alibaba.csp.sentinel.slots.block.degrade.circuitbreaker.CircuitBreaker.State;
 import com.alibaba.csp.sentinel.util.AssertUtil;
 import com.alibaba.csp.sentinel.util.TimeUtil;
+import com.alibaba.csp.sentinel.util.function.BiConsumer;
+import com.alibaba.csp.sentinel.util.function.Consumer;
 
 /**
  * @author Eric Zhao
@@ -61,14 +67,14 @@ public abstract class AbstractCircuitBreaker implements CircuitBreaker {
     }
 
     @Override
-    public boolean tryPass() {
+    public boolean tryPass(Context context, ResourceWrapper r) {
         // Template implementation.
         if (currentState.get() == State.CLOSED) {
             return true;
         }
         if (currentState.get() == State.OPEN) {
             // For half-open state we allow a request for trial.
-            return retryTimeoutArrived() && fromOpenToHalfOpen();
+            return retryTimeoutArrived() && fromOpenToHalfOpen(context);
         }
         return false;
     }
@@ -99,11 +105,24 @@ public abstract class AbstractCircuitBreaker implements CircuitBreaker {
         return false;
     }
 
-    protected boolean fromOpenToHalfOpen() {
+    protected boolean fromOpenToHalfOpen(Context context) {
         if (currentState.compareAndSet(State.OPEN, State.HALF_OPEN)) {
             for (CircuitBreakerStateChangeObserver observer : observerRegistry.getStateChangeObservers()) {
                 observer.onStateChange(State.OPEN, State.HALF_OPEN, rule, null);
             }
+            Entry entry = context.getCurEntry();
+            entry.whenComplete(new BiConsumer<Context, Entry>() {
+                
+                @Override
+                public void accept(Context context, Entry entry) {
+                    if (entry.getBlockError() != null) {
+                        // blocked
+                        currentState.compareAndSet(State.HALF_OPEN, State.OPEN);
+                        return;
+                    }
+                    
+                }
+            });
             return true;
         }
         return false;
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
@@ -22,11 +22,9 @@ import com.alibaba.csp.sentinel.context.Context;
 import com.alibaba.csp.sentinel.slotchain.ResourceWrapper;
 import com.alibaba.csp.sentinel.slots.block.degrade.DegradeRule;
 import com.alibaba.csp.sentinel.slots.block.degrade.DegradeRuleManager;
-import com.alibaba.csp.sentinel.slots.block.degrade.circuitbreaker.CircuitBreaker.State;
 import com.alibaba.csp.sentinel.util.AssertUtil;
 import com.alibaba.csp.sentinel.util.TimeUtil;
 import com.alibaba.csp.sentinel.util.function.BiConsumer;
-import com.alibaba.csp.sentinel.util.function.Consumer;
 
 /**
  * @author Eric Zhao
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
@@ -95,9 +95,7 @@ public abstract class AbstractCircuitBreaker implements CircuitBreaker {
         if (currentState.compareAndSet(prev, State.OPEN)) {
             updateNextRetryTimestamp();
 
-            for (CircuitBreakerStateChangeObserver observer : observerRegistry.getStateChangeObservers()) {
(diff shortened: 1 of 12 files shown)


sczyh30:
cc @cdfive @wavesZh @louyuting

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
```diff
@@ -99,11 +103,24 @@ protected boolean fromCloseToOpen(double snapshotValue) {
...
     }
 
-    protected boolean fromOpenToHalfOpen() {
+    protected boolean fromOpenToHalfOpen(Context context) {
         if (currentState.compareAndSet(State.OPEN, State.HALF_OPEN)) {
             for (CircuitBreakerStateChangeObserver observer : observerRegistry.getStateChangeObservers()) {
                 observer.onStateChange(State.OPEN, State.HALF_OPEN, rule, null);
             }
+            Entry entry = context.getCurEntry();
+            entry.whenComplete(new BiConsumer<Context, Entry>() {
```
wavesZh, line 112: 1. Thread-safe for`whenComplete`.
2. Other requests blocked by half-open state also can transform half-open to open, so the one request allowed by half-state maybe not  transform half-open to close.

jasonjoo2010: > * Thread-safe for`whenComplete`.
> * Other requests blocked by half-open state also can transform half-open to open, so the one request allowed by half-state maybe not  transform half-open to close.

Please refer to the workflow chart above.
I think the design of `Entry` is non-thread-safe because it stores information of single request. One request cannot be paralleled, right?

Talk to this piece of code the mutex between entries(requests) is ensured by CAS operation (from OPEN to HALF-OPEN). So only one request can perform `open -> half -> open` or `open -> half -> closed`.

Or is there any other scenario I missed?

wavesZh: my mistake, no problem.

jasonjoo2010: 😀

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/CircuitBreaker.java
```diff
@@ -46,13 +50,12 @@
...
-     * handle state transformation of the circuit breaker.
+     * Called when a passed invocation finished.
      *
-     * @param rt the response time of this entry
-     * @param error the error of this entry (if present)
+     * @param context context of current invocation
+     * @param wrapper current resource
      */
-    void onRequestComplete(long rt, Throwable error);
+    void afterRequestPassed(Context context, ResourceWrapper wrapper);
```
sczyh30, line 58: Actually IMHO "complete" might be more concrete. In Sentinel, "pass" means allowed by Sentinel (not started but pending), and "complete" means an invocation has completed (the blocked requests never started, so never completed).

sczyh30: And the `ResourceWrapper` seems unused? (it could be taken from `Entry` in `Context`)

jasonjoo2010: > pending

Yeah naming issues often have more than one answer mostly. I tried to make it clear for requests passed by sentinel(`passed`) and it will be invoked as it completed(`after`). Surely we can have other picks. Is `onRequestComplete` implicit whether sentinel guarantees it would be called no matter whether it was blocked?

Talk to the parameters in definition it just follows similar designs in related logics and try to make it more stable in future iterations. `ResourceWrapper` will contains the resource information which it's useful especially in implementations in future. My opinion is more like to keep it as stability of interface.

jasonjoo2010: > pending

Yeah naming issues often have more than one answer mostly. I tried to make it clear for requests passed by sentinel(`passed`) and it will be invoked as it completed(`after`). Surely we can have other picks. Is `onRequestComplete` implicit whether sentinel guarantees it would be called no matter whether it was blocked?

Talk to the parameters in definition it just follows similar designs in related logics and try to make it more stable in future iterations. `ResourceWrapper` will contains the resource information which it's useful especially in implementations in future. My opinion is more like to keep it as stability of interface.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/AbstractCircuitBreaker.java
```diff
@@ -99,11 +103,24 @@ protected boolean fromCloseToOpen(double snapshotValue) {
...
+                
+                @Override
+                public void accept(Context context, Entry entry) {
+                    if (entry.getBlockError() != null) {
+                        // blocked
+                        currentState.compareAndSet(State.HALF_OPEN, State.OPEN);
```
louyuting, line 118: Maybe the state changes need to notify **CircuitBreakerStateChangeObserver**.

jasonjoo2010: > Maybe the state changes need to notify **CircuitBreakerStateChangeObserver**.

Yes, absolutely.

It has been updated.

sczyh30: It's required, but this could be confusing for users (as the internal reverting is not expected for users)😂

jasonjoo2010: > It's required, but this could be confusing for users (as the internal reverting is not expected for users)😂

Yeah indeed but the conditions would be hard to met.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/CircuitBreaker.java
sczyh30: > Yeah naming issues often have more than one answer mostly. I tried to make it clear for requests passed by sentinel(`passed`) and it will be invoked as it completed(`after`). Surely we can have other picks. Is `onRequestComplete` implicit whether sentinel guarantees it would be called no matter whether it was blocked?

Originally the semantics of `onComplete` means the invocation has completed, so it MUST have passed Sentinel's checking. The rejected invocations never passed, so it would never complete.

> Talk to the parameters in definition it just follows similar designs in related logics and try to make it more stable in future iterations. `ResourceWrapper` will contains the resource information which it's useful especially in implementations in future. My opinion is more like to keep it as stability of interface.

Okay for me. Actually I'd like the circuit breaker to be a more generic design (e.g. the context could be generic, and metadata could be retrieved from it).

jasonjoo2010: > Okay for me. Actually I'd like the circuit breaker to be a more generic design (e.g. the context could be generic, and metadata could be retrieved from it).

Naming of callback has been reverted and it has more comments on it.
Talk to the generic context the idea makes sense but it also may introduce more memory footprints. Maybe it can be decided in future.

sczyh30: Yes we may discuss it later in a new issue.

sczyh30:
Nice work. Thanks!

Issue #1661, opened 2020-08-12
fastjson 1.2.7 deserialization problem ：cause the rule configuration to not take effect


fllife:
## Issue Description

Type: bug report

### Describe what happened (or what feature you want)

when I use fastjson 1.2.7 in my application,  config flow rule not take effect by sentinel dashboard,

### Describe what you expected to happen

flow rule  which i config by sentinel dashboard  can take effect 
### How to reproduce it (as minimally and precisely as possible)

1.  add dependency com.alibaba:fastjson:1.2.7 in application
2.  config rule by sentinel dashbord
3.

PR #1663, opened 2020-08-12
fixd #1661: fastjson deserialization problem

fllife:
in ModifyRulesCommandHandler linenumber 66:
```java
List<FlowRule> flowRules = JSONArray.parseArray(data, FlowRule.class);
FlowRuleManager.loadRules(flowRules);
```
when I use fastjson 1.2.7, field `resourse` and` limitApp can not be set value ，cause config rule not take effect because class AbstractRule do not have setter method

fixd #1661
### Describe how you did it
class AbstractRulemake setter method and not return value

### Describe how to verify it
when I use fastjson 1.2.7, filed resourse and limitApp can be set value and config rule can take effect


jasonjoo2010:
Actually fastjson before 1.2.12 (exclusive) is not usable because it's buggy to those classes having inheritance.
Refer to https://github.com/alibaba/Sentinel/blob/master/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/command/handler/ModifyRulesCommandHandler.java#L54

And another solution is under progress on these kinds of mess on #1739 
You can have a look or do some reviews kindly.

sczyh30:
Closing this PR due to no further feedback. Thanks for contributing.

PR #1665, opened 2020-08-12
Enhancement/advanced metric extension tagging entrytype

Billzaifei:
Describe what this PR does / why we need it
For the below interface, We implement it to export prometheus metrics.
com.alibaba.csp.sentinel.metric.extension.MetricExtension

Take the method addPass(String resource, int n, Object... args) for example.
We want to distinguish [in] type from [out] type using it as the prometheus tag.

provider[in]/consumer[out]

Does this pull request fix one issue?
Resolves #1566

Describe how you did it
Enhance com.alibaba.csp.sentinel.metric.extension.MetricExtension with com.alibaba.csp.sentinel.metric.extension.AdvancedMetricExtension

Describe how to verify it
Run test cases.

Special notes for reviews

PR #1675, opened 2020-08-18
Refactor the extended MetricExtension interface to match fundamental events in Sentinel

sczyh30:

Refactor the extended `AdvancedMetricExtension` interface (new added in #1665) to match fundamental events (pass, block, complete) in Sentinel.

Improvement for #1665 

### Describe how you did it

- Unify the extended interface as a few event handlers: onPass, onBlocked, onComplete and onError
- Polish related code

### Describe how to verify it

Run all test cases for regression.

### Special notes for reviews

NONE

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/metric/extension/AdvancedMetricExtension.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/metric/extension/AdvancedMetricExtension.java
@@ -1,97 +1,74 @@
+/*
+ * Copyright 1999-2019 Alibaba Group Holding Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.alibaba.csp.sentinel.metric.extension;
 
 import com.alibaba.csp.sentinel.EntryType;
+import com.alibaba.csp.sentinel.slotchain.ResourceWrapper;
 import com.alibaba.csp.sentinel.slots.block.BlockException;
 
 /**
- * Advanced {@link MetricExtension} extending input parameters of each metric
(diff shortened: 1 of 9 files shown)


jasonjoo2010:
I have done a quick overview on this pr. 
First as a background the previous contribution has never ever been released (#1665), right?

So my concern is that why we insist on keeping the inheritance between event style interface and counter style interface(legacy one)? Since anyone who wants to implement "Advanced" one must implement interfaces, addPass/onPass, addBlock/onBlock, etc., altogether even he only want to follow the method group of event style. Is there any constraint we must merge them into a base, non-empty interface type?

Normally and mostly we may use an empty base interface for referencing and do type detection later. Sure for backward compatibility this way should not be used.

Another thought is that we can separate them into independent interfaces and maintain in two registries or collections.

I will continue to do a quick review after we make these clear.

sczyh30:
> I have done a quick overview on this pr.
> First as a background the previous contribution has never ever been released (#1665), right?

Yes, it's not released yet.

> So my concern is that why we insist on keeping the inheritance between event style interface and counter style interface(legacy one)? Since anyone who wants to implement "Advanced" one must implement interfaces, addPass/onPass, addBlock/onBlock, etc., altogether even he only want to follow the method group of event style. Is there any constraint we must merge them into a base, non-empty interface type?

> Normally and mostly we may use an empty base interface for referencing and do type detection later. Sure for backward compatibility this way should not be used.
> 
> Another thought is that we can separate them into independent interfaces and maintain in two registries or collections.

This is only a temporary workaround for backward compatibility (as some other libraries has been integrating with the SPI). Actually I was planning to deprecate the legacy `MetricExtension` hierarchy in upcoming versions and make the metric hook as the "canonical" event style.

jasonjoo2010:
> This is only a temporary workaround for backward compatibility (as some other libraries has been integrating with the SPI). Actually I was planning to deprecate the legacy `MetricExtension` hierarchy in upcoming versions and make the metric hook as the "canonical" event style.

Understood.
So how does the canonical look like? Is it still in designing?
If so we may mainly do some renaming in this pr and abandon it after a few iterations,  right?

sczyh30:
> So how does the canonical look like? Is it still in designing?

It's not settled down. Discussions are needed later.

> If so we may mainly do some renaming in this pr and abandon it after a few iterations, right?

That's it.

jasonjoo2010:
> > So how does the canonical look like? Is it still in designing?
> 
> It's not settled down. Discussions are needed later.
> 
> > If so we may mainly do some renaming in this pr and abandon it after a few iterations, right?
> 
> That's it.

Got it now and  I  have no further question.

jasonjoo2010 (approved):
LGTM

PR #1680, opened 2020-08-19
Fix bugs and improve default fallback logic of Apache Dubbo adapter

sczyh30:

There are some problems in Apache Dubbo adapter (since 1.8.0-SNAPSHOT):

- Unexpected behavior will occur when there are non-biz errors (e.g. Dubbo RPC timeout), which is caused by NPE in whenComplete hook in consumer filter.
- The default fallback logic (wrap an error result with SentinelRpcException, which wraps a BlockException) may cause error when deserialization.

This PR fixes the bug and improves the default fallback of Apache Dubbo 2.7.x adapter.

NONE

### Describe how you did it

- Add RuntimeException converting method (`toRuntimeException`) in BlockException and polish logic for `BlockException` validation
- Fix NPE bug in consumer filter (when non-biz error occurred)
- Improve default fallback in Dubbo 2.7.x adapter: convert the BlockException to a simple RuntimeException (with necessary message)
- Polish code and comments

### Describe how to verify it

Run the test cases.

### Special notes for reviews

NONE

### DIFF
--- a/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboConsumerFilter.java
+++ b/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboConsumerFilter.java
@@ -19,11 +19,13 @@ import com.alibaba.csp.sentinel.*;
 import com.alibaba.csp.sentinel.adapter.dubbo.config.DubboAdapterGlobalConfig;
 import com.alibaba.csp.sentinel.log.RecordLog;
 import com.alibaba.csp.sentinel.slots.block.BlockException;
+
 import org.apache.dubbo.common.extension.Activate;
 import org.apache.dubbo.rpc.*;
 import org.apache.dubbo.rpc.support.RpcUtils;
 
 import java.util.LinkedList;
+import java.util.Optional;
 import java.util.function.BiConsumer;
 
 import static org.apache.dubbo.common.constants.CommonConstants.CONSUMER;
@@ -38,6 +40,7 @@ import static org.apache.dubbo.common.constants.CommonConstants.CONSUMER;
  *
  * @author Carpenter Lee
  * @author Eric Zhao
+ * @author Lin Liang
  */
 @Activate(group = CONSUMER)
 public class SentinelDubboConsumerFilter extends BaseSentinelDubboFilter {
@@ -64,7 +67,6 @@ public class SentinelDubboConsumerFilter extends BaseSentinelDubboFilter {
         } else {
             return asyncInvoke(invoker, invocation);
         }
-
     }
 
     private Result syncInvoke(Invoker<?> invoker, Invocation invocation) {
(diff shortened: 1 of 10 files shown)


sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboConsumerFilter.java
```diff
@@ -98,24 +101,26 @@ private Result syncInvoke(Invoker<?> invoker, Invocation invocation) {
...
-                        EntryHolder holder = queue.pop();
-                        Tracer.traceEntry(result.getException(), holder.entry);
-                        exitEntry(holder);
-                    }
+            result.whenCompleteWithContext((r, throwable) -> {
+                while (!queue.isEmpty()) {
+                    EntryHolder holder = queue.pop();
+                    Throwable error = Optional.ofNullable(r)
+                        .map(Result::getException)
+                        .orElse(throwable);
```
jasonjoo2010, line 121: Maybe the priority list is `throwable` -> `r.exception` indicating `throwable` is more fatal. How do you think about it?

jasonjoo2010 (changes requested):

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboConsumerFilter.java
sczyh30: Updated, please check.

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboConsumerFilter.java
```diff
@@ -98,24 +101,27 @@ private Result syncInvoke(Invoker<?> invoker, Invocation invocation) {
...
-                @Override
-                public void accept(Result result, Throwable throwable) {
-                    while (!queue.isEmpty()) {
-                        EntryHolder holder = queue.pop();
-                        Tracer.traceEntry(result.getException(), holder.entry);
-                        exitEntry(holder);
+            result.whenCompleteWithContext((r, throwable) -> {
+                while (!queue.isEmpty()) {
+                    EntryHolder holder = queue.pop();
+                    Throwable error = throwable;
```
jasonjoo2010, line 119: It's better to set error outside the loop because there are always more than one elements in stack.

jasonjoo2010 (changes requested):

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboConsumerFilter.java
sczyh30: Updated

jasonjoo2010 (approved):
LGTM

There are some unnecessary changes in pr and I suggest we can avoid it in future to keep pr small and simple.

jasonjoo2010:
Thanks for your contribution!

PR #1533, opened 2020-06-08
webmvc-adapter: improve set entry in request to avoid ErrorEntryFreeException

cdfive:
### Describe what this PR does / why we need it

The `AbstractSentinelInterceptor` of `sentinel-spring-webmvc-adapter` may enter twice under some circumstances, such no view page exists 

### Does this pull request fix one issue?

Relate to #1531, #1482 

### Describe how you did it

Refator `setEntryInRequest` method, if attribute key already exists in request, we don't call `SphU.entry` to create Entry, then Entry get from request can exit normally in `afterCompletion` method.

### Describe how to verify it

Add a tmp test Controller in `sentinel-demo-spring-webmvc` to verify if it works.
```java
@Controller
@RequestMapping("/")
public class TmpTestController {

    @RequestMapping("/view_not_exists")
    public String view_not_exists() {
        return "/view_not_exists.html";
    }
}
```

### Special notes for reviews

The tmp test Controller is used for debug and test, so it has not been committed.

PR #1681, opened 2020-08-19
Improve compatibility for re-dispatched servlet request

jasonjoo2010:
A servlet can be dispatched multiple times in its lifecycle if any redirecting / including action happened or it's an asynchronous request. During these dispatching interpreters will also be called passed exactly the same `request` each time. Besides these the initial dispatching of request wraps all following re-dispatching.

Supplement for pr #1533 

### Describe how you did it
Sentinel will take the initial dispatching of request as target resource while ignore all following re-dispatching. Because the RT of outer dispatching is more precise and meaningful.

### Describe how to verify it
I write a new api named `/forward` implementing a one-child-dispathing sample in `sentinel-demo-spring-webmvc`.

### Special notes for reviews
Nope.


sczyh30 (approved):
LGTM

sczyh30:
Thanks!

Issue #1522, opened 2020-06-05
fastjson 经常出漏洞问题  建议替换fastjson或者增加其他json库适配


woshioosm:
source的读取是可以自定义json库读取的 但是在sentinel-transport-common中是依赖fastjson的 这里能否替代掉 或者可适配其他json库

Warkeeper:
Strongly agree.  
Since the metrics API is enabled by default, it's easy to construct a critical JSON message and send it to a application which uses sentinel. So if sentinel strongly depends a vulnerable json libary like FastJson, it may cause the application easy to attack.
For now, my team uses javassist to mock fastjson-releated class to deceive the Sentinel while completely excludes the fastjson, and uses other json lib as the implementation. But in the long term, I think sentinel should replace the dependency of fastjson with some kind of adapter.

sczyh30:
@jasonjoo2010 @cdfive @linlinisme @zhaoyuguang Any thoughts?

jasonjoo2010:
Im ok about that. 
Or we can just disable insecure features like `autotype`.

As i knew before `autotype` is disabled by default and we also can use our own `ParserConfig` to make sure it running in secured mode.

So is there any other insecure feature or other reason that it should be abandoned?

Warkeeper:
According to this doc [https://github.com/alibaba/fastjson/wiki/security_update_20200601](url) ,  it seems like even if we enable ```safeMode``` to disable ```autoType```, the security still cannot be guaranteed.

> 配置safeMode后，无论白名单和黑名单，都不支持autoType，可**一定程度上缓解**反序列化Gadgets类变种攻击

jasonjoo2010:
> According to this doc [https://github.com/alibaba/fastjson/wiki/security_update_20200601](url) , it seems like even if we enable `safeMode` to disable `autoType`, the security still cannot be guaranteed.
> 
> > 配置safeMode后，无论白名单和黑名单，都不支持autoType，可**一定程度上缓解**反序列化Gadgets类变种攻击

Oh, i see and thank you for your information.

We indeed can make adaptions to kinds of json parsers like fastjson, jackson, gson, etc. 
But notice that not only fastjson but also jackson i remembered also has vulnerable versions. 
So adaptions may not really solve the security issue but give ability to users migrating to another. 
Users may also use a vulnerable version of parser but they don't provide public entrance like HTTP. They are safe in such scenario.

[...]

PR #1739, opened 2020-09-16
[serialization]Abstract JSON serialization and default to use gson

jasonjoo2010:

Though sentinel-core doesn't need json serialization, but in most sentinel modules (adaption, datasource, etc.) they use fastjson for serializing/deserializing json strings.

Unfortunately there are more and more versions of fastjson which were found vulnerable and users must follow the latest patches or versions. If not their application may become weak and easy to attack.

In the other hand JSON serializing is a common need. So most projects have their own way to do that. In this scene it's not necessary to introduce a new one if sentinel can use existing, compatible serializing library in project.

This PR will enable user letting sentinel to use gson / fastjson / jackson or their own implementation through SPI protocol.

#1522 

### Describe how you did it

Introduce a new module `sentinel-serialization` and take `gson` as default json serializing/deserializing library.

- Add new SPI interface `JsonSerializer` and `JsonDeserializer`
- The default implementation adapts to `gson` and `fastjson` (gson prior)
- `gson` can be excluded manually from dependency tree if use custom implementation
- Add a popular implementation `sentinel-serialization-jackson`

### Describe how to verify it

Depends.

### Special notes for reviews

A wide regressing test may be necessary.


zhengpengGitHub:
hello，when does this fix will release?

sentinel-serialization/sentinel-serialization-common/src/main/java/com/alibaba/csp/sentinel/serialization/common/fallback/DefaultTransformer.java
```diff
@@ -0,0 +1,86 @@
...
+    public <T> T deserialize(String str, Type type) {
+        if (StringUtil.isBlank(str)) {
+            return null;
+        }
+        try {
+            return this.adaption.deserialize(str, type);
+        } catch (Exception e) {
+        }
+        return null;
+    }
```
cdfive, line 72: When `serialize` and `deserialize` the exception is ignored, do we need to throw an exception to remind the user to know or record the error log? The same is in `JacksonTransformer` class.

jasonjoo2010: Here just keep the dealing strategy for backward compatibility. (null when failed in most scenarios)
So some logs are needed or a runtime exception should be introduced?

cdfive: I suggest both throw runtime exception and record error log. Although the exception will be occur in rare case, maybe never, but since we catch the Exception, why not do nothing. If exception occurred, user may want to know what happened. If it's sure that we don't need to deal with it here, suggest adding some comments.

sentinel-serialization/sentinel-serialization-common/src/main/java/com/alibaba/csp/sentinel/serialization/common/fallback/FastjsonAdaption.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public String serialize(Object obj) throws Exception {
+        return (String)METHOD_TO_JSON_STRING.invoke(CLASS, obj);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T> T deserialize(String str, Type type) throws Exception {
+        return (T)METHOD_PARSE_OBJECT_BY_TYPE.invoke(CLASS, str, type, null);
+    }
```
cdfive, line 80: Serialization and deserialization methods called by reflection in `FastjsonAdaption` and `GsonAdaption`. Another way is that, the `fastjson` and `gson` depenency can be with `provided` scope, and we can use their class and method calls directly. Which way do you think is better? @jasonjoo2010 @sczyh30

sentinel-serialization/sentinel-serialization-common/src/main/java/com/alibaba/csp/sentinel/serialization/common/fallback/DefaultTransformer.java
```diff
@@ -0,0 +1,86 @@
...
+ *
+ */
+public class DefaultTransformer implements JsonSerializer, JsonDeserializer {
+    private Adaption adaption;
+    
+    public DefaultTransformer() {
+        List<Adaption> adaptions = Arrays.asList(
+            new GsonAdaption(), 
+            new FastjsonAdaption()
+        );
```
cdfive, line 40: `Jackson` seems more widely used, `Gson` as the default first choice maybe need discuss. 
`sentinel-serialization-jackson` is a separate `module`, why not a `JasksonAdaption`, will it be more convenient to use

sentinel-serialization/sentinel-serialization-common/src/main/java/com/alibaba/csp/sentinel/serialization/common/fallback/DateTypeAdapter.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonPrimitive;
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
+
+class DateTypeAdapter implements JsonSerializer<Date>, JsonDeserializer<Date> {
```
cdfive, line 29: `GsonDateTypeAdapter` may be a better name, or add separate package for gson and fastjson fallback.

sentinel-serialization/sentinel-serialization-common/src/main/java/com/alibaba/csp/sentinel/serialization/common/fallback/DefaultTransformer.java
cdfive: If throw runtime exception, it'will affect the test cases, or just record log? I think it's OK to ignore the exception
, just a thought of it for reference,  it's up to you.

brotherlu-xcq:
Hi, @jasonjoo2010. I have some question about this design. 
- I find that you split deserialize and serialize to two interface. but in our scene，the relationship bewteen deserialize and serialize should be like the front and back of a coin. It maybe better to put them in an interface.
- serialize type is not only json, I think design of this module shoud cover other serialize type like hession,etc.

jasonjoo2010:
> Hi, @jasonjoo2010. I have some question about this design.
> 
> * I find that you split deserialize and serialize to two interface. but in our scene，the relationship bewteen deserialize and serialize should be like the front and back of a coin. It maybe better to put them in an interface.
> * serialize type is not only json, I think design of this module shoud cover other serialize type like hession,etc.

Hi Lu

First let's skip the first point you mentioned and go over the second point directly.  
The purpose of this PR is aiming to decouple the serialising/deserialising on json string against fastjson library because of more and more versions of that have security risks. That's why it focuses on dealing json only.  
But if open our minds to other general purpose serialisations it will be too flexible. Which will make things complicated especially the data will be submitted through http transports while hessian/proto/igbinary is binary serialisation. So it's not a RPC case.  
So we want to keep json as the standard encoding in transports.  

Come back to the first point not in java but in many languages they are two interfaces(eg. golang). You even can only customize one of them instead of both which make more sense. In the other hand you must name it something new if you want to combine them together. So it's not a big deal I think.  
Let me use a more detailed example to make it clear. If you use gson but you are not satisfied with the provided implementation because you want all float type to keep up to 2 decimal places. So if you just want to truncate the inputs from outside you only need to implement a new deserialiser.

Issue #1730, opened 2020-09-14
Dependency conflicts on org.apache.httpcomponents:httpcore, leading to inconsistent program behaviors


HelloCoCooo:
Hi, in **Sentinel/sentinel-transport/sentinel-transport-netty-http**, there are mulptiple versions of library **org.apache.httpcomponents:httpcore**. However, according to Maven's dependency management strategy: **_"first declaration wins"_**, only **org.apache.httpcomponents:httpcore:4.4.5** can be loaded, and **org.apache.httpcomponents:httpcore:4.4.6** will be shadowed.

As shown in the following figure, your project expects to invoke method **_<org.apache.http.impl.DefaultConnectionReuseStrategy: keepAlive(Lorg/apache/http/HttpResponse;Lorg/apache/http/protocol/HttpContext;)Z>_**  in library _**org.apache.httpcomponents:httpcore:4.4.6**_ (along the _original dependency path_). As it has been shadowed, this method defined in **_org.apache.httpcomponents:httpcore:4.4.5_** is actually forced to be referenced via the following invocation path (along the _actual dependency path_):

```
<com.alibaba.csp.sentinel.transport.heartbeat.HttpHeartbeatSender: sendHeartbeat()Z> /home/wwww/wangSensor/unzip/Sentinel-1.8.0/sentinel-transport/sentinel-transport-netty-http/target/classes
<org.apache.http.impl.client.CloseableHttpClient: execute(Lorg/apache/http/client/methods/HttpUriRequest;)Lorg/apache/http/client/methods/CloseableHttpResponse;> /home/wwww/.m2/repository/org/apache/httpcomponents/httpclient/4.5.3/httpclient-4.5.3.jar
<org.apache.http.impl.client.CloseableHttpClient: execute(Lorg/apache/http/client/methods/HttpUriRequest;Lorg/apache/http/protocol/HttpContext;)Lorg/apache/http/client/methods/CloseableHttpResponse;> /home/wwww/.m2/repository/org/apache/httpcomponents/httpclient/4.5.3/httpclient-4.5.3.jar
<org.apache.http.impl.client.AbstractHttpClient: doExecute(Lorg/apache/http/HttpHost;Lorg/apache/http/HttpRequest;Lorg/apache/http/protocol/HttpContext;)Lorg/apache/http/client/methods/CloseableHttpResponse;> /home/wwww/.m2/repository/org/apache/httpcomponents/httpclient/4.5.3/httpclient-4.5.3.jar
[...]

PR #1740, opened 2020-09-17
Fix the dependency conflict issue

HelloCoCooo:
Fix the dependency conflict issue #1730 by Remove the conflicting Jar **org.apache.httpcomponents:httpcore:4.4.5** in pom file.
This fix will not affect other libraries or class, except the above duplicate class.


jasonjoo2010 (approved):
LGTM

jasonjoo2010:
Thanks for contributing!

PR #1745, opened 2020-09-18
feat: 添加一个property "io.netty.epoll"来切换使用netty官方实现的epoll(ET)

mikawudi:
添加一个property "io.netty.epoll"来切换使用netty官方实现的epoll(ET)

### DIFF
--- a/sentinel-cluster/sentinel-cluster-server-default/pom.xml
+++ b/sentinel-cluster/sentinel-cluster-server-default/pom.xml
@@ -35,6 +35,12 @@
             <groupId>io.netty</groupId>
             <artifactId>netty-handler</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport-native-epoll</artifactId>
+            <version>${netty.version}</version>
+            <classifier>linux-x86_64</classifier>
+        </dependency>
 
         <dependency>
             <groupId>com.alibaba.csp</groupId>
--- a/sentinel-cluster/sentinel-cluster-server-default/src/main/java/com/alibaba/csp/sentinel/cluster/server/NettyTransportServer.java
+++ b/sentinel-cluster/sentinel-cluster-server-default/src/main/java/com/alibaba/csp/sentinel/cluster/server/NettyTransportServer.java
@@ -28,10 +28,9 @@ import com.alibaba.csp.sentinel.log.RecordLog;
 
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.buffer.PooledByteBufAllocator;
-import io.netty.channel.ChannelFuture;
-import io.netty.channel.ChannelInitializer;
-import io.netty.channel.ChannelOption;
-import io.netty.channel.ChannelPipeline;
+import io.netty.channel.*;
+import io.netty.channel.epoll.EpollEventLoopGroup;
+import io.netty.channel.epoll.EpollServerSocketChannel;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.SocketChannel;
 import io.netty.channel.socket.nio.NioServerSocketChannel;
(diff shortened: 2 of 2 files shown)


sentinel-cluster/sentinel-cluster-server-default/src/main/java/com/alibaba/csp/sentinel/cluster/server/NettyTransportServer.java
```diff
@@ -50,15 +49,17 @@
  */
 public class NettyTransportServer implements ClusterTokenServer {
 
+    private static final int DEFAULT_BOSS_EVENT_LOOP_THREADS = 1;
     private static final int DEFAULT_EVENT_LOOP_THREADS = Math.max(1,
         SystemPropertyUtil.getInt("io.netty.eventLoopThreads", Runtime.getRuntime().availableProcessors() * 2));
+    private static final boolean useEpoll = SystemPropertyUtil.getBoolean("io.netty.epoll", false);
```
jasonjoo2010, line 55: I suggest moving it to `start()` and users maybe easier to set the properties dynamically.
How do you think it?

jasonjoo2010 (changes requested):
And in this PR we limit the dependency under `linux` architecture and `epoll` also only available in linux. That may not make sense for other platforms.

So I have another proposal to discuss.
Compared to introduce a new option/configuration to user,  why not we make supports for `kqueue` (unix) and `epoll` (linux) then make it as a higher priority option when available (using reflection, user introduce the correct dependency manually)?

sentinel-cluster/sentinel-cluster-server-default/src/main/java/com/alibaba/csp/sentinel/cluster/server/NettyTransportServer.java
mikawudi: get property in the <clinit> because DEFAULT_EVENT_LOOP_THREADS get property also in the <clinit>
but you are right,i will move it to start() method

sentinel-cluster/sentinel-cluster-server-default/src/main/java/com/alibaba/csp/sentinel/cluster/server/NettyTransportServer.java
```diff
@@ -74,12 +74,13 @@ public void start() {
         if (!currentState.compareAndSet(SERVER_STATUS_OFF, SERVER_STATUS_STARTING)) {
             return;
         }
-
+        boolean useEpoll = SystemPropertyUtil.getBoolean("io.netty.epoll", false);
```
liangxiao0315, line 77: 这里可以用 io.netty.channel.epoll.Epoll.isAvailable() 这个方法来自动判断，不需要用户配置

sentinel-cluster/sentinel-cluster-server-default/pom.xml
```diff
@@ -35,6 +35,12 @@
             <groupId>io.netty</groupId>
             <artifactId>netty-handler</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport-native-epoll</artifactId>
+            <version>${netty.version}</version>
+            <classifier>linux-x86_64</classifier>
+        </dependency>
```
liangxiao0315, line 43: 不需要这个依赖

sentinel-cluster/sentinel-cluster-server-default/src/main/java/com/alibaba/csp/sentinel/cluster/server/NettyTransportServer.java
jasonjoo2010: +1

Or @mikawudi you can also take my review comment as a sample way. No new "hard" dependency introduced will be better.

mikawudi: emm,就改造成自动根据平台以及用户是否自己引入了native包来决定使用什么EventLoopGroup撒,我这边尝试修改一下再提交,最近公司项目比较忙,没怎么看这边,不好意思~

LearningGp:

Issue #1702, opened 2020-08-28
com.alibaba.csp.sentinel.util.TimeUtil high CPU usage


wenit:
## Issue Description

Type: *bug report* or *feature request*

### Describe what happened (or what feature you want)

com.alibaba.csp.sentinel.util.TimeUtil 中死循环，占用了50%左右的cpu

代码如下：

``` java
static {
        currentTimeMillis = System.currentTimeMillis();
        Thread daemon = new Thread(new Runnable() {
            @Override
            public void run() {
                while (true) {
                    currentTimeMillis = System.currentTimeMillis();
                    try {
                        TimeUnit.MILLISECONDS.sleep(1);
                    } catch (Throwable e) {

                    }
                }
            }
        });
        daemon.setDaemon(true);
        daemon.setName("sentinel-time-tick-thread");
        daemon.start();
    }
```

### How to reproduce it (as minimally and precisely as possible)

1. 
2. 
3. 

### Tell us your environment

sentinel版本1.7.1

redhat7.4

jdk1.8

jasonjoo2010:
Is it necessary to improve TimeUtil as self adjustable model? @sczyh30 

Kind of initial thought:

<img width="765" alt="Screen Shot 2020-09-14 at 23 55 49" src="https://user-images.githubusercontent.com/17573073/93108903-fa916e00-f6e5-11ea-9d69-e531c353565f.png">

PR #1746, opened 2020-09-18
Improve performance of TimeUtil in different load conditions

jasonjoo2010:

Current design of TimeUtil may cost much more than users think in mind in idle systems.

 #1702

### Describe how you did it

To achieve this by making TimeUtil have multiple running states:

- IDLE: Direct invocation state for idle conditions.
- RUNNING: Legacy mode for busy conditions.
- REQUEST: Temporary state from IDLE to RUNNING.

By this design TimeUtil won't cost much in idle systems any more.

### Describe how to verify it

Refer to the unit test `TimeUtilTest` or generate different loads to demo projects.

### Special notes for reviews

No

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/TimeUtil.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/TimeUtil.java
@@ -15,38 +15,202 @@
  */
 package com.alibaba.csp.sentinel.util;
 
+import java.util.List;
 import java.util.concurrent.TimeUnit;
+
+import com.alibaba.csp.sentinel.log.RecordLog;
+import com.alibaba.csp.sentinel.slots.statistic.base.LeapArray;
+import com.alibaba.csp.sentinel.slots.statistic.base.LongAdder;
+import com.alibaba.csp.sentinel.slots.statistic.base.WindowWrap;
+import com.alibaba.csp.sentinel.util.function.Tuple2;
 
 /**
  * Provides millisecond-level time of OS.
+ * <p>
+ * Here we should see that not all the time TimeUtil should 
+ * keep looping 1_000 times every second (Actually about 800/s due to some losses).
+ * <pre>
+ * * In idle conditions it just acts as System.currentTimeMillis();
+ * * In busy conditions (significantly more than 1_000/s) it keeps loop to reduce costs.
+ * </pre>
+ * For detail design and proposals please goto 
+ * <a href="https://github.com/alibaba/Sentinel/issues/1702#issuecomment-692151160">https://github.com/alibaba/Sentinel/issues/1702</a>
  *
  * @author qinan.qn
+ * @author jason
  */
-public final class TimeUtil {
+public final class TimeUtil implements Runnable {
+    private static final long CHECK_INTERVAL = 3000;
+    private static final long HITS_LOWER_BOUNDARY = 800;
+    private static final long HITS_UPPER_BOUNDARY = 1200;
+    
+    public static enum STATE {
+        IDLE, PREPARE, RUNNING;
+    }
+    private static class Statistic {
+        private LongAdder writes = new LongAdder();
+        private LongAdder reads = new LongAdder();
+        public LongAdder getWrites() {
+            return writes;
+        }
+        public LongAdder getReads() {
+            return reads;
+        }
+    }
+    private static TimeUtil INSTANCE;
 
-    private static volatile long currentTimeMillis;
+    private volatile long currentTimeMillis;
+    private volatile STATE state = STATE.IDLE;
+    
+    private LeapArray<Statistic> statistics;
+    
+    // thread private variables
+    private long lastCheck = 0;
 
     static {
-        currentTimeMillis = System.currentTimeMillis();
-        Thread daemon = new Thread(new Runnable() {
-            @Override
-            public void run() {
-                while (true) {
-                    currentTimeMillis = System.currentTimeMillis();
-                    try {
-                        TimeUnit.MILLISECONDS.sleep(1);
-                    } catch (Throwable e) {
+        INSTANCE = new TimeUtil();
+    }
+    
+    public TimeUtil() {
+        this.statistics = new LeapArray<TimeUtil.Statistic>(3, 3000) {
 
-                    }
-                }
+            @Override
+            public Statistic newEmptyBucket(long timeMillis) {
+                return new Statistic();
             }
-        });
+
+            @Override
+            protected WindowWrap<Statistic> resetWindowTo(WindowWrap<Statistic> windowWrap, long startTime) {
+                Statistic val = windowWrap.value();
+                val.getReads().reset();
+                val.getWrites().reset();
+                windowWrap.resetTo(startTime);
+                return windowWrap;
+            }
+        };
+        this.currentTimeMillis = System.currentTimeMillis();
+        this.lastCheck = this.currentTimeMillis;
+        Thread daemon = new Thread(this);
         daemon.setDaemon(true);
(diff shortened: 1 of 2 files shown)


jasonjoo2010:
@sczyh30 Why `TimeUtilTest` fails only in ci? Do you know the reason?
It succeeded in local environment. What's the difference?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/TimeUtil.java
```diff
@@ -15,38 +15,199 @@
...
+            return reads;
+        }
+    }
+    private static TimeUtil INSTANCE;
 
-    private static volatile long currentTimeMillis;
+    private volatile long currentTimeMillis;
+    private AtomicLong lastCheck = new AtomicLong();
+    private LeapArray<Statistic> statistics;
+    private STATE state = STATE.IDLE;
```
liqiangz, line 64: I think state should be a volatile variable.

jasonjoo2010: Hi, liqiangz

Here to use modifiers without a `volatile` is because following considerations:

* Low frequency of switching is assumed.
* All writings to it only occurred in same thread (which maybe we should consider to remove the unnecessary CAS operation
* Use L1/L2 cache to get better performance

So should it still be volatile? Is there any other branch I maybe lack of consideration?

liqiangz: Thanks for your reply, but I still have a confusion.  When state in the daemon thread switch to IDLE or PREPARE, currentTimeMillis will not be updated. But when calling currentTime function  in another thread, state may still be RUNNING.  The `getTime()`  result will always be the old currentTimeMillis.

jasonjoo2010: Hi, liqiangz

Both IDLE and PREPARE are treated as IDLE in `getTime()` while daemon will only switch PREPARE to RUNNING when it actually started to update `currentTimeMillis`.

I will remove the CAS operation which is a kind of legacy after several iterations of design. And I also will consider whether it's necessary to make it volatile in java for not strictly reordering forbidden scenario. As we know in rare condition codes like below in c++:

```c++
bool cond = true;
while (cond);
```

may be "optimized" to be 

```c++
if (cond) {
while(true){}
}
```

But I am not sure it will occur in JVM's byte codes. Because it more like a bug which it's common in CPP compilers. (aha)

What do you think about the visibility latency in practice?

jasonjoo2010: CAS operation has been removed and state is taken as volatile now.
In unit test concurrency has been raised more too.

liqiangz: I think we should be very careful about variables shared by multiple threads. If you want to do some concurrency optimization,  you must really understand what you do.

For programs that are not synchronized, if you want to judge the execution order of memory operations, it is almost impossible to get a correct conclusion . A simple way is to synchronize all variables shared by multiple threads.

In this scenario, the value of state modified by one thread may never be seen by another thread, because we don’t know what kind of optimization the compiler and cpu will perform. But because currentTimeMillis is volatile, There may be no problems in this program.

jasonjoo2010: Yeah totally agree with you. So CAS has been removed and volatile is marked on state while lastCheck is still a private and local variable.

linlinisme:
把`TimeUtils`的行为模式和sleep的时长做成配置化会更贴近用户的需求

jasonjoo2010:
> 把`TimeUtils`的行为模式和sleep的时长做成配置化会更贴近用户的需求

这里的考虑是这样的，越多的配置项，看似自由，实则增加了工具的复杂性
所以对于一些参数化的场景，如果可以斟酌出合理的搭配的话（当然这是前提），那就尽量不增加配置项。

奥卡姆剃刀原理，如无必要，勿增实体

linlinisme:
> > 把`TimeUtils`的行为模式和sleep的时长做成配置化会更贴近用户的需求
> 
> 这里的考虑是这样的，越多的配置项，看似自由，实则增加了工具的复杂性
> 所以对于一些参数化的场景，如果可以斟酌出合理的搭配的话（当然这是前提），那就尽量不增加配置项。
> 
> 奥卡姆剃刀原理，如无必要，勿增实体

我的意思是对大部分sentinel使用者来说直接内部走原生的系统调用不是会有性能的问题的。自适应模式对这类用户的收益不大，有时候还会造成困拢。`TimeUtils`的优化是针对部分已确定是超高流量的情况下使用的，这种占比应该还少数。针对`TimeUtils`的sleep的时长，超高流量的用户他们可能会自己测试不同时长下的影响，这类属于高端玩家，都会有自己的想法。

jasonjoo2010:
> > > 把`TimeUtils`的行为模式和sleep的时长做成配置化会更贴近用户的需求
> > 
> > 
> > 这里的考虑是这样的，越多的配置项，看似自由，实则增加了工具的复杂性
> > 所以对于一些参数化的场景，如果可以斟酌出合理的搭配的话（当然这是前提），那就尽量不增加配置项。
> > 奥卡姆剃刀原理，如无必要，勿增实体
> 
> 我的意思是对大部分sentinel使用者来说直接内部走原生的系统调用是不会有性能问题。自适应模式对这类用户的收益不大，有时候还会造成困拢。`TimeUtils`的优化是针对部分已确定是超高流量的情况下使用的，这种占比应该还少数。针对`TimeUtils`的sleep的时长，超高流量的用户他们可能会自己测试不同时长下的影响，这类属于高端玩家，都会有自己的想法。

这个就看怎么评估了，是以接口方式做两种实现变成配置项（类似jvm的gc算法），还是以一个不显著（同一数量级）高于A、B两实现任一实现的方式做统一实现。
这个的调用点还是蛮多的，且属于开放类，使用者业务系统内也是可以使用的，符合“性能扩展工具包”的角色定位。

所以两个方法的确都能实现这个需求。
就个人来讲，我更倾向自动适配，毕竟jvm的gc算法已经令不仅是初学者眼花缭乱了。
这个可以更多的讨论来确定不同人的想法。

sanxun0325 (approved):

JerryChin:
补充一点小意见：使用 `Thread.sleep(ONE_MILLISECOND)` 比 `TimeUnit.MILLISECONDS.sleep(1)` 方法更高效。

PS: `ONE_MILLISECOND = 1`

sczyh30 (approved):
LGTM

sczyh30:
Nice work. This reduces the CPU footprint under low-frequent requests. Thanks for contributing!

JJFly-JOJO:
Local test verification

- 1. Low-traffic situation
thread count = 16 
Call SphU#entry every other 1_000 ms
QPS = 16

Before optimization: CPU usage=60%-80%
After optimization: CPU usage=about 10%

- 2. High-traffic situation
thread count = 16
Call SphU#entry every other 2 ms
QPS = 8000

Before optimization: CPU usage=about 8%
After optimization: CPU usage=about 8%

- 3. High-concurrency situation
thread count = 32
Call SphU#entry every other 10 ms

Before optimization: CPU usage=about 16%
After optimization: CPU usage=about 18%

Summary:
1.this optimization significantly reduces CPU usage int Low-traffic situation
2.the CPU loss of `private LeapArray<Statistic> statistics` is not obvious in High-traffic situation and High-concurrency situation.

Issue #1753, opened 2020-09-23
About AuthorityRuleChecker.The PassCheck() method implementation problem


JiangZian:
Type: *bug report* 

### Describe what happened (or what feature you want)

This class is primarily used to determine the match of the context rules corresponding to the resource.Why do I need to repeat the validation with indexOf() and EQUALS methods?

### Anything else we need to know?

com.alibaba.csp.sentinel.slots.block.authority.AuthorityRuleChecker

JiangZian:
![image](https://user-images.githubusercontent.com/24679521/93953712-aee26280-fd7e-11ea-9c1e-2c7369b472db.png)
****

chaudharydeepak:
@sczyh30 @JiangZian - let me know if you want me to refactor this, please assign to me. Thank you.

JiangZian:
> @sczyh30 @JiangZian - let me know if you want me to refactor this, please assign to me. Thank you.

Do you have any good Suggestions for improvement?

Jayice-zjw:
for example
"abcd,efg,hijk".indexOf("cd");

PR #1754, opened 2020-09-23
About AuthorityRuleChecker.The PassCheck() method implementation problem

JiangZian:
#1753 

Logic is implemented in a better way

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/authority/AuthorityRuleChecker.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/authority/AuthorityRuleChecker.java
@@ -29,39 +29,39 @@ final class AuthorityRuleChecker {
 
     static boolean passCheck(AuthorityRule rule, Context context) {
         String requester = context.getOrigin();
+        boolean contain = false;
 
-        // Empty origin or empty limitApp will pass.
-        if (StringUtil.isEmpty(requester) || StringUtil.isEmpty(rule.getLimitApp())) {
-            return true;
+        {
+            // Empty origin or empty limitApp will pass.
+            if (StringUtil.isEmpty(requester) || StringUtil.isEmpty(rule.getLimitApp())) {
+                return true;
+            }
         }
 
-        // Do exact match with origin name.
-        int pos = rule.getLimitApp().indexOf(requester);
-        boolean contain = pos > -1;
-
-        if (contain) {
-            boolean exactlyMatch = false;
+        {
+            // Do exact match with origin name.
             String[] appArray = rule.getLimitApp().split(",");
             for (String app : appArray) {
                 if (requester.equals(app)) {
-                    exactlyMatch = true;
+                    contain = true;
                     break;
                 }
             }
-
-            contain = exactlyMatch;
         }
 
-        int strategy = rule.getStrategy();
-        if (strategy == RuleConstant.AUTHORITY_BLACK && contain) {
-            return false;
+        {
+            //The matching result is judged according to the Strategy
+            switch (rule.getStrategy()) {
+                case RuleConstant.AUTHORITY_BLACK:
+                    return !contain;
+                case RuleConstant.AUTHORITY_WHITE:
+                    return contain;
+                default:
+                    return true;
+
+            }
         }
 
-        if (strategy == RuleConstant.AUTHORITY_WHITE && !contain) {
-            return false;
-        }
-
-        return true;
     }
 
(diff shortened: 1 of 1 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/authority/AuthorityRuleChecker.java
```diff
@@ -29,39 +29,39 @@
...
 
-        // Empty origin or empty limitApp will pass.
-        if (StringUtil.isEmpty(requester) || StringUtil.isEmpty(rule.getLimitApp())) {
-            return true;
+        {
+            // Empty origin or empty limitApp will pass.
+            if (StringUtil.isEmpty(requester) || StringUtil.isEmpty(rule.getLimitApp())) {
+                return true;
+            }
         }
```
liqiangz, line 39: Why are there extra { } here

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/authority/AuthorityRuleChecker.java
```diff
@@ -29,39 +29,39 @@
...
 
-        // Do exact match with origin name.
-        int pos = rule.getLimitApp().indexOf(requester);
```
liqiangz, line 39: I think this double check is a performance optimization, I will run a performance test later to see the performance of 'indexOf` and loop.

liqiangz: I have test it, The performance of indexOf is nearly 10 times that of loop traversal.  I think should remain the double check because In most cases the requester may not be in the black or white list and will be blocked by the first check

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/authority/AuthorityRuleChecker.java
JiangZian: This piece of logic is distinguished by a code block, which makes it feel clearer

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/authority/AuthorityRuleChecker.java
JiangZian: Thank you for your correction.I have tested the problem you mentioned, the extra consumption is around 7 times, but it is still very small (600 milliseconds more than 10 million tests).
But through an analysis of the problem you provided, I came up with another bug, which you can see in #1760

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/authority/AuthorityRuleChecker.java
liqiangz: Thank you for your reply.But I didn't see this code style elsewhere in this project. I think we should unify the code style.

JiangZian: What you said is quite reasonable. I have removed this part of the logic

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/authority/AuthorityRuleChecker.java
```diff
@@ -29,40 +29,28 @@
...
 
         // Empty origin or empty limitApp will pass.
         if (StringUtil.isEmpty(requester) || StringUtil.isEmpty(rule.getLimitApp())) {
             return true;
         }
 
         // Do exact match with origin name.
-        int pos = rule.getLimitApp().indexOf(requester);
-        boolean contain = pos > -1;
+        contain = rule.getLimitApp().contains(requester);
```
liqiangz, line 40: If limitApp is `foobar,test`, and requester is ‘foo’， ‘foo’ will be blocked mistakely.  So we need remain the double check.  In most cases the requester will be passed because the first check. The loop second check confirms that requester never be blocked mistakely.

JiangZian: Oh, I really didn't take that into consideration. Thank you for your correction，
Now That I've fixed it, the performance is really down, and after 100 million calls, it's three times faster

LearningGp:

PR #1779, opened 2020-10-05
[Fixes #1769] adjust decision logic when slowRequestRatio equals 1.0

### LINKED ISSUE #1769: When slowRatioThreshold equals 1.0，requests will never be blocked
在慢调用比例策略的降级规则中，把比例阈值设置为1，我发现所有的请求都不会被降级了。
源码是这样写的
if (currentRatio > maxSlowRequestRatio) {
            transformToOpen(currentRatio);
        }
我不太清楚这属不属于一个bug，因为我自己的理解是阈值设置为1时，只要在一个窗口内的所有请求都超过了最大RT，那么资源就会被降级。
不知是我的理解有误还是确实阈值为1时不符合预期。

xierunzi:
#1769

when slowRequestRatio equals 1.0d, requests will never be blocked. Because the former decision logic is use '>' to block the request, but the max ratio is 100%, so the currentRatio will never be bigger than slowRequestRatio.

Fixes #1769 

### Describe how you did it
If maxSlowRequestRatio equals currentRatio && maxSlowRequestRatio equals 1.0d, call this method 'transformToOpen'.

### Describe how to verify it
In the Test Class which called ResponseTimeCircuitBreakerTest.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/ResponseTimeCircuitBreaker.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/degrade/circuitbreaker/ResponseTimeCircuitBreaker.java
@@ -19,7 +19,6 @@ import java.util.List;
 
 import com.alibaba.csp.sentinel.Entry;
 import com.alibaba.csp.sentinel.context.Context;
-import com.alibaba.csp.sentinel.slotchain.ResourceWrapper;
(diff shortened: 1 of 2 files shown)


sczyh30 (approved):
LGTM

sczyh30:
Besides, there is the same problem in error-ratio mode. Would you like to fix it as well?

xierunzi:
> Thanks for contributing and look forward to more!
> Besides, there is the same problem in error-ratio mode. Would you like to fix it as well?

sure.

PR #1783, opened 2020-10-06
fix the potential concurrence issue when rules updates

### LINKED ISSUE #1782: Potential concurrence issue happens when updating rules
## Issue Description

Type: *bug report*

When you update rules, XXXRuleManager.XXXPropertyListener.configUpdate will be called, and it's not atomic action of calling clear() and putAll(). That could make it no rule in an instant.

Whenever you call XXXRuleManager.getRules(), the rules should be initial ones or the updated ones, there should never be empty.

1. create a UT class for FlowRuleManager which contains a loader thread which is loading FlowRules again and again.
2. in the man thread, just do getRules again and again.
3. if the size happens to 0, there must be the issue.

### Tell us your environment
None of env's business

I'll submit a UT later to prove it

vipweihua:
fix the potential concurrence issue when rules updates

Fixes #1782.

### Describe how you did it
make the map wrapped in AtomicReference

### Describe how to verify it
Run the new-added UT 

### Special notes for reviews
If this pr can be acceptable, I plan to fix the same issue in other rule managers. 
Code seems to be strange in DegradeRuleManager because the same piece of code uses synchronized keyword. It seems to be realized, but in the wrong way to fix it.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
@@ -44,6 +44,7 @@ import com.alibaba.csp.sentinel.property.SentinelProperty;
  *
  * @author jialiang.linjl
  * @author Eric Zhao
+ * @author Weihua
  */
 public class FlowRuleManager {
 
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
@@ -16,6 +16,7 @@
 package com.alibaba.csp.sentinel.slots.block.flow;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -58,6 +59,7 @@ public class FlowRuleManager {
         new NamedThreadFactory("sentinel-metrics-record-task", true));
 
     static {
+        flowRules.set(Collections.<String, List<FlowRule>>emptyMap());
         currentProperty.addListener(LISTENER);
         SCHEDULER.scheduleAtFixedRate(new MetricTimerListener(), 0, 1, TimeUnit.SECONDS);
     }
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
@@ -133,14 +133,14 @@ public class FlowRuleManager {
             //the rules was always not null, it's no need to check nullable
             //remove checking to avoid IDE warning
             flowRules.set(rules);
-            RecordLog.info("[FlowRuleManager] Flow rules received: {}", flowRules);
+            RecordLog.info("[FlowRuleManager] Flow rules received: {}", flowRules.get());
         }
 
         @Override
         public void configLoad(List<FlowRule> conf) {
             Map<String, List<FlowRule>> rules = FlowRuleUtil.buildFlowRuleMap(conf);
             flowRules.set(rules);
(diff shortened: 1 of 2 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
```diff
@@ -129,20 +130,16 @@ public static boolean isOtherOrigin(String origin, String resourceName) {
...
         public void configUpdate(List<FlowRule> value) {
             Map<String, List<FlowRule>> rules = FlowRuleUtil.buildFlowRuleMap(value);
-            if (rules != null) {
-                flowRules.clear();
-                flowRules.putAll(rules);
-            }
+            //the rules was always not null, it's no need to check nullable
+            //remove checking to avoid IDE warning
+            flowRules.set(rules);
             RecordLog.info("[FlowRuleManager] Flow rules received: {}", flowRules);
```
jasonjoo2010, line 136: Maybe the variable in logging message should be updated now.

jasonjoo2010 (changes requested):

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
vipweihua: fixed

vipweihua:
@jasonjoo2010

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
```diff
@@ -83,7 +85,7 @@ public static void register2Property(SentinelProperty<List<FlowRule>> property)
      */
     public static List<FlowRule> getRules() {
         List<FlowRule> rules = new ArrayList<FlowRule>();
-        for (Map.Entry<String, List<FlowRule>> entry : flowRules.entrySet()) {
+        for (Map.Entry<String, List<FlowRule>> entry : flowRules.get().entrySet()) {
```
jasonjoo2010, line 88: Is it necessary to initial its value (flowRules) to be an empty map obviously?
It's tricky now because the initial empty value is initialized by `currentProperty.addListener(LISTENER);` in static block. It may trigger a bug if anyone change any of them in future.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
```diff
@@ -129,21 +131,17 @@ public static boolean isOtherOrigin(String origin, String resourceName) {
...
             Map<String, List<FlowRule>> rules = FlowRuleUtil.buildFlowRuleMap(value);
-            if (rules != null) {
-                flowRules.clear();
-                flowRules.putAll(rules);
-            }
-            RecordLog.info("[FlowRuleManager] Flow rules received: {}", flowRules);
+            //the rules was always not null, it's no need to check nullable
+            //remove checking to avoid IDE warning
+            flowRules.set(rules);
+            RecordLog.info("[FlowRuleManager] Flow rules received: {}", flowRules.get());
```
jasonjoo2010, line 137: For eventual consistency maybe `rules` is better. Because there may be duplicated logs with same content when races occurred.

jasonjoo2010 (changes requested):
Hi, wei

I have dropped new comments which I missed first and please have a look when you had time.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
vipweihua: Totally agree.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
vipweihua: Agree.

jasonjoo2010 (changes requested):
LGTM

jasonjoo2010 (approved):
LGTM

jasonjoo2010:
Thanks for contributing!

PR #1784, opened 2020-10-08
fix concurrence issue for AuthorityRuleManager

### LINKED ISSUE #1782: Potential concurrence issue happens when updating rules

vipweihua:
fix concurrence issue for AuthorityRuleManager

Fixes #1782.

Describe how you did it
make the map wrapped in AtomicReference

Describe how to verify it
Run the new-added UT

Special notes for reviews
This is 2nd pr and for AuthorityRuleManager

### DIFF
--- a/sentinel-core/src/test/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManagerTest.java
+++ b/sentinel-core/src/test/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManagerTest.java
@@ -57,7 +57,7 @@ public class FlowRuleManagerTest {
         }
     }
 
-    public Runnable loader = new Runnable() {
+    private static Runnable loader = new Runnable() {
         @Override
         public void run() {
             for(int i = 0; i < 10000; i++){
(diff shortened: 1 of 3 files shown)


vipweihua:
@sczyh30 @jasonjoo2010 Is that OK? Is there anything I need to do?

sentinel-core/src/test/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManagerTest.java
```diff
@@ -57,7 +57,7 @@ public void testLoadAndGetRules(){
         }
     }
 
-    public Runnable loader = new Runnable() {
+    private static Runnable loader = new Runnable() {
```
vipweihua, line 60: It may be conflict with pr#1823, i'll sync them firstly after 1823 is merged

linlinisme:
this is a thread safe issue.  i would like to use `synchronized` to fix it,  at the meanwhile still keep `ConcurrentHashMap`  high performance advantage
```
 synchronized (authorityRules) {
                authorityRules.clear();
                if (rules != null) {
                    authorityRules.putAll(rules);
                }
            }
```

PR #1844, opened 2020-11-10
feat: support cpu usage calculate by single core

fudali113:

when run app in docker or k8s,
if set --cpu-period=100000 & --cpu-quota=100001 on docker run command(or set cpu request limit 1001m on k8s),
{@link OperatingSystemMXBean#getAvailableProcessors()} will return 2;
in this case, currentCpuUsage does not accurately express the percentage of available CPU resources;

### Describe how you did it

Set a parameter so that currentCpuUsage is always calculated against a single core;

### Special notes for reviews


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemStatusListener.java
```diff
@@ -27,9 +27,25 @@
 
 /**
  * @author jialiang.linjl
+ * @author <a href="mailto:doob.yi@qq.com">doob</a>
  */
 public class SystemStatusListener implements Runnable {
 
+    public static final String CALCULATE_BY_SINGLE_CORE_KEY =
+            "csp.sentinel.system.status.cpuUsageCalculateBySingleCore";
```
jasonjoo2010, line 35: I suggest that we could do a lazy initialization in, for example, `run()`?
It will be more friendly for various implementations.

jasonjoo2010 (changes requested):

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/SystemStatusListener.java
fudali113: > I suggest that we could do a lazy initialization in, for example, `run()`?
> It will be more friendly for various implementations.

good idea

fudali113: > I suggest that we could do a lazy initialization in, for example, `run()`?
> It will be more friendly for various implementations.

fixed;

i lazy init in `getProcessCpuUsage()`，because i don't think `run()` should care about this parameter

jasonjoo2010: > i lazy init in `getProcessCpuUsage()`，because i don't think `run()` should care about this parameter

Though it hasn't been synchronized here I think it's not a big deal.

jasonjoo2010:
And @sczyh30 do you have any other approach calculates CPU costs in a better way?
Does this approach only work with single-core bound?

jasonjoo2010:
... and another idea is that we introduce an `INTEGER` instead of a `BOOLEAN` configuration entry indicating the cores. And it can override the value from MXBean.

fudali113:
> ... and another idea is that we introduce an `INTEGER` instead of a `BOOLEAN` configuration entry indicating the cores. And it can override the value from MXBean.

or got `cpu.cfs_period_us & cpu.cfs_quota_us` calculator a double type cores value?

jasonjoo2010:
> > ... and another idea is that we introduce an `INTEGER` instead of a `BOOLEAN` configuration entry indicating the cores. And it can override the value from MXBean.
> 
> or got `cpu.cfs_period_us & cpu.cfs_quota_us` calculator a double type cores value?

Not quite sure. Because there are quite a bit parameters can do various limitation like cpuset_*, cpu_*, cpus, etc.

PR #1896, opened 2020-12-14
Support HTTPS url for -Dcsp.sentinel.dashboard.server attribute

### LINKED ISSUE #1840: Support HTTPS url for -Dcsp.sentinel.dashboard.server attribute
## Issue Description

*bug report*

当配置-Dcsp.sentinel.dashboard.server为指定地址，但这个地址会强制跳转https（443端口）时，会连接不到，对端收不到心跳包，请确认下

请确认下能否使用https地址

polarbear567:
Fixes #1840 

### Describe how you did it
1. The involved HTTPS request is only heartbeat transmission,  so this PR takes the operation of trusting all certificates
2. Because the dashboard does not have a certificate configured by default, this modification does not involve the dashboard. The dashboard is a spring-boot project. If you need to use HTTPS, you need to configure it yourself

### Describe how to verify it
First we verify the https protocol:
1. Add ssl config in dashboard application.properties, like this one:

> server.ssl.key-store=xxx/yyy.keystore
> server.ssl.key-store-password=zzz
> server.ssl.keyStoreType=JKS
> server.ssl.keyAlias=aaa

I used **keytool** to generate certificates
2. Start a project that uses the modified jar with `-Dcsp.sentinel.dashboard.server=https://127.0.0.1:8080`
3. Start the dashboard jar
4. Use browser access https://127.0.0.1:8080, we can see that the dashboard has been able to receive the data of demo
Second, we verify the http protocol
1.  Start a project that uses the modified jar with `-Dcsp.sentinel.dashboard.server=http://127.0.0.1:8080` or `-Dcsp.sentinel.dashboard.server=127.0.0.1:8080`
2. Start the dashboard jar without ssl config
3. Use browser access http://127.0.0.1:8080, also we can see the client.

### Special notes for reviews
1.Is it necessary to set a certificate for the heartbeat of https?
2.Whether it is necessary to set parameters for the dashboard so that the https protocol can be used (because users may directly use the jar package instead of compiling by themselves)

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/function/Endpoint.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/function/Endpoint.java
@@ -1,0 +1,53 @@
+package com.alibaba.csp.sentinel.util.function;
+
+import java.net.InetSocketAddress;
+
+/**
+ * @author Leo Li
+ */
+public class Endpoint {
+    private Protocol protocol;
+
+    private String host;
+
+    private int port;
+
+    public Endpoint(Protocol protocol, String host, int port) {
+        this.protocol = protocol;
+        this.host = host;
+        this.port = port;
+    }
+
+    public Protocol getProtocol() {
+        return protocol;
+    }
+
+    public void setProtocol(Protocol protocol) {
+        this.protocol = protocol;
+    }
+
+    public String getHost() {
+        return host;
+    }
+
+    public void setHost(String host) {
+        this.host = host;
+    }
+
+    public int getPort() {
+        return port;
+    }
+
+    public void setPort(int port) {
+        this.port = port;
+    }
+
+    public InetSocketAddress getInetSocketAddress() {
+        return new InetSocketAddress(host, port);
+    }
+
+    @Override
+    public String toString() {
+        return "Endpoint{" + "protocol=" + protocol + ", host='" + host + ", port=" + port + '}';
+    }
+}
--- a/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
+++ b/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
@@ -19,7 +19,8 @@ import com.alibaba.csp.sentinel.config.SentinelConfig;
 import com.alibaba.csp.sentinel.log.RecordLog;
 import com.alibaba.csp.sentinel.util.HostNameUtil;
 import com.alibaba.csp.sentinel.util.StringUtil;
-import com.alibaba.csp.sentinel.util.function.Tuple2;
+import com.alibaba.csp.sentinel.util.function.Endpoint;
+import com.alibaba.csp.sentinel.util.function.Protocol;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -41,8 +42,6 @@ public class TransportConfig {
 
     private static int runtimePort = -1;
 
-	private static String protocol = "http";
-
     /**
      * Get heartbeat interval in milliseconds.
      *
@@ -59,20 +58,20 @@ public class TransportConfig {
     }
 
     /**
-     * Get a list of (ip/domain, port) indicating Sentinel Dashboard's address.<br>
-     * NOTE: only support <b>HTTP</b> protocol
+     * Get a list of Endpoint(protocol, ip/domain, port) indicating Sentinel Dashboard's address.<br>
+     * NOTE: only support <b>HTTP</b> and <b>HTTPS</b> protocol
      *
-     * @return list of (ip/domain, port) pair. <br>
+     * @return list of Endpoint(protocol, ip/domain, port). <br>
      *         <b>May not be null</b>. <br>
      *         An empty list returned when not configured.
      */
-    public static List<Tuple2<String, Integer>> getConsoleServerList() {
+    public static List<Endpoint> getConsoleServerList() {
         String config = SentinelConfig.getConfig(CONSOLE_SERVER);
-        List<Tuple2<String, Integer>> list = new ArrayList<Tuple2<String, Integer>>();
+        List<Endpoint> list = new ArrayList<Endpoint>();
         if (StringUtil.isBlank(config)) {
             return list;
         }
-        
+
         int pos = -1;
         int cur = 0;
         while (true) {
@@ -95,17 +94,16 @@ public class TransportConfig {
                 continue;
             }
             ipPortStr = ipPortStr.trim();
-			int port = 80;
-			if (ipPortStr.startsWith("http://")) {
-				ipPortStr = ipPortStr.substring(7);
-				port = 80;
-			}
-			if (ipPortStr.startsWith("https://")) {
-				ipPortStr = ipPortStr.substring(8);
-				port = 443;
-				protocol = "https";
-			}
-			int index = ipPortStr.indexOf(":");
+            int port = 80;
+            Protocol protocol = Protocol.HTTP;
+            if (ipPortStr.startsWith("http://")) {
+                ipPortStr = ipPortStr.substring(7);
+            } else if (ipPortStr.startsWith("https://")) {
+                ipPortStr = ipPortStr.substring(8);
+                port = 443;
+                protocol = Protocol.HTTPS;
+            }
+            int index = ipPortStr.indexOf(":");
             if (index == 0) {
                 // skip
                 continue;
@@ -124,7 +122,7 @@ public class TransportConfig {
                 }
                 host = ipPortStr.substring(0, index);
             }
-            list.add(Tuple2.of(host, port));
+            list.add(new Endpoint(protocol, host, port));
         }
         return list;
     }
@@ -132,10 +130,6 @@ public class TransportConfig {
     public static int getRuntimePort() {
         return runtimePort;
     }
-
-	public static String getProtocol() {
-		return protocol;
-	}
 
     /**
      * Get Server port of this HTTP server.
--- a/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
+++ b/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
@@ -27,6 +27,7 @@ import java.util.List;
 /**
  * @author Carpenter Lee
  * @author Jason Joo
+ * @author Leo Li
  */
 public class TransportConfig {
 
@@ -39,6 +40,8 @@ public class TransportConfig {
     public static final String HEARTBEAT_DEFAULT_PATH = "/registry/machine";
 
     private static int runtimePort = -1;
+
+	private static String protocol = "http";
 
     /**
      * Get heartbeat interval in milliseconds.
@@ -92,11 +95,17 @@ public class TransportConfig {
                 continue;
             }
             ipPortStr = ipPortStr.trim();
-            if (ipPortStr.startsWith("http://")) {
-                ipPortStr = ipPortStr.substring(7);
-            }
-            int index = ipPortStr.indexOf(":");
-            int port = 80;
+			int port = 80;
+			if (ipPortStr.startsWith("http://")) {
+				ipPortStr = ipPortStr.substring(7);
+				port = 80;
+			}
+			if (ipPortStr.startsWith("https://")) {
+				ipPortStr = ipPortStr.substring(8);
+				port = 443;
+				protocol = "https";
+			}
+			int index = ipPortStr.indexOf(":");
             if (index == 0) {
                 // skip
                 continue;
@@ -123,6 +132,10 @@ public class TransportConfig {
     public static int getRuntimePort() {
         return runtimePort;
     }
+
+	public static String getProtocol() {
+		return protocol;
+	}
 
     /**
      * Get Server port of this HTTP server.
--- a/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
+++ b/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
@@ -19,8 +19,8 @@ import com.alibaba.csp.sentinel.config.SentinelConfig;
 import com.alibaba.csp.sentinel.log.RecordLog;
 import com.alibaba.csp.sentinel.util.HostNameUtil;
 import com.alibaba.csp.sentinel.util.StringUtil;
-import com.alibaba.csp.sentinel.util.function.Endpoint;
-import com.alibaba.csp.sentinel.util.function.Protocol;
+import com.alibaba.csp.sentinel.transport.endpoint.Endpoint;
+import com.alibaba.csp.sentinel.transport.endpoint.Protocol;
 
 import java.util.ArrayList;
 import java.util.List;
--- a/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/endpoint/Protocol.java
+++ b/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/endpoint/Protocol.java
@@ -1,0 +1,23 @@
+package com.alibaba.csp.sentinel.transport.endpoint;
+
+/**
+ * @author Leo Li
+ */
+public enum Protocol {
+    HTTP("http"),
+    HTTPS("https");
+
+    private String protocol;
+
+    Protocol(String protocol) {
+        this.protocol = protocol;
+    }
+
+    public String getProtocol() {
+        return protocol;
+    }
+
+    public static void main(String[] args) {
+        System.out.println(Protocol.HTTP.name());
+    }
+}
--- a/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/endpoint/Protocol.java
+++ b/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/endpoint/Protocol.java
@@ -16,8 +16,4 @@ public enum Protocol {
     public String getProtocol() {
         return protocol;
     }
-
-    public static void main(String[] args) {
-        System.out.println(Protocol.HTTP.name());
-    }
 }
--- a/sentinel-transport/sentinel-transport-netty-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/HttpHeartbeatSender.java
+++ b/sentinel-transport/sentinel-transport-netty-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/HttpHeartbeatSender.java
@@ -25,7 +25,7 @@ import com.alibaba.csp.sentinel.util.AppNameUtil;
 import com.alibaba.csp.sentinel.util.HostNameUtil;
 import com.alibaba.csp.sentinel.util.PidUtil;
 import com.alibaba.csp.sentinel.util.StringUtil;
-import com.alibaba.csp.sentinel.util.function.Tuple2;
+import com.alibaba.csp.sentinel.util.function.Endpoint;
 
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
@@ -59,14 +59,14 @@ public class HttpHeartbeatSender implements HeartbeatSender {
 
     public HttpHeartbeatSender() {
         this.client = HttpClients.createDefault();
-        List<Tuple2<String, Integer>> dashboardList = TransportConfig.getConsoleServerList();
+        List<Endpoint> dashboardList = TransportConfig.getConsoleServerList();
         if (dashboardList == null || dashboardList.isEmpty()) {
             RecordLog.info("[NettyHttpHeartbeatSender] No dashboard server available");
             consoleHost = null;
             consolePort = -1;
         } else {
-            consoleHost = dashboardList.get(0).r1;
-            consolePort = dashboardList.get(0).r2;
+            consoleHost = dashboardList.get(0).getHost();
+            consolePort = dashboardList.get(0).getPort();
             RecordLog.info("[NettyHttpHeartbeatSender] Dashboard address parsed: <{}:{}>", consoleHost, consolePort);
         }
     }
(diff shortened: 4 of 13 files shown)


sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
```diff
@@ -92,11 +95,17 @@ public static Long getHeartbeatIntervalMs() {
...
-                ipPortStr = ipPortStr.substring(7);
-            }
-            int index = ipPortStr.indexOf(":");
-            int port = 80;
+			int port = 80;
+			if (ipPortStr.startsWith("http://")) {
+				ipPortStr = ipPortStr.substring(7);
+				port = 80;
+			}
+			if (ipPortStr.startsWith("https://")) {
```
jasonjoo2010, line 103: Here an "ELSE" should be introduced as a different condition branch.

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/SimpleHttpHeartbeatSender.java
```diff
@@ -69,7 +70,7 @@ public boolean sendHeartbeat() throws Exception {
         }
 
         InetSocketAddress addr = new InetSocketAddress(addrInfo.r1, addrInfo.r2);
-        SimpleHttpRequest request = new SimpleHttpRequest(addr, TransportConfig.getHeartbeatApiPath());
```
jasonjoo2010, line 72: If a new type was introduced it's highly recommended to pass it into query implementations directly.

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/SimpleHttpClient.java
```diff
@@ -77,57 +90,73 @@ public SimpleHttpResponse post(SimpleHttpRequest request) throws IOException {
...
+		BufferedWriter writer;
+		try {
+			if (StringUtil.equals(protocol, "http")) {
+				socket = new Socket();
+			} else if (StringUtil.equals(protocol, "https")){
+				sslSocketFactory = getSSLSocketFactory();
+				if (sslSocketFactory != null) {
+					socket = sslSocketFactory.createSocket();
+				} else {
+					socket = new Socket();
```
jasonjoo2010, line 113: Maybe it doesn't make any sense here?

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/SimpleHttpClient.java
```diff
@@ -184,6 +213,50 @@ private String encodeRequestParams(Map<String, String> paramsMap, Charset charse
...
+				@Override
+				public X509Certificate[] getAcceptedIssuers() {
+					return null;
+				}
+			};
+		}
+		return x509TrustManager;
+	}
+
+	private SSLSocketFactory getSSLSocketFactory() {
```
jasonjoo2010, line 246: The All-In-One approach makes the class a little ambiguous.
Suggest abstracting the socket creation into different implementations outside.

jasonjoo2010 (changes requested):

sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
```diff
@@ -92,11 +95,17 @@ public static Long getHeartbeatIntervalMs() {
                 continue;
             }
             ipPortStr = ipPortStr.trim();
-            if (ipPortStr.startsWith("http://")) {
-                ipPortStr = ipPortStr.substring(7);
-            }
-            int index = ipPortStr.indexOf(":");
-            int port = 80;
+			int port = 80;
```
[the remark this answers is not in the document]
polarbear567: It's strange, it looks good in my IDE.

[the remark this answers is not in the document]
jasonjoo2010: > It's strange, it looks good in my IDE. Let me check.

Yes, it depends on your editor's preference settings.
Maybe you have the setting using TAB instead of SPACE and setting the width of the TAB to 4 spaces' width.
Sentinel and some other projects use the style of plain spaces with an indent of 4 spaces.

sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
```diff
@@ -40,6 +41,8 @@
 
     private static int runtimePort = -1;
 
+	private static String protocol = "http";
```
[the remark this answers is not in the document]
polarbear567: > I suggest using an enum type to make it easy to compare.
> And the protocol should not be a global state but be a kind of state in the addr-port pairs.

In fact, I've thought about creating a new class similar to Tuple2, but it can have three elements. Do you mean it's still stored in Tuple2 in the form of Tuple2.of(protocol + host, port)?

sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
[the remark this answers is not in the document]
polarbear567: > > It's strange, it looks good in my IDE. Let me check.
> 
> Yes, it depends on your editor's preference settings.
> Maybe you have the setting using TAB instead of SPACE and setting the width of the TAB to 4 spaces' width.
> Sentinel and some other projects use the style of plain spaces with an indent of 4 spaces.

Great!

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/SimpleHttpHeartbeatSender.java
polarbear567: I'm sorry, I didn't get your point here

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/SimpleHttpClient.java
polarbear567: > Maybe it doesn't make any sense here?

Yes, actually, it doesn't work here. It's just used to prevent NPE, but it won't happen in theory. I'll delete it.

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/SimpleHttpHeartbeatSender.java
polarbear567: > If a new type was introduced it's highly recommended to pass it into query implementations directly.

I'm sorry, I didn't get your point here. Could you give me a hint?

sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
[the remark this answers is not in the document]
jasonjoo2010: > > I suggest using an enum type to make it easy to compare.
> > And the protocol should not be a global state but be a kind of state in the addr-port pairs.
> 
> In fact, I've thought about creating a new class similar to Tuple2, but it can have three elements. Do you mean it's still stored in Tuple2 in the form of Tuple2.of(protocol + host, port)?

Elements grow a new type like Endpoint can be introduced so that the Tuple can be replaced. A Tuple3 type maybe looks more complicated now.

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/SimpleHttpHeartbeatSender.java
jasonjoo2010: > > If a new type was introduced it's highly recommended to pass it into query implementations directly.
> 
> I'm sorry, I didn't get your point here. Could you give me a hint?

It's related to the previous design (Tuple3/class Endpoint{}). If a new class was introduced there's no need to split it into three separated arguments I think.

sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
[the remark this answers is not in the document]
polarbear567: > > > I suggest using an enum type to make it easy to compare.
> > > And the protocol should not be a global state but be a kind of state in the addr-port pairs.
> > 
> > 
> > In fact, I've thought about creating a new class similar to Tuple2, but it can have three elements. Do you mean it's still stored in Tuple2 in the form of Tuple2.of(protocol + host, port)?
> 
> Elements grow a new type like Endpoint can be introduced so that the Tuple can be replaced. A Tuple3 type maybe looks more complicated now.

As far as I know, Java can't implement tuples of different lengths in a class unless it's identified by length in a class. I don't know if it's right. If you agree, I'll implement a simple tuple3 as a temporary solution.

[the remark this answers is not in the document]
jasonjoo2010: > > > > I suggest using an enum type to make it easy to compare.
> > > > And the protocol should not be a global state but be a kind of state in the addr-port pairs.
> > > 
> > > 
> > > In fact, I've thought about creating a new class similar to Tuple2, but it can have three elements. Do you mean it's still stored in Tuple2 in the form of Tuple2.of(protocol + host, port)?
> > 
> > 
> > Elements grow a new type like Endpoint can be introduced so that the Tuple can be replaced. A Tuple3 type maybe looks more complicated now.
> 
> As far as I know, Java can't implement tuples of different lengths in a class unless it's identified by length in a class. I don't know if it's right. If you agree, I'll implement a simple tuple3 as a temporary solution.

To be clear, I would show it by examples to make sure we are on the same road:

The old approach:

Tuple2<addr, port>   -> fixed to HTTP protocol

1st approach to improve it:

Tuple3<protocol, addr, port> -> A new type Tuple3 is introduced.

2nd approach:

Endpoint -> A new type Endpoint is introduced as the details below:

```java
public class Endpoint {
    private PROTOCOL protocol;
    private String address;
    private int port;
    // getters and setters
}
```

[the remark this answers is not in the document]
polarbear567: > > > > > I suggest using an enum type to make it easy to compare.
> > > > > And the protocol should not be a global state but be a kind of state in the addr-port pairs.
> > > > 
> > > > 
> > > > In fact, I've thought about creating a new class similar to Tuple2, but it can have three elements. Do you mean it's still stored in Tuple2 in the form of Tuple2.of(protocol + host, port)?
> > > 
> > > 
> > > Elements grow a new type like Endpoint can be introduced so that the Tuple can be replaced. A Tuple3 type maybe looks more complicated now.
> > 
> > 
> > As far as I know, Java can't implement tuples of different lengths in a class unless it's identified by length in a class. I don't know if it's right. If you agree, I'll implement a simple tuple3 as a temporary solution.
> 
> To be clear, I would show it by examples to make sure we are on the same road:
> 
> The old approach:
> 
> Tuple2<addr, port> -> fixed to HTTP protocol
> 
> 1st approach to improve it:
> 
> Tuple3<protocol, addr, port> -> A new type Tuple3 is introduced.
> 
> 2nd approach:
> 
> Endpoint -> A new type Endpoint is introduced as the details below:
> 
> ```java
> public class Endpoint {
>     private PROTOCOL protocol;
>     private String address;
>     private int port;
>     // getters and setters
> }
> ```

I prefer the second approach.

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/SimpleHttpHeartbeatSender.java
polarbear567: > > > If a new type was introduced it's highly recommended to pass it into query implementations directly.
> > 
> > 
> > I'm sorry, I didn't get your point here. Could you give me a hint?
> 
> It's related to the previous design (Tuple3/class Endpoint{}). If a new class was introduced there's no need to split it into three separated arguments I think.

After using Endpoint, I will modify it here and directly pass in Endpoint in the constructor of SimpleHttpRequest to replace protocol and socketAddress

sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/config/TransportConfig.java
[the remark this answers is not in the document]
jasonjoo2010: > 
> I prefer the second approach.

Yes, I think we have reached the same point now. And another suggestion on the design of arguments of different implementations is based on the decision here.
And I totally agree with you here.😀

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/SimpleHttpHeartbeatSender.java
jasonjoo2010: > 
> After using Endpoint, I will modify it here and directly pass in Endpoint in the constructor of SimpleHttpRequest to replace protocol and socketAddress

Yes, that's it. 
Meanwhile, it won't be a backward compatible change and we don't have the perfect one exactly.

polarbear567:
Hi @jasonjoo2010 ,
Please review again, and tell me if you have any suggestions.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/function/Endpoint.java
```diff
@@ -0,0 +1,53 @@
+package com.alibaba.csp.sentinel.util.function;
```
jasonjoo2010, line 1: 😅Is it better to put them in transport-common instead of the core package?
Or are there other pieces of code outside the transport module depending on them?

jasonjoo2010: And please `function` makes little sense here. I believe you really can figure out a good naming💪

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/function/Endpoint.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public int getPort() {
+        return port;
+    }
+
+    public void setPort(int port) {
+        this.port = port;
+    }
+
+    public InetSocketAddress getInetSocketAddress() {
```
jasonjoo2010, line 45: It's strange to put such a piece of code here but I won't consist on it. It looks like no harm.

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/SocketFactory.java
```diff
@@ -0,0 +1,62 @@
...
+            sslContext = SSLContext.getInstance("TLS");
+            sslContext.init(null, new TrustManager[] { x509TrustManager }, null);
+            sslSocketFactory = sslContext.getSocketFactory();
+        } catch (Exception e) {
+            RecordLog.error("get ssl socket factory error", e);
+        }
+    }
+
+    public static Socket getSocket(Protocol protocol) throws IOException {
+        return protocol == Protocol.HTTP ? new Socket() : sslSocketFactory.createSocket();
```
jasonjoo2010, line 60: There is a small tweak here. 
It would be better to make `sslSocketFactory` initialized only if necessary.

sentinel-transport/sentinel-transport-netty-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/HttpHeartbeatSender.java
```diff
@@ -59,14 +59,14 @@
...
+        List<Endpoint> dashboardList = TransportConfig.getConsoleServerList();
         if (dashboardList == null || dashboardList.isEmpty()) {
             RecordLog.info("[NettyHttpHeartbeatSender] No dashboard server available");
             consoleHost = null;
             consolePort = -1;
         } else {
-            consoleHost = dashboardList.get(0).r1;
-            consolePort = dashboardList.get(0).r2;
+            consoleHost = dashboardList.get(0).getHost();
+            consolePort = dashboardList.get(0).getPort();
```
jasonjoo2010, line 69: Maybe HTTPS support is not implemented in `netty-http` module.

jasonjoo2010 (changes requested):

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/function/Endpoint.java
polarbear567: It's my mistake, I'll move them to `com.alibaba.csp.sentinel.transport.endpoint`.

polarbear567:
Hi @jasonjoo2010 ,
It has been optimized according to your suggestions, supported netty http, and verified that the modified code is valid for both http and https. Please review.

jasonjoo2010:
> Hi @jasonjoo2010 ,
> It has been optimized according to your suggestions, supported netty http, and verified that the modified code is valid for both http and https. Please review.

Before reviewing the integration tests failed. You can check and fix it parallelly.

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/SocketFactory.java
```diff
@@ -0,0 +1,27 @@
...
+ * @author Leo Li
+ */
+public class SocketFactory {
+
+    private static class SSLSocketFactoryInstance {
+        private static final SSLSocketFactory SSL_SOCKET_FACTORY = SslFactory.getSslConnectionSocketFactory().getSocketFactory();
+    }
+
+    public static Socket getSocket(Protocol protocol) throws IOException {
+        return protocol == Protocol.HTTP ? new Socket() : getSslSocketFactory().createSocket();
```
jasonjoo2010, line 21: What's the difference between the wrapped form and `Class.varible` form?

jasonjoo2010: I just wrote a small piece testing code and didn't find any different.

sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/endpoint/Protocol.java
```diff
@@ -0,0 +1,23 @@
...
+
+    Protocol(String protocol) {
+        this.protocol = protocol;
+    }
+
+    public String getProtocol() {
+        return protocol;
+    }
+
+    public static void main(String[] args) {
```
jasonjoo2010, line 20: 🤒Do you forget to remove the useless `main`?

polarbear567:
> > Hi @jasonjoo2010 ,
> > It has been optimized according to your suggestions, supported netty http, and verified that the modified code is valid for both http and https. Please review.
> 
> Before reviewing the integration tests failed. You can check and fix it parallelly.

Work fine now.

sentinel-transport/sentinel-transport-netty-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/HttpClientsFactory.java
```diff
@@ -0,0 +1,23 @@
...
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+
+/**
+ * @author Leo Li
+ */
+public class HttpClientsFactory {
+
+    private static class SslConnectionSocketFactoryInstance {
+        private static final SSLConnectionSocketFactory SSL_CONNECTION_SOCKET_FACTORY = new SSLConnectionSocketFactory(SslFactory.getSslConnectionSocketFactory(), NoopHostnameVerifier.INSTANCE);
```
jasonjoo2010, line 16: Hi @sczyh30 

Is bypass certificate verifying acceptable?
I don't insist on it.

jasonjoo2010 (changes requested):

sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/heartbeat/client/SocketFactory.java
polarbear567: > What's the difference between the wrapped form and `Class.varible` form?

This delays init, and threads are safe. Originally written directly in the static block of the class, now it is changed to the inner class.

jasonjoo2010: > 
> This delays init, and threads are safe. Originally written directly in the static block of the class, now it is changed to the inner class.

Yeah, I know. I'm just curious whether it's different to use `SSLSocketFactoryInstance.SSL_SOCKET_FACTORY` simply instead of `getSslSocketFactory()`?
My testing code is for these two approaches.

polarbear567: > > This delays init, and threads are safe. Originally written directly in the static block of the class, now it is changed to the inner class.
> 
> Yeah, I know. I'm just curious whether it's different to use `SSLSocketFactoryInstance.SSL_SOCKET_FACTORY` simply instead of `getSslSocketFactory()`?
> My testing code is for these two approaches.

Oh, it doesn't make any difference, it's a "copy" issue:smile:

jasonjoo2010 (approved):
LGTM.

@sczyh30 Do you have any concerns about the bypass certificate verification approach?

jasonjoo2010:
Thanks for contributing!

PR #1918, opened 2020-12-18
fix concurrentHashMap get and put in large numbers concurrency

samuelxw:
fix concurrentHashMap get and put in large numbers concurrency when MetricFetcher handler response from client

### DIFF
--- a/sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/metric/MetricFetcher.java
+++ b/sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/metric/MetricFetcher.java
@@ -331,25 +331,24 @@ public class MetricFetcher {
                  * aggregation metrics by app_resource_timeSecond, ignore ip and port.
                  */
                 String key = buildMetricKey(machine.getApp(), node.getResource(), node.getTimestamp());
-                MetricEntity entity = map.get(key);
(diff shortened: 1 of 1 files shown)


samuelxw:
代码已提交，麻烦 review

cdfive (approved):
LGTM

cdfive:
The `map` in `handleBody` method is `ConcurrentHashMap`, I suddenly found that it was a local variable, which declared in 
`fetchOnce` method, `fetchOnce->handleResponse->handleBody`, does it affects multithreading safety and performance?

Anyway, using `computeIfAbsent`, it is at least a kind of optimization, which simplifies the writing method and enhances the readability.

cdfive:
> fix concurrentHashMap get and put in large numbers concurrency when MetricFetcher handler response from client

Oh, `in large numbers concurrency`, the size of map may be too big, `computeIfAbsent` is better than `get` and `put`?

If we make sure it's a local variable, how about we use `LinkedHashMap` instead of `ConcurrentHashMap`? since `LinkedHashMap` use default `computeIfAbsent` method in `Map` interface.

How do you think about it? @samuelxw @linlinisme @jasonjoo2010 @sczyh30

cdfive:
Sorry, I didn't read it carefully. Although it is a local variable in the fetchonce method, it is shared by multiple threads.

Issue #1867, opened 2020-11-19
Support setting flush interval of the metric log via SentinelConfig property


yaoyuandota:
您好，我们公司使用sentinel做限流监控，不过metric功能和dashborad有自己的体系，发现metric功能会占用大量内存
180个资源，MetricNode就占用了200M， 虽然这个MetricNode可以gc掉，但是对与我们来说是浪费了。
请问下是不是 重新写下 FlowRuleManager 删除sentinel-metrics-record-task这个定时任务是最方便关闭metric功能的方式

sczyh30:
We may add an attribute `csp.sentinel.metric.flush.interval` to control the flush interval of the metric log. When the interval = 0, the metric log task will not be started (thus disabled). Discussions and contributions are welcomed!

brothelul:
I will try to add the attribute to support the custom interval of metric. need we also support the custom time unit ?
![image](https://user-images.githubusercontent.com/25661357/102683841-c1015f00-420e-11eb-8b34-8926bdb06968.png)
@sczyh30

sczyh30:
Resolved via #1919 (released in 1.8.1)

brucelwl:
> We may add an attribute `csp.sentinel.metric.flush.interval` to control the flush interval of the metric log. When the interval = 0, the metric log task will not be started (thus disabled). Discussions and contributions are welcomed!

加上这个属性之后是不是意味着统计数据会丢失?

PR #1919, opened 2020-12-19
Support setting flush interval of the metric log via SentinelConfig property

brothelul:
…roperty

add a new feature to support control the flush interval of metric via sentinel config: `csp.sentinel.metric.flush.interval`

Fixs [1867](https://github.com/alibaba/Sentinel/issues/1867)
### Describe how you did it
add a new property in sentinel config(`csp.sentinel.metric.flush.interval`). control the metic timer start via this config.
1. if the SentinelConfig#METRIC_FLUSH_INTERVAL more than 0, the timer will run with the SentinelConfig#METRIC_FLUSH_INTERVAL as the rate.
2. if the SentinelConfig#METRIC_FLUSH_INTERVAL less than 0(include) or value is not valid, then means the timer will not be started.

### Describe how to verify it
change the config value,  see the metric log file output and use jvm tool to see if the timer thread is started.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
@@ -26,6 +26,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 import com.alibaba.csp.sentinel.concurrent.NamedThreadFactory;
+import com.alibaba.csp.sentinel.config.SentinelConfig;
 import com.alibaba.csp.sentinel.log.RecordLog;
 import com.alibaba.csp.sentinel.util.AssertUtil;
 import com.alibaba.csp.sentinel.util.StringUtil;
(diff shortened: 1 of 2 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
```diff
@@ -61,9 +62,42 @@
...
+     * <ol>
+     *     <li>if the {@link SentinelConfig#METRIC_FLUSH_INTERVAL} more than 0,
+     * the timer will run with the {@link SentinelConfig#METRIC_FLUSH_INTERVAL} as the rate </li>.
+     *      <li>if the {@link SentinelConfig#METRIC_FLUSH_INTERVAL} less than 0(include) or value is not valid,
+     * then means the timer will not be started </li>
+     * <ol></p>
+     */
+    private static void startMetricTimerListener() {
+        long flushInterval;
+        String flushIntervalStr = SentinelConfig.getConfig(SentinelConfig.METRIC_FLUSH_INTERVAL);
```
sczyh30, line 79: Can we add a `metricLogFlushIntervalSec()` method in `SentinelConfig` to resolve the metric flushing interval, and leverage the method here?

brothelul: yes, that is more reasonable to put the time valid code into entrance. I will change the code later.

brothelul:
@sczyh30 I have change the code as we discussed.

sczyh30 (approved):
LGTM

sczyh30:
Nice work. Thanks for contributing!

Issue #1814, opened 2020-10-23
Expose jmx endpoint for sentinel


qixiaobo:
## Issue Description

Type:  *feature request*

### Describe what happened (or what feature you want)
jmx support for sentinel metrics

so  we can use prometheus jmx_exporter to collect metric
also jmx is standard to manage our java bean.

>>>Java Management Extensions (JMX) was introduced in J2SE 5.0 release. It provides an architecture to manage resources dynamically at runtime. JMX is used mostly in enterprise applications to make the system configurable or to get the state of application at any point of time.

sczyh30:
Good idea. Discussion and contributions are welcomed!

brothelul:
I will try to add this feature.

sczyh30:
> I will try to add this feature.

Look forward to your PR~

PR #1947, opened 2021-01-06
[feature] Expose jmx endpoint for sentinel

brotherlu-xcq:
expose the metric in Sentinel as jmx bean, which can make the Sentinel metric collected by popular monitor tools like prometheus.

Fixs #1814 
### Describe how to verify it
visit the resource which control by Sentinel, the open your JMX tool. you will find the metric info like below.
image
the root type is `Sentinel`. the second type is your sentinel application name, the JMX bean name is you resource name. the metric data is same with the MetricNode which is writed into metric file.


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
```diff
@@ -56,13 +56,14 @@
...
-    private static final ScheduledExecutorService SCHEDULER = Executors.newScheduledThreadPool(1,
+    private static final ScheduledExecutorService SCHEDULER = Executors.newScheduledThreadPool(2,
         new NamedThreadFactory("sentinel-metrics-record-task", true));
 
     static {
         flowRules.set(Collections.<String, List<FlowRule>>emptyMap());
         currentProperty.addListener(LISTENER);
         startMetricTimerListener();
+        SCHEDULER.scheduleAtFixedRate(new MetricBeanTimerListener(), 0, 1, TimeUnit.SECONDS);
     }
```
cdfive, line 67: Now there are two thread to execute, is there data inaccuracy in multithreading?
Tried with `FlowQpsDemo`, the `PassQps`,`BlockQps` are always 0 in jconsole.

`com.alibaba.csp.sentinel.node.metric.MetricWriter#write` 
`com.alibaba.csp.sentinel.node.metric.jmx.MetricBeanWriter#write`
When the two operations are put in a same thread, it works fine and got the expected result in jconsole.

Maybe that's the reason, as the comment mentioned in `com.alibaba.csp.sentinel.node.StatisticNode#metrics`,
**The fetch operation is thread-safe under a single-thread scheduler pool.**
```java
@Override
public Map<Long, MetricNode> metrics() {
    // The fetch operation is thread-safe under a single-thread scheduler pool.
    long currentTime = TimeUtil.currentTimeMillis();
    currentTime = currentTime - currentTime % 1000;
    ...

    return metrics;
}
```

brothelul: thank you to review my code, I will check the codes and sovle it.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/StatisticNode.java
```diff
@@ -113,7 +113,7 @@
     private long lastFetchTime = -1;
 
     @Override
-    public Map<Long, MetricNode> metrics() {
+    public synchronized Map<Long, MetricNode> metrics() {
```
brothelul, line 116: @cdfive because  the two timer have different bussiness scenes, so I am not put them in one timer and add the synchronized to ensure the thread-safe. could you please help to review this when you you are free? I ran the local test and can find the data in jvm tool.

cdfive: I run the `FlowQpsDemo`, and start jconsole, still can't find `Sentinel` node in jconsole UI. Then I make a breakpoint at `com.alibaba.csp.sentinel.node.metric.MetricTimerListener#aggregate`, and `Sentinel` node appeared, but the the PassQps,BlockQps are always 0.

I wonder that once the `StatisticNode#metrics`have been invoked, the data will be changed.
could you please try the demo and give some help? @sczyh30 @brothelul

brothelul: I will try the demo and re-produce the problem you met again, thank you so much!

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/jmx/MetricBeanTimerListener.java
```diff
@@ -40,11 +40,12 @@ public void run() {
     
     private List<MetricNode> getLastMetrics(ClusterNode node) {
         final long currentTime = TimeUtil.currentTimeMillis();
-        final long minTime = currentTime - currentTime % 1000;
+        final long maxTime = currentTime - currentTime % 1000;
+        final long minTime = maxTime - 1000;
         return node.rawMetricsInMin(new Predicate<Long>() {
```
brothelul, line 45: @cdfive I change the poolSize of SCHEDULER into 1, so the two listener will run one by one in thread-safe.  And I use the method  rawMetricsInMin instead of metrics to get the metrics in last sencod so that the JMX will be same as  metrics in file.

brothelul: @cdfive any idea about this?

cdfive: @brothelul Nice work! I can get the expected result in jconsole now.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
```diff
@@ -55,16 +55,26 @@
...
     
+    /**
+     * start the MetricBeanTimerListener at the fix rate, the initialDelay is different with the {@link #startMetricTimerListener()},
+     * so that the two task will start running at different time.
+     */
+    private static void startMetricBeanTimerListener() {
+        SCHEDULER.scheduleAtFixedRate(new MetricBeanTimerListener(), 0, 1, TimeUnit.SECONDS);
+    }
+
```
cdfive, line 77: I found that we can config the interval with `csp.sentinel.metric.flush.interval` in `startMetricTimerListener`,
but `startMetricBeanTimerListener` has one second fixed interval. Do we need to make it configurable? At present, this function is enabled by default, maybe some users want to disable it.

brothelul: yes, that's a good idea! I will change it later.

cdfive:
Nice work! With this jmx endpoint, it will be helpful when collecting monitor data. 
Any other idea? @sczyh30 @jasonjoo2010 

BTW, the issue which this PR is to fix seems not corresponding. Maybe another issue? @brothelul 
```
Fixs #1848
```

brothelul:
yes, it's #1814, not #1848. I changed it.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/jmx/MetricBeanWriter.java
```diff
@@ -0,0 +1,78 @@
...
+            return;
+        }
+        String appName = SentinelConfig.getAppName();
+        if (appName == null) {
+            appName = "sentinel-application";
+        }
+        Set<String> existResource = new HashSet<>();
+        // set or update the new value
+        for (MetricNode metricNode : map.values()) {
+            final String mBeanName = "Sentinel:type=" + appName + ",name=\"" + metricNode.getResource() +"\"";
```
tianshuang, line 51: It is recommended to add classification here, so that Prometheus can search according to classification.
ex:
`final String mBeanName = "Sentinel:type=" + appName + ",name=\"" + metricNode.getResource() +"\"" + ",classification=\"" + metricNode.getClassification() +"\"";`

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/jmx/MetricBeanWriter.java
```diff
@@ -0,0 +1,78 @@
...
+            existResource.add(mBeanName);
+        }
+        // reset the old value
+        List<MetricBean> metricBeans = mBeanRegistry.listAllMBeans();
+        if (metricBeans == null || metricBeans.isEmpty()) {
+            existResource.clear();
+            return;
+        }
+        for (MetricBean metricBean : metricBeans) {
+            final String mBeanName = "Sentinel:type=" + appName + ",name=\"" + metricBean.getResource() +"\"";
```
tianshuang, line 70: Same here as above.
ex: 
`final String mBeanName = "Sentinel:type=" + appName + ",name=\"" + metricBean.getResource() +"\"" + ",classification=\"" + metricBean.getClassification() +"\"";`

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/jmx/MetricBeanWriter.java
brotherlu-xcq: @sczyh30 @cdfive Do you think so? I am not clear about the work of Prometheus?

sczyh30:
Nice work. IMHO, how about make it an extension module (users could enable it by just adding a dependency of the extension module)? This could decouple metrics exporters from `FlowRuleManager` and core (actually initializing it in static block of `FlowRuleManager` is not a good design...)

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
```diff
@@ -55,16 +55,33 @@
...
 
+    /** the corePool size of SCHEDULER must be set at 1, so the two task ({@link #startMetricTimerListener()}, {@link #startMetricBeanTimerListener()}) can run orderly by the SCHEDULER **/
     @SuppressWarnings("PMD.ThreadPoolCreationRule")
     private static final ScheduledExecutorService SCHEDULER = Executors.newScheduledThreadPool(1,
         new NamedThreadFactory("sentinel-metrics-record-task", true));
 
     static {
         flowRules.set(Collections.<String, List<FlowRule>>emptyMap());
         currentProperty.addListener(LISTENER);
+        startMetricBeanTimerListener();
```
sczyh30, line 66: Maybe we could initialize it via `InitFunc` SPI mechanism?

brotherlu-xcq:
> Nice work. IMHO, how about make it an extension module (users could enable it by just adding a dependency of the extension module)? This could decouple metrics exporters from `FlowRuleManager` and core (actually initializing it in static block of `FlowRuleManager` is not a good design...)

I am really agree with you that export the metric function to other module instead of in core module. I want still work on this feature, and make a design first about it.

brotherlu-xcq:
This function need some change after we discussed. please do not merge it before I re-submit the new implement.

sczyh30:
#2275 has been merged

Issue #1967, opened 2021-01-15
建议LoginAuthenticationFilter 属性authFilterExcludeUrls配置支持uri: /xx/**


xiaoxb:
## Issue Description

Type: *bug report* or *feature request*

### Describe what happened (or what feature you want)
url只能精准匹配
if(authFilterExcludeUrls.contains(servletPath)) {
	chain.doFilter(request, response);
	return;
}
### Describe what you expected to happen

1. 

// Exclude the urls which needn't auth 
for (String authFilterExcludeUrl : authFilterExcludeUrls) {
	if (isMatch(authFilterExcludeUrl, servletPath)) {
		chain.doFilter(request, response);
		return;
	}
}

// 判断url是否与规则配置: ? 表示单个字符; * 表示一层路径内的任意字符串，不可跨层级; ** 表示任意层路径;
public Boolean isMatch(String pattern, String url) {
    AntPathMatcher matcher = new AntPathMatcher();
    return matcher.match(pattern, url);
}

2.
private static final AntPathMatcher PATH_MATCHER = new AntPathMatcher();

if (authFilterExcludeUrls.stream().anyMatch(s -> PATH_MATCHER.match(s, servletPath))) {
	chain.doFilter(request, response);
	return;
}

### How to reproduce it (as minimally and precisely as possible)

1. 
2. 
3. 

### Tell us your environment
master

sczyh30:
Would you like to contribute a PR to fix it?

brothelul:
I want to solve it If there's no one.

xiaoxb:
Of course
















At 2021-01-18 09:54:22, "brothelul" <notifications@github.com> wrote:

I want to solve it If there's no one.

sczyh30:
Resolved in #1971

PR #1971, opened 2021-01-19
[feature] Feature #1967 LoginAuthenticationFilter authFilterExcludeUrls match supports uri like /xx/**

brothelul:
LoginAuthenticationFilter authFilterExcludeUrls match supports uri like /xx/**

Fixs #1967 

### Describe how you did it
add the `org.springframework.util.AntPathMatcher` to support matching the uri like /xx/**
 
### Describe how to verify it
request the exclude urls


cdfive:
Good method! How about we apply this new feature? 
In `application.properties`: `auth.filter.exclude-urls=/,/auth/login,/auth/logout,/registry/machine,/version`
`/auth/login,/auth/logout` seems can reduced to `/auth/*`, the other url`/auth/check` can be accessed without login.

brothelul:
it seems can not work ideal with the current excludeUrls, we can keep this feature not change the properties `auth.filter.exclude-urls`. maybe this feature can be apply to the new excludeUrls in the future.

cdfive (approved):
LGTM. Thanks for contributing and provide this new feature!

sczyh30 (approved):
LGTM

sczyh30:

PR #1996, opened 2021-02-01
Update source and target JDK version to 1.8

### LINKED ISSUE #1968: Upgrade min required JDK version to 1.8
## Issue Description

Upgrade min required JDK version to 1.8 so that we could leverage features of JDK 8. This won't affect many users as most of the developers are using JDK 1.8 or above.

sczyh30:

Update project source and target JDK version to 1.8.

Resolves #1968 

### Describe how you did it

Just update the source version and target version in the config of `maven-compiler-plugin`. The code remains unchanged to reduce unnecessary changes.

### Describe how to verify it

Run all test cases for regression.

### Special notes for reviews

This PR contains breaking changes (but not related to API changes).


pom.xml
```diff
@@ -53,8 +53,8 @@
 
         <!-- Build -->
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <java.source.version>1.7</java.source.version>
-        <java.target.version>1.7</java.target.version>
+        <java.source.version>1.8</java.source.version>
+        <java.target.version>1.8</java.target.version>
```
cdfive, line 57: By global search `java.source.version` in project, `sentinel-api-gateway-adapter-common`, `sentinel-cluster`, `sentinel-datasource-zookeeper`, `sentinel-logging-slf4j` module has their own separate configuration `java.source.version` with JDK1.7. Suggest unify them into the configuration of the parent pom.xml, which will be more convenient for upgrade in the future.

sczyh30: Updated

cdfive: There are still some modules has own `java.source.version` configuration with JDK1.8, `sentinel-demo`,`sentinel-apache-dubbo-adapter`,`sentinel-datasource-redis`, do we need to delete them and keep only one in parent module
JDK1.8 has `LongAdder` class, how about replacing `com.alibaba.csp.sentinel.slots.statistic.base.LongAdder`?

sczyh30: Good idea. I've updated, please check.

> The same as Function,Supplier... classes in com.alibaba.csp.sentinel.util.function package.

This may bring breaking changes (as some of them were used as args of exported functions). Maybe we could deprecate them in the next major version.

cdfive (approved):
LGTM

PR #2046, opened 2021-02-24
fix(dashboard): ${user.home} is resolved when maven package

### LINKED ISSUE #2045: Dashboard config logging.file's value should not be resolved after maven package
## Issue Description

Type: *bug report*

### Describe what happened

`application.properties` file in sentinel-dashboard value not correct after `mvn package`

In `sentinel-dashboard.jar`'s file `BOOT-INF/classes/application.properties`

The config `logging.file` in source code `sentinel-dashboard/src/main/resources/application.properties` is 

```properties
logging.file=${user.home}/logs/csp/sentinel-dashboard.log
```

But after `mvn clean package`, the content change.

The value of `${user.home}` is resolved to `C:\Users\bob`, `C:\Users\alice` or `/home/tom`, which base on user's home path.

`${user.home}` is not need to be resolve after maven package, wish it keep until runtime, i.e when dashboard running.

#### 1. package

```bash
cd sentinel-dashboard
mvn clean package -DskipTests
```

#### 2. check

Uncompress `sentinel-dashboard/target/sentinel-dashboard.jar`

Open file `BOOT-INF/classes/application.properties`

View the value of key `logging.file` , `${user.home}/logs/csp/sentinel-dashboard.log` now change to `user-absolute-path/logs/csp/sentinel-dashboard.log`

### Tell us your environment

Windows 10

version: 1.8.1

Anilople:

${user.home} in `application.properties` should be resolved in dashboard runtime, not after maven package.

Fixes #2045 

### Describe how you did it

Change maven-resources-plugin's behavior by use delimiter `@`

### Describe how to verify it

Run `mvn clean package -DskipTests` on project, then check `BOOT-INF/classes/application.properties` in file `sentinel-dashboard/target/sentinel-dashboard.jar`

`application.properties`'s content is changed.

Only 
```properties
sentinel.dashboard.version=@project.version@
```
is changed to
```properties
sentinel.dashboard.version=1.8.2-SNAPSHOT
```

And
```properties
logging.file=${user.home}/logs/csp/sentinel-dashboard.log
```
keep same as source code.

### DIFF
--- a/sentinel-dashboard/pom.xml
+++ b/sentinel-dashboard/pom.xml
@@ -13,6 +13,7 @@
     <packaging>jar</packaging>
 
     <properties>
+        <resource.delimiter>@</resource.delimiter>
         <spring.boot.version>2.0.5.RELEASE</spring.boot.version>
         <curator.version>4.0.1</curator.version>
     </properties>
@@ -138,6 +139,20 @@
 
     <build>
         <finalName>sentinel-dashboard</finalName>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-resources-plugin</artifactId>
+                    <configuration>
+                        <delimiters>
+                            <delimiter>${resource.delimiter}</delimiter>
+                        </delimiters>
+                        <useDefaultDelimiters>false</useDefaultDelimiters>
+                    </configuration>
+                </plugin>
+            </plugins>
+        </pluginManagement>
         <plugins>
             <plugin>
                 <groupId>org.springframework.boot</groupId>
--- a/sentinel-dashboard/src/main/resources/application.properties
+++ b/sentinel-dashboard/src/main/resources/application.properties
@@ -21,4 +21,4 @@ auth.password=sentinel
 
 # Inject the dashboard version. It's required to enable
 # filtering in pom.xml for this resource file.
-sentinel.dashboard.version=${project.version}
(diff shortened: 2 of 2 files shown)


sentinel-dashboard/pom.xml
```diff
@@ -13,6 +13,7 @@
     <packaging>jar</packaging>
 
     <properties>
+        <resource.delimiter>@</resource.delimiter>
```
cdfive, line 16: Do we need to specify the delimiter explicitly? 
It seems that the default delimiter is `@`, and the `maven-resources-plugin` can also be omitted.
Certainly, explicit specification has the effect of emphasis. 
Using default or specify the delimiter and resource plugin, which do you think is better?

cdfive: I'm quite sorry, the project does not inherit spring boot starter parent.
Specify the delimiter is right and adding `maven-resources-plugin` is necessary.

cdfive (approved):
LGTM.
According the document the issue referred, the default delimiters are, 
```xml
<delimiters>
  <delimiter>${*}</delimiter>
  <delimiter>@</delimiter>
</delimiters>
```
So here need to specify the `@` as the only delimiter.
I package it, and get the expected `application.properties` in the fat jar.
Run the jar, get the expected log file.
It works fine and must be very helpful for users, thanks! @Anilople

sczyh30 (approved):

sczyh30:
Thanks for contributing!

Issue #2053, opened 2021-02-27
Dashboard Spring Boot Configuration more flexible support


Anilople:
## Issue Description

Type: *feature request*

### Describe what happened (or what feature you want)

Make dashboard more extensible.

### Describe what you expected to happen

Spring Boot Configuration more flexible

Use `@ConditionalOnMissingBean` not `@Component` directly, combine with `@Configuration` to create a bean.

For example, about `MachineDiscovery`.

We can write code like follow

```java
@Configuration
public class MachineDiscoveryConfiguration {

  @Bean
  @ConditionalOnMissingBean
  public SimpleMachineDiscovery simpleMachineDiscovery() {
    return new SimpleMachineDiscovery();
  }

  @Bean
  @ConditionalOnMissingBean
  public AppManagement appManagement(SimpleMachineDiscovery simpleMachineDiscovery) {
    return new AppManagement(simpleMachineDiscovery);
  }

}
```

And delete annotation `@Component` above class `SimpleMachineDiscovery` and `AppManagement`.

If someone want to custom their `AppManagement`, that will make it more easy.

Same principle can apply to `RuleRepository`, DynamicRuleProvider, DynamicRulePublisher too.

kevinjad:
@sczyh30 @Anilople Can I write this one up?

Anilople:
> @sczyh30 @Anilople Can I write this one up?

Maybe we can use multiple pull request to refactor those configuration.

I have been refactor the configuration of `MetricsRepository` in #2055 

Do you want refactor the configuration of `MachineDiscovery`? Welcome

priyanka010392:
@Anilople Is this issue still open? If yes, I can take this up.

PR #2059, opened 2021-02-27
refactor(dashboard): use AuthConfiguration to manage bean of auth

Anilople:

That will make auth module more flexible.

#2053 

### Describe how you did it

Use `AuthConfiguration` to manage bean under package `com.alibaba.csp.sentinel.dashboard.auth`.

Change `AuthorizationInterceptor` to an interface, and support a default implementation.

Change `LoginAuthenticationFilter` to an interface, and support a default implementation.

### Describe how to verify it

The beans are created successfully, and not break old testes.

### DIFF
--- a/sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/auth/DefaultAuthorizationInterceptor.java
+++ b/sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/auth/DefaultAuthorizationInterceptor.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 1999-2018 Alibaba Group Holding Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.alibaba.csp.sentinel.dashboard.auth;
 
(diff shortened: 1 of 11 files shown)


sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/config/AuthConfiguration.java
```diff
@@ -0,0 +1,42 @@
+package com.alibaba.csp.sentinel.dashboard.config;
+
+import com.alibaba.csp.sentinel.dashboard.auth.*;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import javax.servlet.http.HttpServletRequest;
```
cdfive, line 9: Could you please add license header to the new files?

sentinel-dashboard/src/main/java/com/alibaba/csp/sentinel/dashboard/auth/DefaultAuthorizationInterceptor.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright 1999-2018 Alibaba Group Holding Ltd.
```
Anilople, line 2: Should change to `1999-2021` here?

cdfive: Both are OK, since there are remain other files which marked as `1999-2018`, we may uniform them in the future.

cdfive (approved):
LGTM
After refactoring, it's convenient for users to customize their own implementation of `AuthService`,`LoginAuthenticationFilter`,`AuthorizationInterceptor`, the dashboard module becomes more extensible. The idea of interface design and auto configuration is smart.
Thanks! @Anilople

sczyh30 (approved):
LGTM

sczyh30:
Nice work. Thanks for contributing and look forward to more!

PR #2067, opened 2021-03-04
add server initiative timeout，prevent bizThreadPool full

coderDylan:

The simple webserver in transport-simple-http ,add server initiative timeout，prevent bizThreadPool full。
Although there are already has setSocketSoTimeout，but the eventTask thread will not be released。
When some requests are slow，the bizThreadPool will full，then many normal requests may be affected


coderDylan:
cancel

PR #2072, opened 2021-03-05
add server initiative timeout，prevent bizThreadPool full

coderDylan:

The simple webserver in transport-simple-http ,add server initiative timeout，prevent bizThreadPool full。
Although there are already has setSocketSoTimeout，but the eventTask thread will not be released。
When some requests are slow，the bizThreadPool will full，then many normal requests may be affected

### Describe how you did it
use the JUC Future timeout

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/jul/ConsoleHandler.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/log/jul/ConsoleHandler.java
@@ -15,7 +15,16 @@
  */
 package com.alibaba.csp.sentinel.log.jul;
 
+import com.alibaba.csp.sentinel.concurrent.NamedThreadFactory;
+
 import java.io.UnsupportedEncodingException;
+import java.util.ArrayDeque;
+import java.util.Queue;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.RejectedExecutionHandler;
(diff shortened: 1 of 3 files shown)


sentinel-transport/sentinel-transport-simple-http/src/main/java/com/alibaba/csp/sentinel/transport/command/SimpleHttpCommandCenter.java
```diff
@@ -182,13 +175,23 @@ public void stop() throws Exception {
...
                     socket = this.serverSocket.accept();
                     setSocketSoTimeout(socket);
+                    //Give it to the biz threadPool for processing
                     HttpEventTask eventTask = new HttpEventTask(socket);
-                    bizExecutor.submit(eventTask);
+                    //add server initiative timeout，prevent bizThreadPool full normal requests may be affected
+                    Future<String> future = bizExecutor.submit(eventTask,"ok");
+                    try {
+                        future.get(DEFAULT_SERVER_SO_TIMEOUT, TimeUnit.MILLISECONDS);
+                    }catch (TimeoutException | InterruptedException | ExecutionException ex){
```
sczyh30, line 190: Could you please reformat your code?

coderDylan: I'll try，I always use Eclipse Code Formatter auto format when save the code file。The IDE has just been upgraded, may not work

PR #2082, opened 2021-03-15
Customize fallback config

majianbing:
Sentinel's users can define the Fallback rule, load manually or by data source, when a block happened, return the customize Fallback data to the resource caller.
fallback string usually is a JSON string, If you want to return an Object when blocked(such as Dubbo RPC fallback), you can define your class Reference of this JSON string and serialize it by any serialize framework such as Fastjson/Gson/Jackson.

Fixes: https://github.com/alibaba/Sentinel/issues/2074

### Describe how you did it
The same as the flow rule

### Describe how to verify it
Test unit

### Special notes for reviews
Commit Dubbo and HTTP Web MVC adapter as an example


majianbing:
Hi, do I need to fix other's tests?

LearningGp:

PR #2085, opened 2021-03-15
Add separate demo for slot chain spi

cdfive:

There is only a demo for demonstrating how to customize slot at present, users may want to customize
the slot chain sometimes. Since the SPI mechanism has been refactored in version of 1.8.1,  adding separate 
demo for customize slot and slot chain will be useful for new users.

NONE

### Describe how you did it

Split `sentinel-demo-slot-chain-spi` module into two modules:
* `sentinel-demo-slot-spi` 
Demonstrate how to custom slot.

* `sentinel-demo-slotchain-spi`
Demonstrate how to custom slot chain.
The custom slot chain doesn't has `DegradeSlot`,  
In `DemoFlowRuleApplication`, flow rule takes affect.
In `DemoDegradeRuleApplication`, degrade rule doesn't take affect.

Since the sentinel's default slots and the order of them are very important, comments are added in the demo.

### Describe how to verify it
Run `DemoApplication`,`DemoFlowRuleApplication`,`DemoDegradeRuleApplication` in the two demo module, and watch the console output.

### Special notes for reviews
The `SlotDemo` class name changed to `DemoApplication`, this may help reminding users that it can run directly through the class name.


sczyh30 (approved):
LGTM

sczyh30:
Nice work. Thanks!

PR #2143, opened 2021-04-19
fix(dashboard): registry multiple projects will throw exception when app id doesn't exist

Anilople:

fix(dashboard): registry multiple projects will throw exception when app id doesn't exist.

That will cause authorization failed in the middle.

NONE

### Describe how you did it

Catch exception when use method `com.alibaba.csp.sentinel.dashboard.apollo.service.ApolloPortalService#assignAppRoleToSentinelDashboard`

### Describe how to verify it

Manually.


Anilople:
Sorry again....

PR #2150, opened 2021-04-21
Fix:Sentinel's dubbo provider filter may fail to capture unsuccessful…

goodjava:
sentinel  dubbo provider filter 在方法是异步执行的情况下,会收集不到执行的打点.(如果是dubbo2.7的版本,还会造成异步变同步阻塞)

abc

eeee

### DIFF
--- a/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java
+++ b/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java
@@ -20,12 +20,12 @@ import com.alibaba.csp.sentinel.adapter.dubbo.config.DubboAdapterGlobalConfig;
 import com.alibaba.csp.sentinel.context.ContextUtil;
 import com.alibaba.csp.sentinel.log.RecordLog;
 import com.alibaba.csp.sentinel.slots.block.BlockException;
-
 import org.apache.dubbo.common.extension.Activate;
-import org.apache.dubbo.rpc.Invocation;
-import org.apache.dubbo.rpc.Invoker;
-import org.apache.dubbo.rpc.Result;
-import org.apache.dubbo.rpc.RpcException;
+import org.apache.dubbo.rpc.*;
+import org.apache.dubbo.rpc.support.RpcUtils;
+
+import java.util.LinkedList;
+import java.util.Optional;
 
 import static org.apache.dubbo.common.constants.CommonConstants.PROVIDER;
 
@@ -61,6 +61,7 @@ public class SentinelDubboProviderFilter extends BaseSentinelDubboFilter {
     @Override
     public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
         // Get origin caller.
+        InvokeMode invokeMode = RpcUtils.getInvokeMode(invoker.getUrl(), invocation);
         String origin = DubboAdapterGlobalConfig.getOriginParser().parse(invoker, invocation);
         if (null == origin) {
             origin = "";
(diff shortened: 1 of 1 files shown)


sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java
```diff
@@ -61,6 +61,7 @@ String getInterfaceName(Invoker invoker, String prefix) {
     @Override
     public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
         // Get origin caller.
+        InvokeMode invokeMode = RpcUtils.getInvokeMode(invoker.getUrl(), invocation);
```
cdfive, line 64: Thank you for pointing out the problem of the provider in the asynchronous scenario!

I test it and find two problems to be discussed.

If the interface method using `CompletableFuture<T>` as return type, `RpcUtils.getInvokeMode` will return `InvokeMode.FUTURE`, and `syncInvoke` will be called. In the scenario where the provider does asynchronous execution, should we use `invokemode. Future` to distinguish between sync  and async execution？

Another case is using `AsyncContext`, how can we judge whether it's a async execution in provider

Refer to [https://dubbo.apache.org/zh/docs/v2.7/user/examples/async-execute-on-provider/](https://dubbo.apache.org/zh/docs/v2.7/user/examples/async-execute-on-provider/)

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java
```diff
@@ -70,13 +71,32 @@ public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcExcept
         String prefix = DubboAdapterGlobalConfig.getDubboProviderResNamePrefixKey();
         String interfaceResourceName = getInterfaceName(invoker, prefix);
         String methodResourceName = getMethodName(invoker, invocation, prefix);
+        ContextUtil.enter(methodResourceName, origin);
         try {
             // Only need to create entrance context at provider side, as context will take effect
             // at entrance of invocation chain only (for inbound traffic).
-            ContextUtil.enter(methodResourceName, origin);
```
cdfive, line 76: The comments could be put before `ContextUtil.enter(methodResourceName, origin);`, making it consistent with the code.

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java
```diff
@@ -85,18 +105,62 @@ public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcExcept
...
+                    SphU.asyncEntry(interfaceResourceName, ResourceTypeConstants.COMMON_RPC, EntryType.IN), null));
+            queue.push(new EntryHolder(
+                    SphU.asyncEntry(methodResourceName, ResourceTypeConstants.COMMON_RPC,
+                            EntryType.IN, invocation.getArguments()), invocation.getArguments()));
+            Result result = invoker.invoke(invocation);
+            result.whenCompleteWithContext((r, throwable) -> {
+                Throwable error = throwable;
+                if (error == null) {
+                    error = Optional.ofNullable(r).map(Result::getException).orElse(null);
+                }
```
cdfive, line 132: It's elegant to use `Optional.ofNullable(r)` in some scenarios, but the side effect is that it'll create new `Optional` object, if `r` is not null.
Since here is a simple judgment scenario and the filter will be called every time the interface is called,
Will this be better? `error = r != null ? r.getException() : null;` 

Maybe in most cases, `r` is null if no exception occurred, I think it's acceptable using `Optional.ofNullable(r)`.

How do you think about it? @goodjava @sczyh30

linlinisme: I think `error = r != null ? r.getException() : null` is more clear than using `optional` here

sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/SentinelDubboProviderFilter.java
```diff
@@ -85,18 +105,62 @@ public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcExcept
...
+                    exitEntry(holder);
+                }
+            });
+            return result;
+        } catch (BlockException e) {
+            while (!queue.isEmpty()) {
+                exitEntry(queue.pop());
+            }
+            return DubboAdapterGlobalConfig.getProviderFallback().handle(invoker, invocation, e);
+        }
```
linlinisme, line 145: about the resource clean up(`exitEntry(holder)`),  had better move to a `finally` block which can guarantee it be called

LearningGp:

PR #2154, opened 2021-04-23
An improved method for obtaining ip address and port number from cont…

goodjava:
如果用户代码docker化后,env的操作成本要比-D传进来的成本低很多.因此env是原生云环境下指定ip和端口的比较好的办法.
env
yes
yes2
这样dashboard 推送rule信息修改就没有任何问题了

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfig.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfig.java
@@ -125,6 +125,24 @@ public final class SentinelConfig {
         AssertUtil.notNull(key, "key cannot be null");
         return props.get(key);
     }
+    
+    /**
+     * Get config value of the specific key.
+     *
+     * @param key config key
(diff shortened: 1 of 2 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/config/SentinelConfig.java
```diff
@@ -125,6 +125,17 @@ public static String getConfig(String key) {
         AssertUtil.notNull(key, "key cannot be null");
         return props.get(key);
     }
+    
+    public static String getConfig(String key, boolean env) {
```
sczyh30, line 129: Could you please add comments for this?

goodjava: 好的

sczyh30 (approved):
LGTM

sczyh30:
Good idea. Supporting config from system env is necessary for cloud-native services. The mechanism of `SentinelConfig` could be refactored later.

sczyh30:

PR #2170, opened 2021-04-27
[FIX #2155] the degrade should contain the boundary

### LINKED ISSUE #2155: Use degraderule and add exception rate equals 1.0,it's invalid
### Type:bug report
When use degraderule,add exception rate equals 1.0,it's invalid。
And when I view the source code,it's:
        **double curCount = errCount;
        if (strategy == DEGRADE_GRADE_EXCEPTION_RATIO) {
            // Use errorRatio
            curCount = errCount * 1.0d / totalCount;
        }
        if (curCount > threshold) {
            transformToOpen(curCount);
        }
    }**
so it will not greater than 1，and the value [0.0-1.0] make  any  means？

2、use jdk1.8,sentinel 1.8.0,windows10

brothelul:
when the degrade rule rate value equals to  threshold,  the degrade not work.

Fixes #2155


brothelul:
I will check the CI issue later.

PR #2176, opened 2021-04-29
Improve DefaultDubboFallback implementation(apache-dubbo-adapter).

wuwen5:
The caller cannot determine the type of runtime exception thrown.
I suggest the same way as sentinel-dubbo-adapter(dubbo-2.6.x).

#1252


sczyh30:
Hi, thanks for contributing. The exception type has been carried in the message of the wrapped RuntimeException. SentinelRpcException is not recommended any more.

wuwen5:
SentinelRpcException is a subclass of RuntimeException so callers can capture and judge it easily, while RuntimeException (```ex.toRuntimeException()```) can only be captured and judged by error message, which is very unfriendly.

```
try {...
} catch (RuntimeException e) {
 if (e.getMessage().contains("SentinelBlockException")) {
                //...
  }
}
```

```
try {
} catch (SentinelRpcException e) {
  // is friendly
}
```

PR #2219, opened 2021-05-29
fix spring aop demo that handles BlockException

Roger3581321:
Because of the malfunction of the demo code, this PR enables how handler method handles BlockException

YES

### Describe how you did it
throws a BlockException in a random way

### Describe how to verify it
Call the method a dozens of times and check the console


sczyh30:
Hi, throwing `BlockException` directly is not the recommended way for the demo (and also not a practice to put `throws BlockException` in method signature). If you're trying the block handler, you may add a flow rule for it first (e.g. QPS=0).

PR #2274, opened 2021-07-01
Use StringBuilder#append(Char) to improve performance

hunjipo:
This commit switches to StringBuilder.append(char) instead of StringBuilder.append(String) whenever we append a single character to a StringBuilder.
StringBuffer also is the same.

no, just improve performance.


brotherlu-xcq:
@sczyh30 this PR maybe need approve to run CI Actions.

PR #2275, opened 2021-07-02
[feature] Expose jmx endpoint for sentinel.

### LINKED ISSUE #1814: Expose jmx endpoint for sentinel

brotherlu-xcq:
The new implement of JMX export after we discussed. old implement please refer #1947 
the JMX data like below:
image

Fixes #1814


sczyh30:
Hi, any progress on this PR?

brotherlu-xcq:
It looks like there have no one review it.

sczyh30:
> It looks like there have no one review it.

I'll take a review these days. cc @jasonjoo2010 @cdfive

brotherlu-xcq:
thanks

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/exporter/jmx/MetricBeanWriter.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+        for (MetricBean metricBean : metricBeans) {
+            final String mBeanName = "Sentinel:type=" + appName + ",name=\"" + metricBean.getResource()
+                    +"\",classification=\"" + metricBean.getClassification() +"\"";
+            if (existResource.contains(mBeanName)) {
+                continue;
+            }
+            metricBean.reset();
+        }
+        existResource.clear();
```
jasonjoo2010, line 97: Currently, we use an extra data structure to reset those elements in the registry that are not included in this round of write.

How about introducing a version(or take the current timestamp as a version) in each write? Then no extra structure is needed and it's also more efficient.

And btw, do they have the opportunity to get removed from the registry if they only appear for a short time?

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/exporter/jmx/MetricBean.java
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * reset the MBean value to the initialized value.
+     */
+    public void reset() {
+        this.blockQps = 0;
+        this.passQps = 0;
+        this.timestamp = System.currentTimeMillis();
+        this.exceptionQps = 0;
+        this.occupiedPassQps = 0;
+        this.successQps = 0;
```
jasonjoo2010, line 113: The reset() is not completely paired with set(). No need to reset other properties like timestamp, concurrency, rt, and so on?

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/exporter/jmx/MBeanRegistry.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+    
+    /**
+     * unregister the MetricBean
+     * @param bean MetricBean
+     */
+    public void unRegister(MetricBean bean) {
+        assert bean != null;
+        String beanName = mapBean2Name.get(bean);
```
jasonjoo2010, line 88: How about putting the bean name into `MetricBean` hence we can only maintain one map?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
```diff
@@ -61,7 +62,7 @@
         currentProperty.addListener(LISTENER);
         startMetricTimerListener();
     }
-
+
```
jasonjoo2010, line 65: I think there's no need to reorder the imports and introduce extra spaces.

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/MetricExporterInit.java
```diff
@@ -0,0 +1,81 @@
...
+    
+    /**
+     * the list of metric exporters.
+     */
+    private static List<MetricExporter> metricExporters = new ArrayList<>();
+    
+    /**
+     * the metric collector.
+     */
+    private static MetricCollector metricCollector = new MetricCollector();
```
jasonjoo2010, line 46: I think it's not necessary to keep the reference here. And even it's not necessary to pass the reference into the constructor of JMXMetricExporter.

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/collector/MetricCollector.java
```diff
@@ -0,0 +1,92 @@
...
+    
+    /**
+     * Get the last second {@link MetricNode} of {@link ClusterNode}
+     * @param node {@link ClusterNode}
+     * @return the list of {@link MetricNode}
+     */
+    private List<MetricNode> getLastMetrics(ClusterNode node) {
+        final long currentTime = TimeUtil.currentTimeMillis();
+        final long maxTime = currentTime - currentTime % 1000;
+        final long minTime = maxTime - 1000;
```
jasonjoo2010, line 65: The time range could be passed in to reuse and keep the consistency in the same loop.

jasonjoo2010 (changes requested):

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
brotherlu-xcq: Yes, I will remove the extra space.

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/MetricExporterInit.java
brotherlu-xcq: Yes, Now it can be kept into `JMXMetricExporter` directly instead of though constructor.

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/collector/MetricCollector.java
brotherlu-xcq: Ok, I will use the time range as a argument in `getLastMetrics`

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/exporter/jmx/MBeanRegistry.java
brotherlu-xcq: actually, we only use the first map, so I think remove other one is OK. but if we put the beanName into `MetricBean` it will be displayed in jmx, maybe the unused filed will confused user.

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/exporter/jmx/MetricBean.java
brotherlu-xcq: I will keep it same with set.

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/exporter/jmx/MetricBeanWriter.java
brotherlu-xcq: This scene should not occur often, I think we can removed it from registry instead of reset it.

brotherlu-xcq: but, I still need the `existResouce` to know the MetricBean which need to be removed.

jasonjoo2010: Yeah. By the version comparison design I mean you can use a version property to identify which are updated in this round. Example:

```
A{ver: 1}
B{ver: 1}
C{ver: 1}
```

After update
```
A{ver: 2}
B{ver: 2}
C{ver: 1}
```

Then you know C is skipped in the previous round. And it should be removed. By doing this no extra data structure is needed. And you can also use timestamp as the version instead of adding a new one.

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/exporter/jmx/MBeanRegistry.java
jasonjoo2010: OK to me then. Any different suggestions, @sczyh30 ?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleManager.java
jasonjoo2010: Oh, I see the latest change on spaces. But how about the reordering on the imports?

sentinel-extension/sentinel-metric-exporter/src/main/java/com/alibaba/csp/sentinel/metric/exporter/jmx/MetricBeanWriter.java
brotherlu-xcq: Thank you so much, I will try to change it as your way.

jasonjoo2010 (approved):
LGTM

Thanks for your patience and great contribution.

sczyh30 (approved):
LGTM

sczyh30:
Great. Thanks for contributing! 🎉

Issue #2327, opened 2021-08-11
Remove redundant functional interfaces


hunjipo:
## Issue Description

In sentinel-core project, there are many functional interfaces, such as `Function`, `Consumer`, `BiConsumer`, etc. they are all copied jdk8. I guess these functional interfaces were created because sentinel used a JDK version lower than 8, but we wanted to use functional interfaces. But now, sentinel depends on the JDK version. The minimum requirement is 8. Therefore, we can delete these interfaces and use the interfaces provided by JDK.

### How to reproduce it (as minimally and precisely as possible)

1. 
2. 
3.

hunjipo:
@sczyh30 bro, if you agree with me, can I mention a PR to delete these interfaces?

sczyh30:
Contributions are welcomed.

PR #2328, opened 2021-08-11
Remove redundant functional interfaces

hunjipo:
See issues #2327.
See issues #2327.


hunjipo:
@sczyh30 An abstract class `AbstractApiMatcher` implements the `Predicate` of sentinel. Should I also change it to the `Predicate` of JDK?

PR #2342, opened 2021-08-18
Adding an utility method clear() for convenience

Roger3581321:
It is convenient to use clear() to remove all of current concurrency.

No.

### Describe how you did it
Called ConcurrentHashMap's clear() method straightforward. 

### Describe how to verify it
Call the clear() method and check the result.

### Special notes for reviews
None.


sczyh30:
Could you please illustrate the scenario? It seems not necessary to export this method.

Roger3581321:
> Could you please illustrate the scenario? It seems not necessary to export this method.
Calling clear method to clear all the mappings rather than calling remove method to remove the mapping one by one. It may be useful to prevent memory leak in some cases.

Issue #1906, opened 2020-12-16
热点参数集群限流,如果参数是一个对象,因为netty在encode时会过滤掉,在tokenserver那decode时data会是null,导致npe


zuoqi1993125:
## Issue Description

热点参数集群限流,如果参数是一个对象,因为netty在encode时会过滤掉,在tokenserver那解码时data会是null,导致空指针

### Describe what happened (or what feature you want)
我在使用@sentinelresource注解时,使用了对象为参数,测试热点集群限流,希望可以总体阈值生效,但是metric日志总是不生效

### How to reproduce it (as minimally and precisely as possible)

![image](https://user-images.githubusercontent.com/33927898/102344597-181cef00-3fd7-11eb-8c0c-a28d348c1c44.png)

### Tell us your environment
jdk1.8
sentinel1.8

### Anything else we need to know?
编码时跳过参数
```java
int calculateParamTransportSize(Object value) {
        if (value == null) {
            return 0;
        }
        // Layout for primitives: |type flag(1)|value|
        // size = original size + type flag (1)
        if (value instanceof Integer || int.class.isInstance(value)) {
            return 5;
        } else if (value instanceof String) {
            // Layout for string: |type flag(1)|length(4)|string content|
            String tmpValue = (String) value;
            byte[] tmpChars = tmpValue.getBytes();
            return 1 + 4 + tmpChars.length;
        } else if (boolean.class.isInstance(value) || value instanceof Boolean) {
            return 2;
        } else if (long.class.isInstance(value) || value instanceof Long) {
            return 9;
        } else if (double.class.isInstance(value) || value instanceof Double) {
            return 9;
        } else if (float.class.isInstance(value) || value instanceof Float) {
            return 5;
        } else if (byte.class.isInstance(value) || value instanceof Byte) {
            return 2;
        } else if (short.class.isInstance(value) || value instanceof Short) {
            return 3;
        } else {
            // Ignore unexpected type.
            return 0;
        }
```
解码时
```java
public class ParamFlowRequestDataDecoder implements EntityDecoder<ByteBuf, ParamFlowRequestData> {

    @Override
    public ParamFlowRequestData decode(ByteBuf source) {
        if (source.readableBytes() >= 16) {
            ParamFlowRequestData requestData = new ParamFlowRequestData()
                .setFlowId(source.readLong())
                .setCount(source.readInt());
           
            int amount = source.readInt();
            //因为amount为0,所以跳出判断,返回一个null
[...]

PR #2546, opened 2022-01-19
fix bug https://github.com/alibaba/Sentinel/issues/1906

howiekang:
修复了 #1906

在Cluster客户端 进行参数校验，不是基本类型的参数 会抛出异常，不再发送到Cluster Server

 #1906

### DIFF
--- a/sentinel-cluster/sentinel-cluster-client-default/src/main/java/com/alibaba/csp/sentinel/cluster/client/NettyTransportClient.java
+++ b/sentinel-cluster/sentinel-cluster-client-default/src/main/java/com/alibaba/csp/sentinel/cluster/client/NettyTransportClient.java
@@ -16,16 +16,11 @@
 package com.alibaba.csp.sentinel.cluster.client;
 
 import java.util.AbstractMap.SimpleEntry;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-import java.util.Objects;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.stream.Collectors;
 
 import com.alibaba.csp.sentinel.cluster.ClusterErrorMessages;
 import com.alibaba.csp.sentinel.cluster.ClusterTransportClient;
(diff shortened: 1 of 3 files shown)


brotherlu-xcq (changes requested):

sentinel-cluster/sentinel-cluster-client-default/src/main/java/com/alibaba/csp/sentinel/cluster/client/NettyTransportClient.java
```diff
@@ -196,7 +202,7 @@ public void stop() throws Exception {
     }
 
     private boolean validRequest(Request request) {
-        return request != null && request.getType() >= 0;
+        return request != null && request.getType() >= 0 ;
```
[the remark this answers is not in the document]
howiekang: I changed it.

brotherlu-xcq (dismissed):
LGTM

sentinel-cluster/sentinel-cluster-client-default/src/main/java/com/alibaba/csp/sentinel/cluster/client/NettyTransportClient.java
```diff
@@ -212,6 +218,10 @@ public ClusterResponse sendRequest(ClusterRequest request) throws Exception {
         if (!validRequest(request)) {
             throw new SentinelClusterException(ClusterErrorMessages.BAD_REQUEST);
         }
+        // fix bug https://github.com/alibaba/Sentinel/issues/1906
+        if (!checkParamFlowData(request)){
```
sczyh30, line 222: `NettyTransportClient` is designed as the transport-layer implementation, which is not associated to the concrete request type (and content). Maybe adding the logic in `DefaultClusterTokenClient` seems better?

sczyh30 (changes requested):

sentinel-cluster/sentinel-cluster-client-default/src/main/java/com/alibaba/csp/sentinel/cluster/client/NettyTransportClient.java
howiekang: Moved `checkParamFlowData` to `DefaultClusterTokenClient`. The logic of `checkParamFlowData` has not changed. However, if there are other `ClusterTokenClient` implementations in the future, the `checkParamFlowData` logic cannot be reused.

PR #2562, opened 2022-02-09
Feature/jdbc-flow-control, add JDBC adapter

icodening:
新增jdbc适配模块，用户可以使用Sentinel对jdbc流量进行控制。

Fixes#2553 

### Describe how you did it
包装JDBC组件，如DataSource、Connection、Statement。
在Statement execute执行SQL前记录SQL，或Connection在构造PrepareStatement时记录SQL，在execute时使用Sentinel进行资源控制，SQL statement作为资源，例如 "select id,name,age from student where id = ?" 就是一个Sentinel的资源.

### Describe how to verify it
junit test

### Special notes for reviews
该方式实现思路简单且代理的是JDBC组件，可以直接适配主流框架如JPA、MyBatis。例子如下:
````java
DataSource sentinelJDBCDataSource = new SentinelJDBCDataSource(originalDataSource);
Connection connection = sentinelJDBCDataSource.getConnection();
//...............operate your db................
````

### CODE REVIEW

File: sentinel-adapter/sentinel-jdbc-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/jdbc/JDBCContext.java [hunk 1]
```diff
@@ -0,0 +1,32 @@
+package com.alibaba.csp.sentinel.adapter.jdbc;
+
+
+/**
+ * @author icodening
+ * @date 2022.02.09
+ */
+public interface JDBCContext {
```
sczyh30, line 8: Could you please add some javadocs for the interface (and methods)?

sentinel-adapter/sentinel-jdbc-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/jdbc/JDBCContext.java
```diff
@@ -0,0 +1,32 @@
+package com.alibaba.csp.sentinel.adapter.jdbc;
```
sczyh30, line 1: Please prepend the license header here.

sentinel-adapter/sentinel-jdbc-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/jdbc/AbstractSentinelJDBCComponent.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    protected <E> E sentinelEntryInternal(SQLExceptionCallable<E> callable, boolean clearAttachmentsOnComplete) throws SQLException {
+        Entry entry = null;
+        try {
+            String resourceName = getResourceName();
+            if (StringUtil.isBlank(resourceName)) {
+                return callable.call();
+            }
+            ContextUtil.enter(JDBC_CONTEXT_NAME, "");
```
sczyh30, line 80: Actually the DB access should not indicate the entrance of the call stack, so maybe entrance context is not needed here.

sczyh30 (changes requested):

sentinel-adapter/sentinel-jdbc-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/jdbc/SentinelJDBCStatement.java
```diff
@@ -0,0 +1,116 @@
...
+public class SentinelJDBCStatement extends StatementDelegate {
+
+    public SentinelJDBCStatement(Statement delegate) {
+        super(delegate);
+    }
+
+    @Override
+    public ResultSet executeQuery(String sql) throws SQLException {
+        setSQL(sql);
+        return sentinelEntryInternal(() -> super.executeQuery(sql), true);
```
sczyh30, line 37: For normal JDBC statement, it might bring a large number of different SQLs, which generates too much Sentinel "resource" and cause huge memory footprint. Maybe we need some mechanism to avoid it.

sentinel-adapter/sentinel-jdbc-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/jdbc/JDBCContext.java
icodening: OK, i will slove it

sentinel-adapter/sentinel-jdbc-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/jdbc/JDBCContext.java
icodening: > Please prepend the license header here.

OK, i will slove it

sentinel-adapter/sentinel-jdbc-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/jdbc/SentinelJDBCStatement.java
icodening: > For normal JDBC statement, it might bring a large number of different SQLs, which generates too much Sentinel "resource" and cause huge memory footprint. Maybe we need some mechanism to avoid it.

I have to think about this problem, maybe we can add a custom resource method, and use SQL when the custom resource cannot be obtained. example:
````java
protected String getResourceName() {
    String currentResourceName = JDBCResourceUtil.getCurrentThreadJDBCResourceName();
    if (StringUtil.isNotBlank(currentResourceName)) {
        return currentResourceName;
    }
    return getSQL();
}
````
However, when doing this, it is also necessary to adapt JPA and MyBatis, the "resource" needs to use repoInterface.methodName/namespace+id(in MyBatis), such as: com.alibaba.repository.UserRepository.saveUser. MyBatis interceptor example:
````java
public Object intercept(Invocation invocation) throws Throwable {
    MappedStatement ms = getMappedStatement(invocation);
    String resource = ms.getId(); //namespace+sqlName
    JDBCResourceUtil.setCurrentThreadJDBCResourceName(resource);
    return invocation.proceed();
}
````
How about this way?

icodening:
I used Apache Calcite to parse sql and replace parameters to avoid memory consumption as much as possible.
please review @sczyh30

PR #2709, opened 2022-05-07
system rules support ibm J9

lyh200:
Now system metrics are collected from com.sun.management.OperatingSystemMXBean for HotSpot JVM, but Ibm J9 is in com.ibm.lang.management.OperatingSystemMXBean, so need to deduce the target source.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/OperatingSystemBeanManager.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/OperatingSystemBeanManager.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 1999-2018 Alibaba Group Holding Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.alibaba.csp.sentinel.slots.system;
+
+import com.alibaba.csp.sentinel.log.RecordLog;
+import com.alibaba.csp.sentinel.util.MethodUtil;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * OperatingSystemBeanManager related
+ *
+ * @author yanhom
+ */
+public class OperatingSystemBeanManager {
+
+    private OperatingSystemBeanManager() {}
+
+    /**
+     * com.ibm for J9
+     * com.sun for HotSpot
+     */
+    private static final List<String> OPERATING_SYSTEM_BEAN_CLASS_NAMES = Arrays.asList(
+            "com.sun.management.OperatingSystemMXBean", "com.ibm.lang.management.OperatingSystemMXBean");
+
+    private static final OperatingSystemMXBean OPERATING_SYSTEM_BEAN;
+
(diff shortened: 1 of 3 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/MethodUtil.java
```diff
@@ -76,4 +76,21 @@ public static String resolveMethodName(Method method) {
...
+        } catch (Exception e) {
+            return Double.NaN;
+        }
+    }
+
+    public static long invokeAndReturnLong(Method method, Object targetObj) {
+        try {
+            return method != null ? (long) method.invoke(targetObj) : -1;
+        } catch (Exception e) {
+            return -1;
```
brotherlu-xcq, line 92: maybe we need print the exeception?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/MethodUtil.java
```diff
@@ -76,4 +76,21 @@ public static String resolveMethodName(Method method) {
     static void clearMethodMap() {
         methodNameMap.clear();
     }
+
+    public static double invokeAndReturnDouble(Method method, Object targetObj) {
+        try {
+            return method != null ? (double) method.invoke(targetObj) : Double.NaN;
+        } catch (Exception e) {
+            return Double.NaN;
```
brotherlu-xcq, line 84: maybe we need print the exeception?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/OperatingSystemBeanManager.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private static Method deduceMethod(String name) {
+        if (Objects.isNull(OPERATING_SYSTEM_BEAN_CLASS)) {
+            return null;
+        }
+        try {
+            OPERATING_SYSTEM_BEAN_CLASS.cast(OPERATING_SYSTEM_BEAN);
+            return OPERATING_SYSTEM_BEAN_CLASS.getDeclaredMethod(name);
+        } catch (Exception e) {
+            return null;
```
brotherlu-xcq, line 93: maybe we need print the exeception?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/OperatingSystemBeanManager.java
```diff
@@ -0,0 +1,96 @@
...
+public class OperatingSystemBeanManager {
+
+    private OperatingSystemBeanManager() {}
+
+    /**
+     * com.ibm for J9
+     * com.sun for HotSpot
+     */
+    private static final List<String> OPERATING_SYSTEM_BEAN_CLASS_NAMES = Arrays.asList(
+            "com.sun.management.OperatingSystemMXBean", "com.ibm.lang.management.OperatingSystemMXBean");
```
brotherlu-xcq, line 42: I think write the target class here is not a good way to extends other JVM. cloud you provide a common way to introduce the `OperationSystemMXBean`?

brotherlu-xcq (changes requested):

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/OperatingSystemBeanManager.java
lyh200: done!

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/system/OperatingSystemBeanManager.java
lyh200: The difference only in the bean class name, treat as a common constant, I think there is no need to design a very abstract

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/MethodUtil.java
lyh200: done

sentinel-core/src/main/java/com/alibaba/csp/sentinel/util/MethodUtil.java
lyh200: done

yanhom1314:
Can this PR be merged?

PR #2735, opened 2022-05-31
Add append rules

tain198127:
Add two method to append and delete rule of degrade,flow and authority

### Describe how you did it

Add two method append and delete in each of FlowRuleManager, AuthorityRuleManager and DegradeRuleManager

### Describe how to verify it

image
It's submit some unittest for those method.


sczyh30:
Could you please illustrate your design in PR description?

Issue #1171, opened 2019-11-20
sentinel相关埋点对象内存占用的问题


zcwdlove:
### Describe what happened (or what feature you want)
线上发布应用后查看堆栈信息发现sentinel相关对象占用太高, 项目中除了测试用例几乎没有埋点信息，请问如何优化以减少内存占用？
详情见图:
http://ww1.sinaimg.cn/large/e6ed9ff1gy1g94u9b6ig5j20lz0c10t6.jpg
### Tell us your environment
1.5.2

sczyh30:
This might be caused by inappropriate metric item generating in `StatisticNode`:

https://github.com/alibaba/Sentinel/blob/d45ceb06dcc61356adc920772e142b81c1114f19/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/StatisticNode.java#L121

Every second Sentinel will aggregate metrics of every resource and write to the metric log, but when aggregating the metric of last second, it actually generates metric items in the past minute (i.e. 60s), which is not necessary. This could be improved, but the handling logic of fetchTime should be considered.

Discussions and contributions are welcomed!

sczyh30:
Since 1.7.0 we've provided a useful method `rawMetricsInMin` in `Node` interface, which could help us to improve the code:

https://github.com/alibaba/Sentinel/blob/d45ceb06dcc61356adc920772e142b81c1114f19/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/StatisticNode.java#L136-L138

The legacy method `metrics` could be deprecated.

xzy-max1:
Has this problem been solved?

PR #2798, opened 2022-07-26
Improve performance of StatisticNode#metrics.(#1171)

nathanchoui:
Every second Sentinel will aggregate metrics of every resource and write to the metric log, but when aggregating the metric of last second, it actually generates metric items in the past minute , which is not necessary. 

issue 1171

### Describe how you did it
I modified StatisticNode#metrics. It only aggregates the metrics within flush interval.  It would be reduce unnecessary metric items

### Describe how to verify it
Verify the function by running the unit test in StatisticNodeTests#testMetrics


nathanchoui (commented):
I have reviewed the code, and the unit test is passed.

PR #2802, opened 2022-07-26
Improve performance of StatisticNode#metrics. issue 1171

nathanchoui:
Every second Sentinel will aggregate metrics of every resource and write to the metric log, but when aggregating the metric of last second, it actually generates metric items in the past minute , which is not necessary.

issue 1171

Describe how you did it
I modified StatisticNode#metrics. It only aggregates the metrics within flush interval. It would be reduce unnecessary metric items

Describe how to verify it
Verify the function by running the unit test in StatisticNodeTests#testMetrics

Special notes for reviews

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/StatisticNode.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/StatisticNode.java
@@ -20,6 +20,7 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.LongAdder;
 
+import com.alibaba.csp.sentinel.config.SentinelConfig;
 import com.alibaba.csp.sentinel.node.metric.MetricNode;
 import com.alibaba.csp.sentinel.slots.statistic.metric.ArrayMetric;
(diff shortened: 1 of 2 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/StatisticNode.java
```diff
@@ -115,10 +116,15 @@ public class StatisticNode implements Node {
     @Override
     public Map<Long, MetricNode> metrics() {
         // The fetch operation is thread-safe under a single-thread scheduler pool.
-        long currentTime = TimeUtil.currentTimeMillis();
-        currentTime = currentTime - currentTime % 1000;
+        final long currentTime = TimeUtil.currentTimeMillis();
+        final long endTime = currentTime - currentTime % 1000;
+        long flushIntervalSec = SentinelConfig.metricLogFlushIntervalSec();
```
sczyh30, line 121: The `StatisticNode` should not be aware of outside "metric log". Maybe you could refactor `MetricTimerListener` with `StatisticNode#rawMetricsInMin`?

sczyh30 (changes requested):

PR #2822, opened 2022-08-10
Fix Dubbo spi path

AlbumenJ:

Fix https://github.com/apache/dubbo-sentinel-support/issues/5


sczyh30 (approved):
LGTM

PR #2868, opened 2022-09-07
Removing Unnecessary Checks

PepoRobert:

Removing Unnecessary Checks     Condition 'Constants.ENTRY_NODE == null' is always 'false'  and Other Constants.ENTRY_NODE  references are not checked
### Describe how you did it

read code


PepoRobert:
re pr

PR #2869, opened 2022-09-07
Removing unnecessary checks in SystemRuleManager#checkSystem

PepoRobert:
Removing Unnecessary Checks

Removing Unnecessary Checks Condition 'Constants.ENTRY_NODE == null' is always 'false' and Other Constants.ENTRY_NODE references are not checked


sczyh30 (approved):
LGTM

sczyh30:
Thanks for contributing!

PR #2891, opened 2022-09-26
Fix the sysrule

huangyiminghappy:
All rules, including flowrule,degradeRule rules and even AbstractRule, are constructed using  builder pattern to set attributes,In order to unify the use of rules and code style, we should set the SystemRule

### Describe how you did it
 builder pattern to set attributes in SystemRule 

### Describe how to verify it
set attribute


sczyh30:
Hi, thanks for contributing. This may lead to breaking changes (of API) and we might have to be careful of backward compatibility.

huangyiminghappy:
> 

thank you reply,we just add builder pattern in set method,and return it self replace no return , Theoretically it can not  lead to breaking changes (of API)

Issue #2866, opened 2022-09-06
feature: 配置集群限流时候,建议有个单机的阈值兜底


brucelwl:
feature: 配置集群限流时候,建议有个单机的阈值兜底. 目前的如果选择集群限流后, count是集群限流的阈值, 如果出现降级, 使用的是集群限流的阈值, 建议加一个单机阈值, 降级时候使用单机阈值,而不是集群阈值

azhsmesos:
Please assigned it to me，i want to try it，thanks!

sczyh30:
> Please assigned it to me，i want to try it，thanks!

@azhsmesos Looking forward to your PR :)

azhsmesos:
thanks

naget:
@sczyh30 I made a pr, could you give me a review

PR #2919, opened 2022-10-22
add localCount for fallback when clusterMode failed

naget:
#2866 

### Describe how you did it
add a new configuration item on the dashboard, use it as the flow control threshold count when cluster flow restriction degenerate to local flow restriction


sczyh30:
Hi, could you please resolve conflicts here?

PR #2962, opened 2022-11-21
Fix issue: connection leak risk when using Nacos data-source.

### LINKED ISSUE #2953: Connection leak risk when using Nacos data-source | Nacos 数据源 close 存在连接泄漏问题
## Issue Description

Type: *bug report*

when `com.alibaba.csp.sentinel.datasource.nacos.NacosDataSource ` init, it will create a new `ConfigService` and cached it.
But when NacosDataSource close, the `ConfigService` is not be shutdown so that the connections can threads may be not closed.

Once `com.alibaba.csp.sentinel.datasource.nacos.NacosDataSource ` rebuild, the connection and thread will leak.

shutdown `ConfigService` when `com.alibaba.csp.sentinel.datasource.nacos.NacosDataSource ` close.

The other and recommended fixed way is re-use the `ConfigService`  for all datasources. Now sentinel will create one connection to nacos for each rule. 
In fact, multiple rule can be listened by one `ConfigService`.

pandaapo:
Connection in `NacosDataSource` exists leak risk.

Fixes #2953

### Describe how you did it
`ConfigService#shutDown()` when `NacosDataSource#close()`.

### DIFF
--- a/sentinel-extension/sentinel-datasource-nacos/src/main/java/com/alibaba/csp/sentinel/datasource/nacos/NacosDataSource.java
+++ b/sentinel-extension/sentinel-datasource-nacos/src/main/java/com/alibaba/csp/sentinel/datasource/nacos/NacosDataSource.java
@@ -145,6 +145,7 @@ public class NacosDataSource<T> extends AbstractDataSource<String, T> {
     public void close() {
         if (configService != null) {
             configService.removeListener(dataId, groupId, configListener);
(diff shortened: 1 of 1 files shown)


sczyh30 (approved):
LGTM

sczyh30:
Thanks for contributing!

For better solution, we may consider reusing the single `ConfigService` for all `NacosDataSource`.

Issue #2986, opened 2022-12-06
Add support for Spring Boot 3.x (Spring 6.x)


sczyh30:
## Issue Description

Type: *feature request*

### Describe what happened (or what feature you want)

Add support for Spring Boot 3.x (Spring 6.x). The following adapters might need to be checked or updated:

* [sentinel-spring-webmvc-adapter](https://github.com/alibaba/Sentinel/tree/master/sentinel-adapter/sentinel-spring-webmvc-adapter)
* [sentinel-zuul-adapter](https://github.com/alibaba/Sentinel/tree/master/sentinel-adapter/sentinel-zuul-adapter)
* [sentinel-spring-webflux-adapter](https://github.com/alibaba/Sentinel/tree/master/sentinel-adapter/sentinel-spring-webflux-adapter)
* [sentinel-spring-cloud-gateway-adapter](https://github.com/alibaba/Sentinel/tree/master/sentinel-adapter/sentinel-spring-cloud-gateway-adapter)

Also refer #2998 for raw Jakarta EE support.

JavaLionLi:
期待啊

LiYangSir:
I can try it

Prince-1110:
If this issue is not resolved yet, can I work on it?

SahanSamarakoon:
@sczyh30 Is this issue still open ?

saikrishnasaski:
Is this issue still open?

ilaotan:
Is this issue still open?

mhraza95:
Hi @sczyh30 , If this issue still open, can i work on this?

PR #2997, opened 2022-12-13
Add sentinel-webmvc-6x-adapter module for Spring Boot 3 (Spring 6.x)

ruansheng8:

Support `spring-webmvc 6.x` and `JDK17`

Related to #2986

### DIFF
--- a/sentinel-adapter/sentinel-spring-webmvc-6x-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/spring/webmvc/AbstractSentinelInterceptor.java
+++ b/sentinel-adapter/sentinel-spring-webmvc-6x-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/spring/webmvc/AbstractSentinelInterceptor.java
@@ -1,0 +1,200 @@
+/*
+ * Copyright 1999-2019 Alibaba Group Holding Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.alibaba.csp.sentinel.adapter.spring.webmvc;
+
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+
+import com.alibaba.csp.sentinel.Entry;
+import com.alibaba.csp.sentinel.EntryType;
+import com.alibaba.csp.sentinel.ResourceTypeConstants;
+import com.alibaba.csp.sentinel.SphU;
+import com.alibaba.csp.sentinel.Tracer;
+import com.alibaba.csp.sentinel.adapter.spring.webmvc.config.BaseWebMvcConfig;
+import com.alibaba.csp.sentinel.context.ContextUtil;
+import com.alibaba.csp.sentinel.log.RecordLog;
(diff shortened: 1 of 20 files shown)


sczyh30:
cc @LiYangSir @HaojunRen for review

HaojunRen:
I suggest  to add a new branch to support Spring Boot 3.0
1. NOT enough to only add a sentinel-webmvc-6x-adapter module
2. It is not so friendly to user to upgrade seamless, they should change their jar name

LiYangSir:
I admit that adding a new module to springboot3.x alone is indeed an effective solution. The core change point should be that the javax package is replaced by jakarta. But I feel that adding a new module will have a lot of duplicate code.
Is it possible to make a warpper adapter? The webmvc module does not depend on javax or jakarta. Instead, it depends on our wrapper adapter. This way you don't need to add a new module.

ruansheng8:
I prefer to create a new branch to solve it. Adding `sentinel-webmvc-6x-adapter` module is a temporary solution discussed in the dingtalk group yesterday to support the upgrade of other projects.

sentinel-adapter/sentinel-spring-webmvc-6x-adapter/src/test/java/com/alibaba/csp/sentinel/adapter/spring/webmvc/config/InterceptorConfig.java
```diff
@@ -0,0 +1,92 @@
...
+    private void addSpringMvcInterceptor(InterceptorRegistry registry) {
+        //Config
+        SentinelWebMvcConfig config = new SentinelWebMvcConfig();
+
+        config.setBlockExceptionHandler(new BlockExceptionHandler() {
+            @Override
+            public void handle(HttpServletRequest request, HttpServletResponse response, BlockException e) throws Exception {
+                String resourceName = e.getRule().getResource();
+                //Depending on your situation, you can choose to process or throw
+                if ("/hello".equals(resourceName)) {
```
complone, line 56: Is it possible to avoid hardcoding?

sentinel-adapter/sentinel-spring-webmvc-6x-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/spring/webmvc/AbstractSentinelInterceptor.java
```diff
@@ -0,0 +1,200 @@
...
+            
+            if (increaseReferece(request, this.baseWebMvcConfig.getRequestRefName(), 1) != 1) {
+                return true;
+            }
+            
+            // Parse the request origin using registered origin parser.
+            String origin = parseOrigin(request);
+            String contextName = getContextName(request);
+            ContextUtil.enter(contextName, origin);
+            Entry entry = SphU.entry(resourceName, ResourceTypeConstants.COMMON_WEB, EntryType.IN);
```
complone, line 105: This slot is mainly responsible for collecting resource paths and storing the call paths of these resources. The call result entry may be null. Do you need to add the following code?
```
  if (entry!= null) {
    entry. exit();
 }
```

sentinel-adapter/sentinel-spring-webmvc-6x-adapter/src/test/java/com/alibaba/csp/sentinel/adapter/spring/webmvc/config/InterceptorConfig.java
ruansheng8: This is unit test code and this module code was copied from the `sentinel-spring-webmvc-adapter` module for temporary compatibility.

sentinel-adapter/sentinel-spring-webmvc-6x-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/spring/webmvc/AbstractSentinelInterceptor.java
ruansheng8: The code in this branch is always based on the `sentinel-spring-webmvc-adapter` module and only performs compatibility modifications.

PR #2999, opened 2022-12-14
Fix log print reformat

hncboy:
1.Modify some log level: change the log level of some exceptions from info to warn
2.Modify the print format of some logs: replace string splicing with placeholders.

NONE

### DIFF
--- a/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/init/HeartbeatSenderInitFunc.java
+++ b/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/init/HeartbeatSenderInitFunc.java
@@ -73,12 +73,12 @@ public class HeartbeatSenderInitFunc implements InitFunc {
         Long intervalInConfig = TransportConfig.getHeartbeatIntervalMs();
         if (isValidHeartbeatInterval(intervalInConfig)) {
             RecordLog.info("[HeartbeatSenderInitFunc] Using heartbeat interval "
-                + "in Sentinel config property: " + intervalInConfig);
+                    + "in Sentinel config property: {}", intervalInConfig);
             return intervalInConfig;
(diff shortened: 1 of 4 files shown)


sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/init/HeartbeatSenderInitFunc.java
```diff
@@ -73,12 +73,12 @@ long retrieveInterval(/*@NonNull*/ HeartbeatSender sender) {
         Long intervalInConfig = TransportConfig.getHeartbeatIntervalMs();
         if (isValidHeartbeatInterval(intervalInConfig)) {
             RecordLog.info("[HeartbeatSenderInitFunc] Using heartbeat interval "
-                + "in Sentinel config property: " + intervalInConfig);
+                    + "in Sentinel config property: {}", intervalInConfig);
```
jnan806, line 76: since to `replace string splicing with placeholders`, why not delete the `plus symbol` incidentally 😃

sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/transport/init/HeartbeatSenderInitFunc.java
```diff
@@ -73,12 +73,12 @@ long retrieveInterval(/*@NonNull*/ HeartbeatSender sender) {
...
             return intervalInConfig;
         } else {
             long senderInterval = sender.intervalMs();
             RecordLog.info("[HeartbeatSenderInit] Heartbeat interval not configured in "
-                + "config property or invalid, using sender default: " + senderInterval);
+                    + "config property or invalid, using sender default: {}", senderInterval);
```
jnan806, line 81: same above.

Issue #3015, opened 2023-01-05
Fix test failure under JDK 17 | 修复 JDK 17 下测试框架及用例报错的问题


sczyh30:
## Issue Description

Type: *bug report*

### Describe what happened (or what feature you want)

The test cases will fail under JDK 17, due to JDK-compatibility of some test frameworks (e.g. PowerMock).

JDK 17 下测试用例会失败，部分原因是我们测试框架不支持 JDK 17 导致的（可能是版本太低，或确实不支持）。欢迎社区修复解决。

Some error messages:

```
com.alibaba.csp.sentinel.slots.statistic.base.LeapArrayTest

java.lang.ExceptionInInitializerError
	at jdk.internal.reflect.GeneratedSerializationConstructorAccessor4.newInstance(Unknown Source)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
	at org.objenesis.instantiator.sun.SunReflectionFactoryInstantiator.newInstance(SunReflectionFactoryInstantiator.java:48)
	at org.objenesis.ObjenesisBase.newInstance(ObjenesisBase.java:73)
	at org.mockito.internal.creation.instance.ObjenesisInstantiator.newInstance(ObjenesisInstantiator.java:19)
	at org.mockito.internal.creation.bytebuddy.SubclassByteBuddyMockMaker.createMock(SubclassByteBuddyMockMaker.java:47)
	at org.mockito.internal.creation.bytebuddy.ByteBuddyMockMaker.createMock(ByteBuddyMockMaker.java:25)
	at org.powermock.api.mockito.mockmaker.PowerMockMaker.createMock(PowerMockMaker.java:41)
	at org.mockito.internal.util.MockUtil.createMock(MockUtil.java:35)
	at org.mockito.internal.MockitoCore.mock(MockitoCore.java:62)
	at org.mockito.Mockito.mock(Mockito.java:1908)
	at org.powermock.api.mockito.internal.mockcreation.DefaultMockCreator.createMethodInvocationControl(DefaultMockCreator.java:108)
	at org.powermock.api.mockito.internal.mockcreation.DefaultMockCreator.doCreateMock(DefaultMockCreator.java:61)
	at org.powermock.api.mockito.internal.mockcreation.DefaultMockCreator.createMock(DefaultMockCreator.java:53)
[...]

PR #3028, opened 2023-02-02
add java17 support for test

xiangtianyu:
Do some fix to support jdk17 test

related to #3015 

### Describe how you did it

modify pom.xml to add "--add-opens" to jvmargs when jdk version is large than 8

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -278,16 +278,6 @@
                 </configuration>
             </plugin>
             <plugin>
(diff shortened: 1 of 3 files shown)


sczyh30:
Hi, our test cases should pass **without** `--add-opens` args under JDK 11/17.

xiangtianyu:
> Hi, our test cases should pass **without** `--add-opens` args under JDK 11/17.

emmm, i got it, but why?

Issue #263, opened 2018-11-26
Expand more statistic dimensions( such as 20s, 1min, 20min), not only for 1 second


ro9er:
## Issue Description

Type: *bug report* or *feature request*

### Describe what happened (or what feature you want)
In some situation, we want to perform flow control not only based on the statistics of 1 second. For example, as for some origin/callers, we wanna count the call times in 1 minute and block the caller if the count exceeds the threshold. Sentinel does not support this kind of statistics.

### Describe what you expected to happen
Sentinel supports more statistic dimension of unit. It would be perfect if programmers can build the statistic rule by setting count parameters.

### How to reproduce it (as minimally and precisely as possible)

1. 
2. 
3.

sczyh30:
This is a good idea giving users more flexible strategy. Any thoughts on its implementation?

ro9er:
Maybe one way to implement that is make one resource contains N clusterNode, which correspond to one time dimension rule. For example, if one resource has 4 rules and the statistic unit is 1s 1s 20s 20min, the resource will have 3 cluster Node. And we perform flow control on that.
Another way is that we use just one clusterNode based on 1s like now. And if we want to support more statistic dimension, we log the statistic per second by retrieving the clusternode per second or on requests entering.

Frankzhaopku:
This issue has seen no activity in 15 days, it has been automatically closed for house-keeping purpose.

PR #3106, opened 2023-04-25
Add implementation of token bucket for flow-control

LearningGp:

This PR adds a simple implementation of the token bucket algorithm.
The biggest advantage of sentinel's sliding window data structure is that it can handle some extreme high concurrency scenarios, so we pay some performance costs, but in some scenarios, we need more statistical dimensions and do not care about extreme high concurrency scenarios. In these scenarios, the implementation of a lightweight token bucket algorithm may be a better choice.

This PR is a pre-work to fix the following issues:
#263
#2231

### Describe how you did it

In the implementation, the number of tokens in the token bucket is not continuously updated, it will be calculated at the time of acquisition, thereby reducing the overall performance costs.

### Describe how to verify it

Run the test cases.


sczyh30 (approved):
LGTM

sczyh30:
Thanks for contributing!

ShenJunkun:
Has the code in this PR been utilized yet? I couldn't find any exposed interfaces or public APIs for it.

PR #3243, opened 2023-10-19
make all ThreadPool static final

robberphex:

There are some thread pools associated with class instance, which might cause many thread pools created.

NONE

### Describe how you did it

make all thread poll is static final, do not create thread pool dynamically.

### Describe how to verify it

Start an application with sentinel, we could notice that all threads associated with an static field.


LearningGp (approved):
LGTM

PR #3251, opened 2023-10-27
resource rules (flow/degrade/param/authority) support regex matching

### LINKED ISSUE #3247: Resource rules support regular expression matching 【资源规则支持正则表达式匹配】
## Issue Description

Resource rules (flow, degrade, authority, hot param) support regex matching

Traditional resource rules strictly follow the string matching mechanism to obtain all rules under the current resource. Now if you want to enhance this capability, you can use regular expressions for matching. When users add/delete hidden interfaces, there is no need to manually add/delete rules.

传统的资源规则严格按照字符串匹配的机制获取当前资源下的所有规则，现希望增强该能力，可以使用正则表达式进行匹配。当用户增加/删除埋点接口时，无需手动增加/删除规则。

Applicable scene:
- Flow/hotparam rules: allows configuration of unified **backup rules**;
- Authorization/degrade rules: Allow users to **batch configure** rules;

适用场景：
- 流控 / 热点规则：允许配置统一的 **兜底规则**；
- 授权 / 降级规则：允许用户 **批量配置** 规则；

Example:
Configure flow control rules (resource name: .*, regex: true, QPS threshold: 5), 2 buried points resA, resB. Effect: resA, resB respectively limit QPS threshold = 5

举例：
配置流控规则（资源名称：.*, 支持正则: true，QPS 阈值：5），2 个埋点 resA, resB。效果：resA, resB 分别限制 QPS 阈值 = 5

### Describe your initial design (if present)
It has been initially implemented, and PR will be added in the future.

已经初步实现，后续会补充PR

### Additional context
Risk point: Performance issues with regular expression matching.

风险点：担心正则表达式匹配的性能问题。

Solution:
- Cache the results of regular expression matching;
- When reloading rules, the cache is rebuilt instead of cache invalidated to avoid the performance impact of relying on request reconstruction in high concurrency situations due to cache invalidation;

解决方案：
- 缓存正则表达式匹配的结果；
- 当重新 load 规则时缓存重建，而非缓存失效，避免由于缓存失效后，依赖请求重建在高并发情况下的性能影响；

LiYangSir:
Resource rules support matching resource names according to regular rules, allowing users to configure traffic limiting rules or batch configuration.
When adding/deleting interfaces, there is no need to manually create/delete rules.

资源规则支持按照正则匹配资源名称，允许用户按照配置兜底限流规则或批量配置的能力。
当新增/删除接口时，无需手动创建/删除规则。

Fixes #3247 

### Describe how you did it
Add a RuleManger to uniformly manage ordinary rules and regular rules. In order to reduce the performance loss caused by regular parsing/matching, the following two methods are used:
1. Add a cache to cache the current limiting rules after regular parsing. When matching the current limiting rules, first search from the cache;
2. Rebuild the cache. When the current limiting rule is updated, new cache data is rebuilt based on the old cache data to reduce the performance loss caused by cache failure after the rule is updated;

增加一个 RuleManger 统一管理普通规则和正则规则，为了减少正则解析/匹配所带来的性能损耗，使用下面两种方式：
1. 增加缓存，缓存正则解析后的限流规则，匹配限流规则时先从缓存中去查找；
2. 重建缓存，当限流规则出现更新时，依据旧缓存数据重建新的缓存数据，减少由于更新规则后，缓存失效带来的性能损耗；

### Describe how to verify it

Regular unit tests have been added to each of the four rules to ensure normal performance; if regular rules are turned on, current limiting rules that meet regular matching but have different resource names will take effect.

对四种规则分别增加了正则的单元测试，用于保证表现正常；如果开启了正则规则，满足正则匹配但资源名称不同的限流规则均会生效。
1. Authority Rule：com.alibaba.csp.sentinel.slots.block.authority.AuthorityPartialIntegrationTest
2. Degrade Rule：com.alibaba.csp.sentinel.slots.block.degrade.DegradePartialIntegrationTest
4. Flow Rule：com.alibaba.csp.sentinel.slots.block.flow.FlowPartialIntegrationTest
6. Hot Pram Rule：com.alibaba.csp.sentinel.slots.block.flow.param.ParamFlowPartialIntegrationTest

### Special notes for reviews
None

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/RuleManager.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/RuleManager.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright 1999-2018 Alibaba Group Holding Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.alibaba.csp.sentinel.slots.block;
+
+import com.alibaba.csp.sentinel.util.function.Function;
+import com.alibaba.csp.sentinel.util.function.Predicate;
+
+import java.util.*;
+import java.util.regex.Pattern;
+
+/**
+ * Unified rule management tool, mainly used for matching and caching of regular rules and simple rules.
+ * @author quguai
+ * @date 2023/10/9 20:35
+ */
+public class RuleManager<R> {
+
+    private Map<String, List<R>> originalRules = new HashMap<>();
+    private Map<Pattern, List<R>> regexRules = new HashMap<>();
+    private Map<String, List<R>> regexCacheRules = new HashMap<>();
+    private Map<String, List<R>> simpleRules = new HashMap<>();
+    private Function<List<R>, List<R>> generator = Function.identity();
+
+    private final Predicate<R> predicate;
+
+    public RuleManager() {
+        predicate = r -> r instanceof AbstractRule && ((AbstractRule) r).isRegex();
+    }
+
+    public RuleManager(Function<List<R>, List<R>> generator, Predicate<R> predicate) {
+        this.generator = generator;
+        this.predicate = predicate;
+    }
+
+    /**
+     * Update rules from datasource, split rules map by regex,
+     * rebuild the regex rule cache to reduce the performance loss caused by publish rules.
+     *
+     * @param rulesMap origin rules map
+     */
+    public void updateRules(Map<String, List<R>> rulesMap) {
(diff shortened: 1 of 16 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/RuleManager.java
```diff
@@ -0,0 +1,155 @@
+package com.alibaba.csp.sentinel.slots.block;
```
LearningGp, line 1: Could you please prepend the license header here?

sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/RuleManager.java
LearningGp, line 1: Could you please add unit tests for this class?

LearningGp (changes requested):

LearningGp:
It would be helpful to have a benchmark report comparing the original rules. :rocket:

LiYangSir:
> It would be helpful to have a benchmark report comparing the original rules. 🚀

### Throughput (s)
image

### Throughput (ns)
image

### AverageTime (ns)
image

> ps: testNormal: Original Type; testRegex: Direct use regex parse; testRegexWitchCache: use cache regex result

Due to the need to retrieve rules from the regex cache map and regular rules, the throughput of the previous retrieval method is twice that of direct retrieval, but it is also within an acceptable range.

LearningGp (approved):
LGTM

LearningGp:
Thanks for contributing! 🚀

PR #3262, opened 2023-11-09
Remove exception throwing in DefaultController

### LINKED ISSUE #3261: The canpass method of DeafultController cannot throw exceptions
## Issue Description

bug

If the current priority is true and the flow control mode is fast failure, exceptions should not be thrown, otherwise this priority setting is meaningless.

remove the exception throwing

1. Make sure your priority is set to true and the flow control mode is fail fast

isysc1:
…ast fail

If the current priority is true and the flow control mode is fast failure, exceptions should not be thrown, otherwise this priority setting is meaningless.

Fixes #3261

### Describe how you did it
I took the initiative to remove the exception throwing


isysc1:
[关联Issue #3261](https://github.com/alibaba/Sentinel/issues/3261)

PR #3283, opened 2023-12-01
httpclient-adapter bug

zgq101034:
GET  http://host:port/path?param1=value1
The result of "request.getRequestLine().getUri()" is "http://host:port/path?param1=value1",
while the result of "request.getURI().getPath()" is "http://host:port/path".


zgq101034:
你好！庄国乾已收到，谢谢！

LearningGp:
The CLA needs to be signed before the code is merged. :black_nib:

zhuangguoqian:
你好！庄国乾已收到，谢谢！

PR #3348, opened 2024-02-27
Support spring boot 3.x

robberphex:

* Add module sentinel-spring-webmvc-v6x-adapter, which hold package `com.alibaba.csp.sentinel.adapter.spring.webmvc_v6x`.
* Upload coverage reports to Codecov.


panxiaojun233 (dismissed):
LGTM

PR #3365, opened 2024-03-28
feat:Add a batch incremental Add rule for the Circuit Breaker manager

aliyx:
The gateway integrates sentinel and needs to initialize a circuit breaker for each resource (api+ip) at startup. The ip corresponding to the api is obtained by the service registry in batches. When there are a large number of apis, the set method of the degree manager is frequently invoked, resulting in a very slow startup and a batch new method is required.

### Describe how you did it
Added the method of adding rules incrementally in batches.


LearningGp (changes requested):
Unit tests are required.

Issue #3110, opened 2023-04-27
Add global fallback support for AspectJ annotation extension


sczyh30:
## Issue Description

Type: *feature request*

### Describe what feature you want

Add global fallback support for AspectJ annotation extension.

`@SentinelResource` 注解方式埋点支持全局自定义 fallback。

### Describe your initial design (if present)

Add an interface like this, and support creating `SentinelResourceAspect` with it:

```java
public interface SentinelAnnotationGlobalFallback {

    Object handle(Method originalMethod, Object[] args, Throwable t) throws Throwable;
}
```

luffy0223:
Referring to the revised issue description, I have redesigned the `SentinelResourceAspect`  and its usage.

PR #3383, opened 2024-04-23
Add global fallback support for AspectJ annotation extension

luffy0223:
Add global fallback support for AspectJ annotation extension

Fixes https://github.com/alibaba/Sentinel/issues/3110

### Describe how you did it
add global fallback interface and let it as the member of @SentinelResource,
cache and call the global fallback instance in the aspect when origin method throws Throwable

### Describe how to verify it
Use unit test to verify the correctness

### Special notes for reviews
NONE

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
@@ -79,6 +79,13 @@ public @interface SentinelResource {
      */
     String defaultFallback() default "";
 
+
+    /**
+     * The {@code globalFallback} is used as the global fallback handler class.
+     * The returned class should implement the {@link IGlobalFallback} interface.
+     * If no specific global fallback handler is provided, the default {@link DefaultGlobalFallback} class will be used.
+     * @return The class that serves as the global fallback handler.
+     */
     Class<? extends IGlobalFallback> globalFallback() default DefaultGlobalFallback.class;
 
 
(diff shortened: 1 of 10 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/annotation/SentinelResource.java
```diff
@@ -77,6 +79,16 @@
...
     String defaultFallback() default "";
 
+
+    /**
+     * The {@code globalFallback} is used as the global fallback handler class.
+     * The returned class should implement the {@link IGlobalFallback} interface.
+     * If no specific global fallback handler is provided, the default {@link DefaultGlobalFallback} class will be used.
+     * @return The class that serves as the global fallback handler.
+     */
+    Class<? extends IGlobalFallback> globalFallback() default DefaultGlobalFallback.class;
```
LearningGp, line 89: IMO, the scenario the related issue(#3110) aims to solve is to have a global fallback even when neither fallback nor defaultFallback is configured in the annotations. So, the global fallback should be configured in the construction method of SentinelResourceAspect, rather than through annotations (like what defaultFallback did).

sentinel-extension/sentinel-annotation-aspectj/src/test/java/com/alibaba/csp/sentinel/annotation/aspectj/integration/service/FooService.java
```diff
@@ -53,6 +53,25 @@ public String fooWithFallback(int i) throws Exception {
         return "Hello for " + i;
     }
 
+    @SentinelResource(value = "apiFooWithFallback", globalFallback = AnnotationGlobalFallback.class,
```
LearningGp, line 56: If convenient, could you please add relevant example code in the sentinel-demo-annotation-spring-aop module?

Issue #3314, opened 2024-01-24
Proposal about expand more channel for metric export


Daydreamer-ia:
## Issue Description
Hi, community. I am a new learner for sentinel whose design is amazing to me. When i was reading the source code, i found that the mertic export is so simple currently which can be found in `MetricWriter`. The channel for exporting is only to export to a file. And this has been hard code in project which makes it difficult to extern, for example, we hope to collect all app metric into database, mongodb and so on to do data analysis but no way with current design.

### Describe what feature you want
Add the more channel or way to expand for metric exporting.

### Describe your initial design (if present)
I wanna to design a interface to do mertic export. Here are definition:
```java
public interface MetricExporter {
   // Timed call in MetricTimerListener
   void export(Map<Long, List<MetricNode>> metric);
}
```
The default implement is the origin way to export: file export.
```java
public class DefaultExporter implements MetricExporter {
    private static final MetricWriter metricWriter = new MetricWriter(SentinelConfig.singleMetricFileSize(),
        SentinelConfig.totalMetricFileCount());
   @Override
   public void export(Map<Long, List<MetricNode>> metric){
       for (Entry<Long, List<MetricNode>> entry : maps.entrySet()) {
           try {
                 metricWriter.write(entry.getKey(), entry.getValue());
            } catch (Exception e) {
                 RecordLog.warn("[MetricTimerListener] Write metric error", e);
            }
       }
  }
}
```
This interface will be held in `MetricTimerListener`.
```java
public class MetricTimerListener implements Runnable {
    public static final MetricExporter exporter = null;
    static {
       exporter = SpiLoader.of(MetricExporter .class).loadFirstInstanceOrDefault();
    }
    @Override
    public void run() {
        ....
        exporter.export(maps);
    }
}
```
I am not sure whether my idea is right. Please correct me if any mistake. Thanks.

PR #3416, opened 2024-06-09
feature: add metric exporter interface

Daydreamer-ia:
https://github.com/alibaba/Sentinel/issues/3314

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/MetricTimerListener.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/MetricTimerListener.java
@@ -25,16 +25,34 @@ import com.alibaba.csp.sentinel.Constants;
 import com.alibaba.csp.sentinel.config.SentinelConfig;
 import com.alibaba.csp.sentinel.log.RecordLog;
 import com.alibaba.csp.sentinel.node.ClusterNode;
+import com.alibaba.csp.sentinel.node.metric.export.DefaultExporter;
+import com.alibaba.csp.sentinel.node.metric.export.MetricExporter;
 import com.alibaba.csp.sentinel.slotchain.ResourceWrapper;
 import com.alibaba.csp.sentinel.slots.clusterbuilder.ClusterBuilderSlot;
+import com.alibaba.csp.sentinel.spi.SpiLoader;
 
 /**
  * @author jialiang.linjl
  */
 public class MetricTimerListener implements Runnable {
 
-    private static final MetricWriter metricWriter = new MetricWriter(SentinelConfig.singleMetricFileSize(),
-        SentinelConfig.totalMetricFileCount());
+    private static MetricExporter exporter = null;
+
+    static {
+        MetricExporter instance = null;
+        try {
+            instance = SpiLoader.of(MetricExporter.class).loadFirstInstance();
+        } catch (Throwable t) {
+            // ignore
+        }
+        if (instance == null) {
+            MetricTimerListener.exporter = new DefaultExporter();
+        } else {
+            MetricTimerListener.exporter = instance;
+        }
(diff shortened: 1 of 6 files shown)


sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/MetricTimerListener.java
```diff
@@ -25,16 +25,34 @@
...
-    private static final MetricWriter metricWriter = new MetricWriter(SentinelConfig.singleMetricFileSize(),
-        SentinelConfig.totalMetricFileCount());
+    private static MetricExporter exporter = null;
+
+    static {
+        MetricExporter instance = null;
+        try {
+            instance = SpiLoader.of(MetricExporter.class).loadFirstInstance();
+        } catch (Throwable t) {
+            // ignore
```
LearningGp, line 46: It would be better to log the exception using RecordLog.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/export/DefaultExporter.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final MetricWriter metricWriter = new MetricWriter(SentinelConfig.singleMetricFileSize(),
+            SentinelConfig.totalMetricFileCount());
+
+    @Override
+    public void export(Map<Long, List<MetricNode>> metrics) {
+        for (Map.Entry<Long, List<MetricNode>> entry : metrics.entrySet()) {
+            try {
+                metricWriter.write(entry.getKey(), entry.getValue());
+            } catch (Exception e) {
+                RecordLog.warn("[MetricTimerListener] Write metric error", e);
```
LearningGp, line 43: MetricTimerListener -> DefaultExporter

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/MetricTimerListener.java
```diff
@@ -25,16 +25,34 @@
...
 public class MetricTimerListener implements Runnable {
 
-    private static final MetricWriter metricWriter = new MetricWriter(SentinelConfig.singleMetricFileSize(),
-        SentinelConfig.totalMetricFileCount());
+    private static MetricExporter exporter = null;
+
+    static {
+        MetricExporter instance = null;
+        try {
+            instance = SpiLoader.of(MetricExporter.class).loadFirstInstance();
```
LearningGp, line 44: We may need to discuss whether it's more suitable to get the first implementation or to obtain a list to support multiple exporters.

LearningGp (changes requested):

LearningGp:
We may need to discuss whether we should support having multiple exporters active at the same time. I'm not sure about your perspective on this.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/MetricTimerListener.java
Daydreamer-ia: done

sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/export/DefaultExporter.java
Daydreamer-ia: done

Daydreamer-ia:
> We may need to discuss whether we should support having multiple exporters active at the same time. I'm not sure about your perspective on this.

I keep one leave is order to sync the old logic, to be honest, i agree with you that support multiple exporters here. Do i keep more exporters active this pr?

LearningGp:
> > We may need to discuss whether we should support having multiple exporters active at the same time. I'm not sure about your perspective on this.
> 
> I keep one leave is order to sync the old logic, to be honest, i agree with you that support multiple exporters here. Do i keep more exporters active this pr?

For this PR, let's keep the single export implementation as is.

Issue #3434, opened 2024-07-23
[BUG] csp.sentinel.log.dir参数对sentinel-dashboard无效


Bryan2333:
## Issue Description

Type: *bug report*

### Describe what happened
-Dcsp.sentinel.log.dir=/xx/xx设置日志保存路径后，sentinel-dashboard依然会在home目录下面创建logs文件夹

### Describe what you expected to happen
sentinel dashboard日志保存到指定的文件夹

### How to reproduce it (as minimally and precisely as possible)

1. 
2. 
3.

furaul:
![image](https://github.com/user-attachments/assets/1ce2405c-58f2-4ef5-b26b-1a2be851ffad)
![image](https://github.com/user-attachments/assets/fa0408e5-5416-479a-a453-101310a7bc7a)
Because of this logging setting, logback will make '/logs' dir while DashboardApplication start up.
And the 'sentinel-dashboard.log' is still in the '/logs/cps' dir.
I can help to fix this problem.

PR #3435, opened 2024-07-24
Fix sentinel-dashboard.log wrong path problems

furaul:
Fix sentinel-dashboard.log wrong path problems while user defined {csp.sentinel.log.dir}.
https://github.com/alibaba/Sentinel/issues/3434

### Describe how you did it
correct logging.file.name setting in sentinel-dashboard/src/main/resources/application.properties.

### Describe how to verify it
1. add -Dcsp.sentinel.log.dir=/Users/XXXX/logs1 in VM options.
2. While the DashboardApplication startup, the ${user.home}/logs/csp will not be created.


LearningGp:
Suggest keeping default logging path. To override, use logging.file.name property.

PR #3444, opened 2024-08-24
Fix 3443

### LINKED ISSUE #3443: [BUG] DefaultBlockExceptionHandler treats all BlockException as FlowException
## Issue Description

Type: *bug report*

### Describe what happened
The DefaultBlockExceptionHandler treat all BlockException as FlowException, even if it is a DegradeException. This is not good.

It will set appropriate http status code and message accroding to the type of BlockException.

1.  try to create a degrade exception
2.  it will get 429(too many request) and ”Blocked by Sentinel (flow limiting)"
3.

Linzyoo:
This PR is to update DefaultExceptionHandler with appropriate httpstatus code and message for different types of BlockException.

Fixes #3443
### Describe how you did it
Create an enum for all BlockException and set appropriate httpstatus code and message.

### Describe how to verify it
Add test cases for DefaultExceptionHandler

### DIFF
--- a/sentinel-adapter/sentinel-spring-webmvc-adapter/README.md
+++ b/sentinel-adapter/sentinel-spring-webmvc-adapter/README.md
@@ -28,6 +28,12 @@ public class InterceptorConfig implements WebMvcConfigurer {
         // Add to the interceptor list.
         registry.addInterceptor(new SentinelWebInterceptor(config)).addPathPatterns("/**");
     }
+
+    @Bean
+    public SentinelExceptionAware sentinelExceptionAware(){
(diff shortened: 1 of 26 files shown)


LearningGp:
Maintaining consistency in the exceptions DefaultBlockExceptionHandler throws is advisable. In typical use cases, users uniformly handle traffic control exceptions from Sentinel (notably AuthorityException, which serves as an edge case not central to Sentinel's core functionality). Recommended practices suggest allowing users to implement their own BlockExceptionHandler interface to clearly differentiate these exceptions during implementation.

PR #3465, opened 2024-10-27
Fix for Incorrect method of reading indicator files

### LINKED ISSUE #3464: [BUG] 避免字符串和MetricNode对象的转换,减少内存占用
https://github.com/alibaba/Sentinel/blob/195150bc745927429e9f14f501907310b46d702f/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/MetricsReader.java#L54-L56

55行没必要为了获取时间戳而将字符串转成``MetricNode``对象, 直接通过字符串截取就行, 
```java
int timestampIndex = s.indexOf("|");
String timestamp = Long.parseLong(s.substring(0, timestampIndex));
```
https://github.com/alibaba/Sentinel/blob/195150bc745927429e9f14f501907310b46d702f/sentinel-transport/sentinel-transport-common/src/main/java/com/alibaba/csp/sentinel/command/handler/SendMetricCommandHandler.java#L97-L100
SendMetricCommandHandler中为了得到`ThinString`, 又将`MetricNode`转成字符串, `MetricNode`成了冗余转换, 会增加内存占用
不如直接采用`FatString`字符串分割
```java
int timestampIndex = s.indexOf("|");
String timestamp = s.substring(0, timestampIndex);

int i = s.indexOf("|", timestampIndex + 20);
String metric = s.substring(i);

System.out.println(timestamp + metric);
```

manojks1999:
Fixes for #3462 

Fixes #3464 

### Describe how you did it
Removed line 209 of sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/MetricSearcher.java

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/MetricSearcher.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/MetricSearcher.java
@@ -206,7 +206,6 @@ public class MetricSearcher {
             long second;
             lastPosition.offsetInIndex = in.getChannel().position();
             while ((second = indexIn.readLong()) < beginSecond) {
-                offset = indexIn.readLong();
                 lastPosition.offsetInIndex = in.getChannel().position();
             }
             offset = indexIn.readLong();


brucelwl (approved):
ok

LearningGp:
This code is currently processing an index file. I'm having trouble mapping the issue description to the observed behavior - could you please elaborate on the exact conditions/reproduction steps you're encountering? Additionally, if we remove this line, the while condition would incorrectly compare Index with beginSecond, which is semantically invalid as these represent different dimensional units in our processing pipeline.

PR #3480, opened 2024-12-13
fix window deprecated check

SpecialYang:
The logic for determining whether the window has expired is incorrect.

The bucket time range is [startTime, startTime + interval).


LearningGp:
FYI #568

PR #3569, opened 2025-10-23
fix the bugs about http-method-specify in spring-webmvc-v6x and incorrect time unit in dashboard

ylnxwlp:
Some bugs have caused problems for users when they were using the product.

Fix the issue in the spring-webmvc-v6x-adapter where, after configuring http-method-specify:true, the prefixes for different request methods were not concatenated.
And Fix incorrect time unit in the tooltip when adding a new circuit breaker rule with statistical window duration.

fixes the issue #3562 and #3563 

### Describe how you did it
Add the judgment of isHttpMethodSpecify and isMseHttpMethodSpecify, and according to the conditions, add the request type prefix for them.

Modify app.js to ensure the correct unit of time.

### Describe how to verify it
unit test


sentinel-adapter/sentinel-spring-webmvc-v6x-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/spring/webmvc_v6x/SentinelWebInterceptor.java
```diff
@@ -62,6 +63,9 @@ protected String getResourceName(HttpServletRequest request) {
         if (config.isContextPathSpecify() && request.getContextPath() != null) {
             resourceName = request.getContextPath() + resourceName;
         }
+        if (StringUtil.isNotEmpty(resourceName) && (config.isHttpMethodSpecify() || SentinelWebMvcConfig.isMseHttpMethodSpecify())) {
```
LearningGp: The mseHttpMethodSpecify field is a previously introduced mistake and should not exist. Please help remove it.

LearningGp (changes requested):

sentinel-adapter/sentinel-spring-webmvc-v6x-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/spring/webmvc_v6x/SentinelWebInterceptor.java
ylnxwlp: Fixed as suggested.

ylnxwlp:
I noticed that there was an issue with the execution sequence of the sub-thread and the main-thread in a test class, which resulted in the main thread being stuck in an endless wait state. Is it a bug?

LearningGp (approved):
LGTM

LearningGp:
Thanks for contributing!

PR #3587, opened 2025-12-22
修复使用FileRefreshableDataSource作为规则数据源时，初始化空规则文件会导致StringIndexOutOfBoun…

genjigjj:
…dsException异常

### Describe how you did it
判断规则文件是否有数据

### Describe how to verify it
加载空的规则文件看是否有异常打印

### DIFF
--- a/sentinel-extension/sentinel-datasource-extension/src/main/java/com/alibaba/csp/sentinel/datasource/FileRefreshableDataSource.java
+++ b/sentinel-extension/sentinel-datasource-extension/src/main/java/com/alibaba/csp/sentinel/datasource/FileRefreshableDataSource.java
@@ -118,6 +118,9 @@ public class FileRefreshableDataSource<T> extends AutoRefreshDataSource<String, 
                     + ", is bigger than bufSize=" + buf.length + ". Can't read");
             }
             int len = inputStream.read(buf);
+            if (len == -1) {
+                return null;
+            }
             return new String(buf, 0, len, charset);
         } finally {
             if (inputStream != null) {


genjigjj:
#3575 基于1.8修复该问题后提交新的PR

sentinel-extension/sentinel-datasource-extension/src/main/java/com/alibaba/csp/sentinel/datasource/FileRefreshableDataSource.java
```diff
@@ -118,6 +118,9 @@ public String readSource() throws Exception {
                     + ", is bigger than bufSize=" + buf.length + ". Can't read");
             }
             int len = inputStream.read(buf);
+            if (len == -1) {
+                return null;
```
oss-sentinel-ai, line 122: Nice catch. Returning `null` for an empty file is consistent with downstream semantics: rule managers treat a null rule list as "no rules" (e.g. `FlowRuleUtil.buildFlowRuleMap` returns an empty map for null input), so an empty rule file now correctly clears rules instead of throwing `StringIndexOutOfBoundsException`.

One suggestion (non-blocking): consider adding a regression test — e.g. create an empty temp file, construct a `FileRefreshableDataSource` with a converter that tolerates null, and assert `readSource()` returns null / no exception is thrown. Note that converters calling methods directly on the source (e.g. `source.trim()`) will now receive null instead of crashing earlier; that is acceptable but worth documenting in the javadoc of `readSource()`.

oss-sentinel-ai (approved):
### Summary
Fixes `StringIndexOutOfBoundsException` when `FileRefreshableDataSource` initializes from an empty rule file: `inputStream.read(buf)` returns `-1` for an empty file, and passing that length to `new String(buf, 0, len, charset)` throws. The added guard returns `null`, which downstream rule managers treat as "no rules" (verified: `FlowRuleUtil.buildFlowRuleMap` returns an empty map for null input). Minimal, correct and backward compatible.

### Findings
- **[Info]** FileRefreshableDataSource.java:122 — suggest a regression test for the empty-file case and a javadoc note that converters may now receive null.

LGTM.

---

PR #3592, opened 2026-01-21
【Enhancement】优化主流程lookProcessChain方法 issue #3513

LilMosey:
issue https://github.com/alibaba/Sentinel/issues/3513
（详细信息都在issue里）

issue https://github.com/alibaba/Sentinel/issues/3513

### Describe how you did it
image

### Describe how to verify it
实现逻辑参照Collections.synchronizedMap()
image

原逻辑做了一次copyOnWrite操作，加锁后其实没必要进行copyOnWrite造成了资源浪费增加gc压力


LilMosey:
@LearningGp  大佬能帮忙review一下吗？去年提过一次pr，但是当时用的是master分支

sentinel-core/src/main/java/com/alibaba/csp/sentinel/CtSph.java
```diff
@@ -192,22 +190,18 @@ public Entry entry(ResourceWrapper resourceWrapper, int count, Object... args) t
      * @return {@link ProcessorSlotChain} of the resource
      */
     ProcessorSlot<Object> lookProcessChain(ResourceWrapper resourceWrapper) {
-        ProcessorSlotChain chain = chainMap.get(resourceWrapper);
+        ProcessorSlotChain chain = CHAIN_MAP.get(resourceWrapper);
         if (chain == null) {
-            synchronized (LOCK) {
-                chain = chainMap.get(resourceWrapper);
+            synchronized (CHAIN_MAP) {
```
oss-sentinel-ai, line 195: **[Info]** Locking on the `ConcurrentHashMap` instance itself works here because every mutation path (`put` here, `clear` in `resetChainMap`) consistently uses the same monitor. However, the CHM Javadoc explicitly discourages locking on the map instance for auxiliary protocols. Keeping a dedicated `private static final Object LOCK` (as before) would separate the lock role from the data structure and prevent future callers/tests from accidentally sharing this monitor. Pure style suggestion — no correctness issue.

sentinel-core/src/main/java/com/alibaba/csp/sentinel/CtSph.java
```diff
@@ -221,7 +215,9 @@ ProcessorSlot<Object> lookProcessChain(ResourceWrapper resourceWrapper) {
      * @since 0.2.0
      */
     public static int entrySize() {
-        return chainMap.size();
+        synchronized (CHAIN_MAP) {
```
oss-sentinel-ai, line 218: **[Info]** `ConcurrentHashMap.size()` and `clear()` are already thread-safe, so the added `synchronized` blocks in `entrySize()` and `resetChainMap()` are not strictly required. Harmless, but they serialize these calls with chain creation unnecessarily. Note that `resetChainMap()` actually becomes *safer* with CHM: the old code called `HashMap.clear()` on the live map that lock-free readers were concurrently reading.

oss-sentinel-ai (approved):
### Summary
Replaces the volatile `HashMap` + copy-on-write design in `CtSph` with a final `ConcurrentHashMap`, removing the full-map copy on every new-resource registration while keeping the hot-path `lookProcessChain` read lock-free. The change is correct: chain creation stays serialized under a single monitor, so the `MAX_SLOT_CHAIN_SIZE` boundary semantics are preserved exactly — the new capacity-boundary test would not pass with a naive `computeIfAbsent` replacement, since per-bin locking could admit two concurrent inserts at the boundary. It also fixes a latent data race: the old `resetChainMap()` called `HashMap.clear()` on the very map read lock-free by concurrent readers, while `ConcurrentHashMap.clear()` is properly thread-safe. Tests are deterministic (latch-triggered starts, timeouts used only as hang guards) and Java 8 compatible. LGTM — two minor style notes inline.

### Findings
- **[Info]** sentinel-core/src/main/java/com/alibaba/csp/sentinel/CtSph.java:195 — `synchronized (CHAIN_MAP)`: locking on the map instance is correct here (all mutation paths share the monitor) but is discouraged by the CHM Javadoc; a dedicated lock object would be cleaner
- **[Info]** sentinel-core/src/main/java/com/alibaba/csp/sentinel/CtSph.java:218 — the added `synchronized` blocks in `entrySize()` / `resetChainMap()` are not strictly required since CHM `size()`/`clear()` are already thread-safe; harmless

### Suggestions
- Optional: keep a dedicated `private static final Object LOCK` (as before) for the creation critical section instead of locking on `CHAIN_MAP` itself.

---

PR #3616, opened 2026-05-14
Fix invalid prototype slot node issue

### LINKED ISSUE #3007: *bug report* ProcessorSlot的Spi中的isSingleton = false设置不生效
## Issue Description

Type: *bug report*

当在非单例模式的ProcessorSlot（A）前加一个单例模式的ProcessorSlot（B）后，所有ProcessorSlotChain中B的next都会同步修改为最新创建的A

在上述行为后，不同的ProcessorSlotChain中B的next应该为各自新创建的A，而非共同的A

1. 添加一个ProcessSlot实现类，添加注解@Spi(order = Spi.ORDER_HIGHEST)

### Tell us your environment
sentinel 1.8.3

DefaultProcessorSlotChain中添加next的方式是直接修改传入的AbstractLinkedProcessorSlot，但是传入的ProcessorSlot有可能是单例或非单例，倘若是单例，则在其后添加next会影响所有的ProcessorSlotChain

weixq709:

When a singleton `ProcessorSlot` is placed before a prototype (non-singleton)
`ProcessorSlot` in `DefaultProcessorSlotChain`, all chains share the same singleton
instance. Each time a new chain is built, `addLast` calls `setNext` directly on
the shared singleton, overwriting its `next` pointer. This causes previously built
chains to lose their own prototype slot node — making the prototype slot invalid.

This PR introduces `PrototypeSlotWrapper` and modifies `DefaultProcessorSlotChain#addLast`
to wrap every slot in a new `PrototypeSlotWrapper` per chain. Each wrapper owns an
independent `next` pointer (inherited from `AbstractLinkedProcessorSlot`), so the
singleton's `next` field is never modified during chain construction, and each chain
maintains its own correct node linkage.

Fixes #3007

### Describe how you did it

- Added `PrototypeSlotWrapper` (`slots/block/flow/PrototypeSlotWrapper.java`):
  a thin wrapper extending `AbstractLinkedProcessorSlot` that delegates `entry`,
  `exit`, `fireEntry`, and `fireExit` to the underlying slot, while keeping its
  own chain-local `next` pointer via the inherited field.

- Modified `DefaultProcessorSlotChain#addLast`
  (`slotchain/DefaultProcessorSlotChain.java`):
  wrap the incoming slot in a `PrototypeSlotWrapper` before appending it to the
  chain, so `end.setNext` and `end = processor` always operate on a fresh wrapper
  instance rather than the slot itself.

### Describe how to verify it

Two test cases added in `ProcessorSlotChainTest` (`slots/ProcessorSlotChainTest.java`):

- `testBeforeFix_nextOfSingletonIsShared`: uses a Mockito spy to simulate the old
  `addLast` logic (no wrapper) and asserts that after building two chains, both
  chains' prototype slot node is the same object — confirming the bug.

- `testAfterFix_eachChainHasItsOwnWrapper`: uses the fixed `DefaultProcessorSlotChain`
  and asserts that the two chains hold different `PrototypeSlotWrapper` instances as
  their second node — confirming the fix.

### Special notes for reviews

`PrototypeSlotWrapper` does not override `getNext` / `setNext`, so the chain-local
`next` field on the wrapper is used for node linkage. `fireEntry` and `fireExit` are
delegated to the underlying slot to preserve the original propagation behavior.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slotchain/DefaultProcessorSlotChain.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slotchain/DefaultProcessorSlotChain.java
@@ -16,6 +16,7 @@
 package com.alibaba.csp.sentinel.slotchain;
 
 import com.alibaba.csp.sentinel.context.Context;
+import com.alibaba.csp.sentinel.slots.block.flow.PrototypeSlotWrapper;
 
 /**
  * @author qinan.qn
@@ -50,8 +51,9 @@ public class DefaultProcessorSlotChain extends ProcessorSlotChain {
 
     @Override
     public void addLast(AbstractLinkedProcessorSlot<?> protocolProcessor) {
-        end.setNext(protocolProcessor);
-        end = protocolProcessor;
+        PrototypeSlotWrapper processor = new PrototypeSlotWrapper(protocolProcessor);
+        end.setNext(processor);
+        end = processor;
     }
 
     /**
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/PrototypeSlotWrapper.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/PrototypeSlotWrapper.java
@@ -1,0 +1,39 @@
+package com.alibaba.csp.sentinel.slots.block.flow;
+
+import com.alibaba.csp.sentinel.context.Context;
+import com.alibaba.csp.sentinel.slotchain.AbstractLinkedProcessorSlot;
+import com.alibaba.csp.sentinel.slotchain.ResourceWrapper;
+
+/**
+ * @author joyce
+ * @date 2026/5/14
+ * @since 1.8.5
+ **/
+public class PrototypeSlotWrapper extends AbstractLinkedProcessorSlot<Object> {
+
+    private final AbstractLinkedProcessorSlot delegate;
+
+    public PrototypeSlotWrapper(AbstractLinkedProcessorSlot delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public void fireEntry(Context context, ResourceWrapper resourceWrapper, Object obj, int count, boolean prioritized, Object... args) throws Throwable {
+        this.delegate.fireEntry(context, resourceWrapper, obj, count, prioritized, args);
+    }
+
+    @Override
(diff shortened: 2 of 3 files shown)


LearningGp:
image

TwistedRiCen:
Thanks for working on #3007. I reproduced the bug on the current `1.8` branch and built a local-only context-wrapper PoC (not pushed / no competing PR).

Baseline regression: with a shared predecessor B and per-chain successors A1/A2, building chain 2 makes chain 1 execute `[B, A2]` instead of `[B, A1]`.

One important propagation detail for a delegate wrapper: the delegate calls its own inherited `fireEntry/fireExit`, so a wrapper-owned `next` alone is not visible to it. In my PoC each chain owns a wrapper node, and the wrapper installs a source-bound, dynamically scoped successor context while invoking the delegate. `AbstractLinkedProcessorSlot.getNext()` consults that context only when the active source is the current delegate. The previous context is restored in `finally`.

That variant passed:
- the original entry and exit regression;
- repeated `fireEntry` from one slot;
- nested chains;
- delegate exceptions / ThreadLocal cleanup;
- 32 concurrent chains × 100 invocations;
- SPI lifecycle assertions: prototype `NodeSelectorSlot` delegates differ, singleton `LogSlot` delegate is shared;
- all 228 `sentinel-core` tests;
- `mvn -f sentinel-core/pom.xml -DskipTests verify` (PMD).

Remaining review points: chain traversal now exposes wrapper nodes rather than concrete slots, each invocation pays ThreadLocal get/set/remove costs, and slots overriding `fireEntry/fireExit` could bypass the bridge. So I see this as evidence that the wrapper direction is viable, but not yet proof that its compatibility trade-off is acceptable. I can provide the local patch/test file if useful.

TwistedRiCen:
Cross-linking for maintainers' visibility: the wrapper + scoped successor context PoC mentioned in my Aug 13 comment is now submitted as #3646 (against the `1.8` branch).

It builds directly on the wrapper design of this PR — same chain-local `next` ownership in a wrapper node — and adds a source-scoped successor context so that a delegate invoking the inherited `fireEntry()`/`fireExit()` propagates through the wrapper's successor instead of stopping at the first wrapped node. No reflection, no per-chain re-instantiation, singleton contract preserved.

@weixq709 your review of the trade-offs would be very welcome. Either design could proceed; if maintainers prefer the simpler wrapper-only variant, I'm glad to reshape #3646 accordingly — and this PR's tests for wrapper distinctness would carry over either way.

PR #3619, opened 2026-05-15
fix: allow regex resource names with all control behaviors

### LINKED ISSUE #3544: [BUG] 资源名称使用正则表达式时，当使用非快速失败的CONTROL_BEHAVIOR时，配置的流控规则不会加载
## Issue Description

Type: *bug report*

### Describe what happened
请问下流控规则，资源名称要使用正则表达式时，流控效果CONTROL_BEHAVIOR_DEFAULT要限制是快速失败才允许使用regex，当我使用其他CONTROL_BEHAVIOR时，配置的流控规则不会加载
com/alibaba/csp/sentinel/slots/block/flow/FlowRuleUtil.java:249
` if (rule.isRegex()) {
      return !rule.isClusterMode() && rule.getControlBehavior() == RuleConstant.CONTROL_BEHAVIOR_DEFAULT;
}`

1. 
2. 
3.

EvanYao826:
## Issue Description

Fixes #3544

Previously, regex resource names were only allowed with `CONTROL_BEHAVIOR_DEFAULT` (fast failure). This restriction prevented users from using warm-up or queue behaviors with regex-matched resources.

## Changes

In `FlowRuleUtil.checkRegexField()`, removed the control behavior restriction for regex rules. Now regex resource names can be used with any control behavior (DEFAULT, WARM_UP, RATE_LIMITER, WARM_UP_RATE_LIMITER), as long as the rule is not in cluster mode.

**Before:**
```java
if (rule.isRegex()) {
    return !rule.isClusterMode() && rule.getControlBehavior() == RuleConstant.CONTROL_BEHAVIOR_DEFAULT;
}
```

**After:**
```java
if (rule.isRegex()) {
    return !rule.isClusterMode();
}
```

## Testing

Verified that flow rules with regex resource names can now be loaded with all control behaviors.

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleUtil.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleUtil.java
@@ -246,7 +246,7 @@ public final class FlowRuleUtil {
             return false;
         }
         if (rule.isRegex()) {
-            return !rule.isClusterMode() && rule.getControlBehavior() == RuleConstant.CONTROL_BEHAVIOR_DEFAULT;
+            return !rule.isClusterMode();
         }
         return true;
     }


sentinel-core/src/main/java/com/alibaba/csp/sentinel/slots/block/flow/FlowRuleUtil.java
```diff
@@ -246,7 +246,7 @@ private static boolean checkRegexField(FlowRule rule) {
             return false;
         }
         if (rule.isRegex()) {
-            return !rule.isClusterMode() && rule.getControlBehavior() == RuleConstant.CONTROL_BEHAVIOR_DEFAULT;
+            return !rule.isClusterMode();
```
oss-sentinel-ai, line 249: WarmUpController and WarmUpRateLimiterController maintain per-instance token bucket state (`storedTokens`, `lastFilledTime`) that is shared across all resources matching the regex pattern, but `syncToken()` reads `passQps()` from the individual per-resource Node. When a regex rule matches resources with different traffic levels, the warm-up behavior becomes unpredictable: sync triggered by a low-QPS resource keeps the bucket full (perpetual warm-up state), while a high-QPS resource drains it quickly. ThrottlingController (rate limiter) works correctly since its shared `latestPassedTime` naturally implements a global rate limit. Consider either: (1) restricting this change to only allow `CONTROL_BEHAVIOR_RATE_LIMITER` with regex, or (2) documenting that warm-up behaviors with regex have shared-state semantics across all matched resources.

oss-sentinel-ai, line 249: The identical regex + control-behavior restriction exists in `ParamFlowRuleUtil.checkRegexField()` (sentinel-extension/sentinel-parameter-flow-control). If the restriction is being relaxed here for flow rules, consider whether the same change should apply to parameter flow rules for consistency.

oss-sentinel-ai (commented):
### Summary
This PR relaxes `checkRegexField` so regex flow rules can be combined with all control behaviors (previously restricted to `CONTROL_BEHAVIOR_DEFAULT`). The change is safe for `DefaultController` (stateless) and `ThrottlingController` (its shared `latestPassedTime` naturally acts as a global rate limiter), but there is a semantic mismatch with the warm-up behaviors — see inline comments.

### Findings
- **[Warning]** `FlowRuleUtil.java:249` — Warm-up controllers keep per-instance token-bucket state shared across all regex-matched resources while `syncToken()` reads per-resource QPS; warm-up behavior becomes unpredictable when matched resources have different traffic (details inline).
- **[Info]** `FlowRuleUtil.java:249` — The identical restriction still exists in `ParamFlowRuleUtil.checkRegexField()` for parameter flow rules (details inline).

### Suggestions
- Consider allowing only `CONTROL_BEHAVIOR_RATE_LIMITER` (plus default) with regex rules, or explicitly document that warm-up behaviors operate on shared state across all matched resources.
- Add an integration test covering regex + rate-limiter behavior (existing tests only cover regex + default).

---

PR #3626, opened 2026-06-01
fix: stabilize flaky ParamFlowDefaultCheckerTest by using mocked time

### LINKED ISSUE #2426: [bug] test case ParamFlowDefaultCheckerTest#testParamFlowDefaultCheckSingleValueCheckQpsMultipleThreads() assert  error
// 我看到社区大家都用中文，我也用中文好了

## Issue Description

单元测试运行失败，或许是测试用例逻辑错误，或许是代码被修改后逻辑不符合预期

我试着运行这个项目，但在check阶段出现了测试错误，或许是测试逻辑问题，或许是代码本身有bug

1.拉取master最新的代码

2. run :
`mvn clean install`

ParamFlowDefaultCheckerTest 状态为未通过

3. run :
`mvn test`

ParamFlowDefaultCheckerTest 状态为未通过

4.单独运行ParamFlowDefaultCheckerTest 
方法testParamFlowDefaultCheckSingleValueCheckQpsMultipleThreads()断言失败

```
java.lang.AssertionError: 
Expected :8
Actual   :5

... ...
	at com.alibaba.csp.sentinel.slots.block.flow.param.ParamFlowDefaultCheckerTest.testParamFlowDefaultCheckSingleValueCheckQpsMultipleThreads(ParamFlowDefaultCheckerTest.java:323)
... ...
```
5. 重复运行testParamFlowDefaultCheckSingleValueCheckQpsMultipleThreads()断言成功
6. 多次重复运行testParamFlowDefaultCheckSingleValueCheckQpsMultipleThreads()会出现断言失败
```
java.lang.AssertionError: 
Expected :6
Actual   :5

... ...
	at com.alibaba.csp.sentinel.slots.block.flow.param.ParamFlowDefaultCheckerTest.testParamFlowDefaultCheckSingleValueCheckQpsMultipleThreads(ParamFlowDefaultCheckerTest.java:323)
... ...
```
多次运行的结果：会出现几率性运行成功，但大概率会失败
### Tell us your environment
oracle-jdk8 / master/ mac0s

使用oracle-jdk16 OccupiableBucketLeapArrayTest会实现失败，目前尚未确定是否是因为jdk版本缘故

EvanYao826:
## Summary

The test `testParamFlowDefaultCheckSingleValueCheckQpsMultipleThreads` was flaky because it used real time (`TimeUnit.sleep` and wall-clock-based loops) for concurrent assertions. This caused race conditions when the test ran in isolation vs. after other tests.

### Root Cause

The test's second phase used `TimeUtil.currentTimeMillis()` in a busy loop with real thread sleep, making the assertion result depend on:
- Exact thread scheduling timing
- Whether a previous test's time mock was still active
- Wall-clock drift during the 2.5-second loop

### Fix

- Wrapped the entire test in `MockedStatic<TimeUtil>` (consistent with all other tests in this class)
- Replaced the real-time busy loop with a clean second round of concurrent requests after advancing the mock clock past the duration window
- This makes the test deterministic regardless of thread scheduling or test execution order

Fixes #2426

### DIFF
--- a/sentinel-extension/sentinel-parameter-flow-control/src/test/java/com/alibaba/csp/sentinel/slots/block/flow/param/ParamFlowDefaultCheckerTest.java
+++ b/sentinel-extension/sentinel-parameter-flow-control/src/test/java/com/alibaba/csp/sentinel/slots/block/flow/param/ParamFlowDefaultCheckerTest.java
@@ -248,71 +248,65 @@ public class ParamFlowDefaultCheckerTest extends AbstractTimeBasedTest {
 
     @Test
     public void testParamFlowDefaultCheckSingleValueCheckQpsMultipleThreads() throws Exception {
-        final String resourceName = "testParamFlowDefaultCheckSingleValueCheckQpsMultipleThreads";
-        final ResourceWrapper resourceWrapper = new StringResourceWrapper(resourceName, EntryType.IN);
-        int paramIdx = 0;
+        try (MockedStatic<TimeUtil> mocked = super.mockTimeUtil()) {
+            final String resourceName = "testParamFlowDefaultCheckSingleValueCheckQpsMultipleThreads";
+            final ResourceWrapper resourceWrapper = new StringResourceWrapper(resourceName, EntryType.IN);
+            int paramIdx = 0;
 
-        long threshold = 5L;
+            long threshold = 5L;
 
-        final ParamFlowRule rule = new ParamFlowRule();
-        rule.setResource(resourceName);
-        rule.setCount(threshold);
-        rule.setParamIdx(paramIdx);
-        rule.setDurationInSec(3);
+            final ParamFlowRule rule = new ParamFlowRule();
+            rule.setResource(resourceName);
+            rule.setCount(threshold);
+            rule.setParamIdx(paramIdx);
+            rule.setDurationInSec(3);
 
-        final String valueA = "valueA";
-        ParameterMetric metric = new ParameterMetric();
-        ParameterMetricStorage.getMetricsMap().put(resourceWrapper.getName(), metric);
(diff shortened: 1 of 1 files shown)


sentinel-extension/sentinel-parameter-flow-control/src/test/java/com/alibaba/csp/sentinel/slots/block/flow/param/ParamFlowDefaultCheckerTest.java
```diff
@@ -248,71 +248,65 @@ public void testParamFlowDefaultCheckQpsInDifferentDuration() throws Interrupted
...
+                t.setName("sentinel-simulate-traffic-task-" + i);
+                t.start();
+            }
+            waitLatch.await();
+
+            assertEquals(threshold, successCount.get());
+            successCount.set(0);
+
+            // Advance time past the duration window to reset counters
+            sleep(mocked, rule.getDurationInSec() * 1000);
```
oss-sentinel-ai, line 291: `MockedStatic` scopes stubbing to the thread that created it, so the 40 worker threads still call the real `TimeUtil.currentTimeMillis()` inside `ParamFlowChecker`. Advancing this mock therefore does not advance their token-counter timestamps; the second round runs less than three real seconds after the first and is rejected rather than allowing `threshold` requests. Use a time source that is shared by the worker threads (or install the static mock within each worker) before removing the real wait.

oss-sentinel-ai (changes requested):
### Summary
The PR replaces the wall-clock delay with mocked time, but the thread-local static mock does not affect the concurrent checks, so the revised test is not correct or deterministic.

---

PR #3630, opened 2026-06-24
Guard MetricNode.fromThinString against lines with fewer than 7 fields

vasiliy-mikhailov:
## Problem

`MetricNode.fromThinString` splits the line on `|` and then reads `strs[0]` through `strs[6]` unconditionally:

```java
String[] strs = line.split("\\|");
node.setTimestamp(Long.parseLong(strs[0]));
...
node.setRt(Long.parseLong(strs[6]));
if (strs.length >= 8) {            // trailing fields are already guarded
    node.setOccupiedPassQps(Long.parseLong(strs[7]));
}
```

A line with fewer than seven fields (a short or truncated metric line) throws `ArrayIndexOutOfBoundsException`, even though the optional trailing fields are already protected with length checks.

## Fix

Apply the same length guard to the leading fields, so a short line is parsed defensively (populating the fields that are present) instead of crashing — consistent with the existing handling of `strs[7]`/`strs[8]`.

## Test

Adds `testFromThinStringShortInput` to `MetricNodeTest` with a 3-field line. It fails on `1.8` (`ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3`) and passes with this change.

## AI assistance disclosure

### DIFF
--- a/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/MetricNode.java
+++ b/sentinel-core/src/main/java/com/alibaba/csp/sentinel/node/metric/MetricNode.java
@@ -183,13 +183,27 @@ public class MetricNode {
     public static MetricNode fromThinString(String line) {
         MetricNode node = new MetricNode();
         String[] strs = line.split("\\|");
-        node.setTimestamp(Long.parseLong(strs[0]));
-        node.setResource(strs[1]);
-        node.setPassQps(Long.parseLong(strs[2]));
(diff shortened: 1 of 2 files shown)


oss-sentinel-ai (approved):
### Summary
LGTM. Adds length guards for the first seven fields in `MetricNode.fromThinString`, consistent with the existing guards for the later optional fields, so short/malformed lines no longer throw `ArrayIndexOutOfBoundsException`. Regression test included.

---

PR #3632, opened 2026-06-25
Guard TokenClientHandler.getRemoteAddress against unresolved remote address

vasiliy-mikhailov:
`TokenClientHandler.getRemoteAddress` does:

```java
inetAddress.getAddress().getHostAddress()
```

For an unresolved `InetSocketAddress` (the remote address never resolved, or `createUnresolved(...)`), `getAddress()` returns `null`, so this throws `NullPointerException`. It is reached from `channelUnregistered`, which logs the remote address during channel cleanup, so a channel whose remote address is unresolved NPEs on unregister.

This falls back to `getHostName()` when `getAddress()` is null. Adds a test that fails before the change (NPE) and passes after.

## AI assistance disclosure


oss-sentinel-ai (approved):
### Summary
LGTM. Handles unresolved `InetSocketAddress` in `TokenClientHandler.getRemoteAddress` by falling back to the hostname when `getAddress()` returns null, avoiding an NPE. Test included.

---

PR #3633, opened 2026-06-25
Guard MetricFileNameComparator against filenames with fewer than 3 parts

vasiliy-mikhailov:
`MetricWriter.MetricFileNameComparator.compare` splits the metric file name on `"."` and reads index `[2]` (and `[3]` for pid-prefixed names) without checking the array length. A name with fewer parts (for example `"metrics.log"`) therefore throws `ArrayIndexOutOfBoundsException` during sorting. This falls back to an empty string when the part is absent. Adds a test.

## AI assistance disclosure


oss-sentinel-ai (approved):
### Summary
LGTM. Guards `MetricFileNameComparator` against metric file names with fewer dot-separated parts than expected, preventing `ArrayIndexOutOfBoundsException`; the added regression test covers the short-filename case. Trivial, well-scoped defensive fix.

---

